Issue #1, opened 2022-01-18
Flow-style logging for the protocol adaptors


ted-ross:
A separate logging/tracing facility is needed for the protocol adaptors to serve users that are more accustomed to network/firewall/router logs and are not concerned with AMQP/messaging/link-protocol details.

This feature will make the use of the protocol adaptors more observable and debuggable to network-familiar developers and operators.

PR #2, opened 2022-01-18
Issue #1 - Flow-style protocol logging

ted-ross:


src/router_core/router_core_thread.c
```diff
@@ -164,8 +175,10 @@ void qdr_adaptors_finalize(qdr_core_t *core)
     //
     qdrc_adaptor_t *adaptor = DEQ_TAIL(registered_adaptors);
     while (adaptor) {
+        DEQ_REMOVE(registered_adaptors, adaptor);
         adaptor->on_final(adaptor->context);
-        adaptor = DEQ_PREV(adaptor);
+        free(adaptor);
+        adaptor = DEQ_TAIL(registered_adaptors);
```
jiridanek, line 181: Could you leave this the way it was?

1. Not feering is not (considered) a leak not to free the `qdrc_adaptor_t*` because the memory is reachable through a global variable (`registered_adaptors`) at the point of shutdown.
2. Freeing means that it is not possible to initialize the router in the same process twice, for testing purposes. Once you destruct `registered_adaptors`, there is no easy way to get it back, given the way it is initialized at program startup.

I noticed this because I saw the `c-benchmarks` test failing. What it does is that it starts router in-process and tries to measure latency of small TCP transfers going through the router. On the second router startup, the binary crashes because adaptors are not initialized.

Thanks for your consideration. It is my fervent hope that there can be enough value to this kind of testing to balance having to make considerations like these.

kgiusti (approved):
LGTM - ship it.

PR #3, opened 2022-01-19
remove docs

pwright:
As discussed in email


kgiusti (approved):

pwright:
after discussing with @ssorj , we should hold off on this

PR #14, opened 2022-02-01
#13: Update CMake: 2.8.12 -> 3.20

### LINKED ISSUE #13: Update CMake: 2.8.12 -> 3.20
There should not be any large gotchas, with maybe the exception of [Policy CMP0065](https://cmake.org/cmake/help/latest/policy/CMP0065.html). CMake used to compile all binaries (under GCC) with `-rdynamic` by default, but according to this new policy it removed the flag. Dispatch needs the flag so it can `dlopen()` its own binary. Any tests that do router initializations also need the flag.

The solution is to either set [`ENABLE_EXPORTS`](https://cmake.org/cmake/help/latest/prop_tgt/ENABLE_EXPORTS.html#prop_tgt:ENABLE_EXPORTS) for each target that needs it, or to set [`CMAKE_ENABLE_EXPORTS`](https://cmake.org/cmake/help/latest/variable/CMAKE_ENABLE_EXPORTS.html#variable:CMAKE_ENABLE_EXPORTS) to get the old behavior.

jiridanek:
fixes #13


jiridanek:
CC @ganeshmurthy @kgiusti ready for review

(I don't have perms to touch the right-side bar in this page, so I can't notify reviewers there.)

kgiusti (approved):

jiridanek:
@kgiusti thanks; whoever has rights to merge, feel free to merge, since I don't have rights for that either ;P

PR #34, opened 2022-02-03
NO-JIRA Upgrade doctest.h: 2.4.7 -> 2.4.8 (#1505)

jiridanek:
This is brought over from qpid-dispatch, so I am keeping the commit message


jiridanek:
Ready for review. There is nothing new, only a dump of upstream code from the doctest assertion framework latest release.

PR #40, opened 2022-02-04
NO-JIRA Add macro to disable thread sanitizer

jiridanek:
Adds `ATTRIBUTE_NO_SANITIZE_THREAD` macro. This is in apache/qpid-dispatch. It will be useful for the rest of https://github.com/apache/qpid-dispatch/pull/1469 when I manage to get that working well enough.


kgiusti (approved):
Seems like an improvement to me!

PR #43, opened 2022-02-04
Review requested: proposal for an alternative implementation of router-annotations

kgiusti:
Quickly slapped together to give us something to start with.  As usual - names/conventions/values have been randomly picked by my dog: feel free to propose alternatives.

I think the implementation delta between this proposal and the re-worked MA stuff is pretty small effort-wise.  A new message section is probably the most significant change but it doesn't scare me all that much.

Of course this lacks benchmarks, but at the very least it should de-complicate the code by removing all the user-supplied MA handling at the very least.

### DIFF
--- a/docs/notes/router-annotations.adoc
+++ b/docs/notes/router-annotations.adoc
@@ -19,9 +19,8 @@ under the License
 
 = Router Message Annotations
 
-This document describes the structure and meaning of the router meta
-data included in messages passed between router nodes (interior and
-edge).
+This document describes the implementation of the meta-data included
+in messages passed between router nodes (interior and edge).
 
 == Description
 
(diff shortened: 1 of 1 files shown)


docs/notes/router-annotations.adoc
```diff
@@ -0,0 +1,128 @@
...
+forwarding path for various reasons.  Currently this information
+includes:
+
+* The identifier for the ingress router (the router attached to the
+original sending endpoint).
+
+* The trace list: an ordered list of router identifiers. This list
+reflects the path the message has travelled in the router network. The
+first entry on the list is the ingress router *(Q: can we use this*
+*instead of a dedicated ingress entry field?)*.  The current router identifier
```
jiridanek, line 38: > Q: can we use this instead of a dedicated ingress entry field?

Isn't reading a single field more efficient than decoding a list and reading the first item? (Regarding proton-c capabilities.) So if you sometimes want to only know the ingress router without knowing the full trace, separate field maybe can be more performant?

kgiusti: It would be more efficient, yes.  But I don't believe there's any case where we don't need to decode the trace list - we need it to do the proper forwarding of the message.

docs/notes/router-annotations.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+== Description
+
+Each message passed between routers includes meta-data used to
+communicate per-message state. This data is used by routers along the
+forwarding path for various reasons.  Currently this information
+includes:
+
+* The identifier for the ingress router (the router attached to the
+original sending endpoint).
```
ted-ross, line 33: To be clear.  The ingress is the first _Interior_ router in the trace.  An edge cannot be an ingress router.

kgiusti: Good point - doc updated.
@ted-ross since the trace-list (currently) does not include edge-mode routers I was thinking we could drop the ingress-router annotation entirely and instead use the first element in the trace-list for the ingress-router.  An optimization and reduction of duplicate information.  What do you think?

ted-ross: This was considered and rejected in the past.  I don't remember why.

ted-ross: I suggest that we leave this as it is.  The reason is that we may in the near future wish to include the edge-mode routers in the trace.  We can handle this by simply ignoring unrecognized trace-ids when building the trace-mask.  If we do this, the "ingress" router will not necessarily be the first router in the trace.

jiridanek (approved):

kgiusti:
https://github.com/skupperproject/skupper-router/commit/71c8962bc0dac83ca4fbf63ea99f1ba5005f7dcf

PR #44, opened 2022-02-04
WIP: Build and publish skupper-router image

### LINKED ISSUE #39: Build and publish skupper-router image

fgiorgetti:
Closes #39


.github/scripts/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM fedora:34
```
jiridanek, line 1: Why not CentOS Stream 8? That seems to me more production-like choice.

.github/scripts/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN dnf -y install glibc cyrus-sasl-lib cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 libuuid openssl gettext hostname iputils libwebsockets-devel libnghttp2 gdb valgrind && dnf -y update && dnf clean all
+ADD qpid-proton-image.tar.gz skupper-router-image.tar.gz /
```
jiridanek, line 3: The build strategy should mirror production usage. If the production build is going to compile Proton as a static library to be linked into Skupper Router, then this should probably be followed (and it also should be followed in the Skupper Router CI, which it currently is not).

@mcressman afaik the plan is to continue to compile static Proton into the qdrouterd, right?

.github/scripts/Dockerfile.builder
```diff
@@ -0,0 +1,3 @@
+FROM fedora:34
```
jiridanek, line 1: Docker (and also Podman) have support for multistage Dockerfiles. https://docs.docker.com/develop/develop-images/multistage-build/ This approach would be imo cleaner and more obvious.

.github/scripts/builder-compile.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/bash
```
jiridanek, line 1:
```bash
set -Eeuo pipefail
```
https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

.github/scripts/builder-compile.sh
```diff
@@ -0,0 +1,36 @@
...
+
+wget ${1:-http://github.com/apache/qpid-proton/archive/main.tar.gz} -O qpid-proton.tar.gz
+
+mkdir -p qpid-proton-src build staging proton_build proton_install
+tar -zxf qpid-proton.tar.gz -C qpid-proton-src --strip-components 1
+
+do_patch "patches/proton" qpid-proton-src
+
+cd proton_build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_CPP=OFF -DBUILD_RUBY=OFF -DBUILD_GO=OFF -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON -DBUILD_TLS=ON $WORKING/qpid-proton-src/ \
```
jiridanek, line 27: `-DBUILD_BINDINGS=python` works better; say what you want

Dockerfile
```diff
@@ -1,72 +0,0 @@
...
-
-# To launch a container running the new image with an interactive shell prompt:
-# > sudo docker run -i -t <username>/dispatch /bin/bash
-
-################# Begin code #######
-
-# Get the latest Centos version from dockerhub
-FROM quay.io/centos/centos:stream8
-
-MAINTAINER "dev@qpid.apache.org"
```
jiridanek, line 42: why delete this?

scripts/auto_mesh.py
```diff
@@ -0,0 +1,290 @@
+# Adds connectors in order to ensure a full mesh for a deployment or
+# similar
+
+from __future__ import print_function
```
jiridanek, line 4: this is needed only in Python 2

scripts/auto_mesh.py
```diff
@@ -0,0 +1,290 @@
...
+import ssl
+import sys
+import time
+import traceback
+
+TOKEN_FILE  = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+NAMESPACE_FILE  = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+IP=0 # Field to reference pod IP address
+TS=1 # Field to reference pod timestamp
```
jiridanek, line 18: PEP 8; two spaces before comment; CI should probably run tox tests on this

scripts/auto_mesh.py
```diff
@@ -0,0 +1,290 @@
...
+    mode = os.environ.get("QDROUTERD_AUTO_MESH_DISCOVERY")
+    if mode and len(sys.argv) > 1:
+        try:
+            if mode.upper() == "QUERY":
+                connectors = query()
+            elif mode.upper() == "INFER":
+                connectors = infer()
+            else:
+                raise Exception("Invalid value for QDROUTERD_AUTO_MESH_DISCOVERY, expected 'QUERY' or 'INFER'")
+            get_config(sys.argv[1]).append_connectors(connectors)
```
jiridanek, line 287: Should this feature be (eventually) moved into skuperrouter itself?

scripts/configure.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
```
jiridanek, line 2: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

jiridanek (commented):
I know the PR is WIP, so don't take my comments to seriously if I commented on something prematurely. Thanks.

jiridanek:
@mcressman Would it be possible to copy the qdrouterd RPM spec file directly into this repo, so that rpmbuild can be used to do builds?

.github/scripts/Dockerfile
fgiorgetti: Agreed. I will try it out and evaluate.

.github/scripts/builder-compile.sh
fgiorgetti: ... then I can get rid of these, right: `-DBUILD_CPP=OFF -DBUILD_RUBY=OFF -DBUILD_GO=OFF` ?

Dockerfile
fgiorgetti: This is no longer an apache related build. Therefore I was going to keep just the "existing"
Dockerfile used to build the upstream image which is currently used by Skupper.

Any special reason to keep this one?

scripts/auto_mesh.py
fgiorgetti: True. I will update these python scripts as well.

.github/scripts/builder-compile.sh
jiridanek: yes, that is the idea; GitHub CI does it this way too, https://github.com/skupperproject/skupper-router/blob/main/.github/workflows/build.yaml#L42-L49

Dockerfile
jiridanek: If the skupper image is going to use quay.io/centos/centos:stream8, then no reason to keep this. Other Dockerfiles in https://github.com/skupperproject/skupper-router/tree/main/dockerfiles can also be purged, I think. But that raises the question whether it should be done in this PR, or in subsequent PR later.

scripts/auto_mesh.py
fgiorgetti: It is possible. After the router is started, Skupper already maintains connectors properly.
However these scripts are used only to prepare the config for the router startup, before qdrouterd is executed.

scripts/configure.sh
fgiorgetti: +1 Although I would leave `-x` out of the scope here to keep pod logs cleaner.

scripts/auto_mesh.py
jiridanek: Since skupper-router is going to be exclusively in skupper, then making the option 'QDROUTERD_AUTO_MESH_DISCOVERY' be qdrouterd's option sounds sensible to me. It's an enhancement idea, however; not relevant for this PR right now.

fgiorgetti: This is certainly something we can discuss and re-evaluate.

.github/scripts/Dockerfile
fgiorgetti: Resulting image when using CentOS 8 Stream had ~ 800MB while with Fedora 34 we got ~ 450MB.
@kgiusti @ganeshmurthy any thoughts?

.github/scripts/builder-compile.sh
fgiorgetti: I have also replaced `-DBUILD_TLS=ON` with `-DSSL_IMPL=openssl` as the former is no longer a valid option.

Dockerfile
fgiorgetti: I will remove it now, but the others under dockerfiles will be left for an upcoming PR.

.github/scripts/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ENV PROTON_SOURCE_URL=${PROTON_SOURCE_URL:-http://archive.apache.org/dist/qpid/proton/${PROTON_VERSION}/qpid-proton-${PROTON_VERSION}.tar.gz}
+RUN .github/scripts/compile.sh
+
+
+FROM fedora:34
+
+# FROM quay.io/centos/centos:stream8
+# RUN dnf -y install epel-release && dnf config-manager --set-enabled powertools && dnf -y install glibc cyrus-sasl-lib cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 libuuid openssl gettext hostname iputils libwebsockets-devel libnghttp2 gdb valgrind && dnf -y update && dnf clean all
+
+RUN dnf -y install glibc cyrus-sasl-lib cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 libuuid openssl gettext hostname iputils libwebsockets-devel libnghttp2 gdb valgrind && dnf -y update && dnf clean all
```
jiridanek, line 39: valgrind? I understand that the other packages are there to simplify debugging (gdb), but valgrind?

libwebsockets-devel? Why install a devel package?

cyrus-sasl-md5? Insecure! Burn it, cleanse it with fire!

.github/scripts/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# under the License.
+#
+
+FROM fedora:34 as builder
+
+# FROM quay.io/centos/centos:stream8 as builder
+# Enable additional package repositories for CentOS
+#  note: PowerTools is called CodeReady Linux Builder in RHEL 8
+# RUN dnf -y install epel-release && dnf config-manager --set-enabled powertools && dnf -y install gcc gcc-c++ make cmake cyrus-sasl-devel cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 openssl-devel libuuid-devel nodejs swig wget patch findutils git valgrind libwebsockets-devel python3-devel libnghttp2-devel && dnf clean all -y
+RUN dnf -y install gcc gcc-c++ make cmake cyrus-sasl-devel cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 openssl-devel libuuid-devel nodejs swig wget patch findutils git valgrind libwebsockets-devel python3-devel libnghttp2-devel && dnf clean all -y
```
jiridanek, line 26: cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5? These are needed for running tests, should not be needed for build.

nodejs? Not needed, we don't have webconsole anymore.

valgrind? Not needed here.

.github/scripts/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+WORKDIR /build
+COPY . .
+ENV PROTON_VERSION=0.36.0
+ENV PROTON_SOURCE_URL=${PROTON_SOURCE_URL:-http://archive.apache.org/dist/qpid/proton/${PROTON_VERSION}/qpid-proton-${PROTON_VERSION}.tar.gz}
+RUN .github/scripts/compile.sh
+
+
+FROM fedora:34
```
jiridanek, line 34: @kgiusti @ganeshmurthy Is there any reason why our libwebsockets (in downstream) is built with `LWS_WITH_LIBUV=ON` and `LWS_WITH_HTTP2=OFF`? Thing is, libwebsockets has multiple backend options, libuv is only one of multiple event loops possible. Why was this one chosen?

(I tried asking for libwebsockets be added to ubi8 image, but I got rebuffed, since the library is not part of standard rhel8 either. https://bugzilla.redhat.com/show_bug.cgi?id=2051603)

.github/scripts/Dockerfile
fgiorgetti: Valgrind is actually needed. Skupper has an initialization option to debug using valgrind or gdb, see:
```
--router-debug-mode string             Enable debug mode for router ('valgrind' or 'gdb' are valid values)
```

There was a reason to keep libwebsockets-devel in the final image, but I do not recall now (I will retest without it).
@ganeshmurthy if you remember why, let me know. Otherwise I will remove and validate if it is really needed.

@grs @ganeshmurthy can we get rid of cyrus-sasl-md5? Any comment?

.github/scripts/Dockerfile
fgiorgetti: Those cyrus-sasl-* were useful in the former upstream image, but I believe we can get rid of them now.
Same for nodejs, since there is no more console here.

.github/scripts/Dockerfile
mcressman: Yes, we will continue static compile downstream.  Downstream is useful because downstream product releases for clients/proton might be only twice a year and we may need to use a more recent version.  Upstream is optional, but as you suggest, doing the same upstream and downstream is better.  I'd say do the static compile upstream for now and see if we run into any issues.

.github/scripts/Dockerfile
jiridanek: > `--router-debug-mode string`

Whoa! That IMO deserves an explanatory comment in the Dockerfile. Actually, the deps list is somewhat opaque: something is there for proton, something for router, something for skupper's way of dealing with the router.

scripts/launch.sh
```diff
@@ -0,0 +1,39 @@
...
+
+if [ -f $CONFIG_FILE ]; then
+    ARGS="-c $CONFIG_FILE"
+fi
+
+if [[ $QDROUTERD_DEBUG = "gdb" ]]; then
+    exec gdb -batch -ex "run" -ex "bt" --args qdrouterd $ARGS
+elif [[ $QDROUTERD_DEBUG = "valgrind" ]]; then
+    exec valgrind qdrouterd $ARGS
+else
```
jiridanek, line 37: Ha, here's the `--router-debug-mode`.

Few thoughts:

* wouldn't `catchsegv` do the backtrace printing just as well as gdb, only with less dependencies?
* or possibly `SEGFAULT_SIGNALS="all" LD_PRELOAD=/lib64/libSegFault.so`
* it would be probably also beneficial to compile with `-fstack-protector-all` and `-D_FORTIFY_SOURCE=2`, I sometimes saw glibc print stacktrace on its own
* alternatively, what about compiling with `-fsanitize=address`? That adds runtime checks and also a backtrace printing handler.

scripts/configure.sh
jiridanek: Ok. I prefer to see the commands being executed, but if it's not necessary, then fine.

Just keep in mind that some perfectly legal bash scripts need to be fixed after you add the `set`. E.g. if the script used undefined variable, expecting it will evaluate to empty string, that will no longer work now.

scripts/launch.sh
jiridanek: * `LIBC_FATAL_STDERR_=1`

.github/scripts/Dockerfile
jiridanek: @mcressman Thanks for the info! Would it be possible to opensource the .spec file that you maintain (and put a copy of it to this repo, so it can be directly used to do builds?)

.github/scripts/Dockerfile
ganeshmurthy: I was looking here - https://github.com/interconnectedcloud/qdr-image/blob/master/Dockerfile and that Dockerfile is using Fedora. In the past, was the image created from that Dockerfile ?

Dockerfile
ganeshmurthy: I agree, no reason to keep this

.github/scripts/Dockerfile
ganeshmurthy: I believe cyrus-sasl-md5 can be removed.

.github/scripts/Dockerfile
fgiorgetti: That is correct Ganesh. The former image was using Fedora 30. Some deps like libnghttp2 need at least Fedora 33
to be used. But since we now require CMake 3.20, I had to bump it to Fedora 34.

ganeshmurthy: I meant that since the former image was using Fedora, there is no harm in the current image using Fedora as well. Fedora 34 sounds good.

scripts/launch.sh
grs: Is there any overhead for -fsanitize=address?

Using catchsegv might well be fine (though I think we might want to include gdb anyway). I'd suggest making those changes as separate PRs though, and focus initially on removing the need for the interconnectedcloud image in this one.

.github/scripts/Dockerfile
grs: The cyrus sasl plugins could be useful in the image, but I agree there is no need to have them in the builder.

scripts/launch.sh
jiridanek: > Is there any overhead for -fsanitize=address?

Yes, something in the order of 2x or 4x for both memory and CPU. Plus at the start there is huge virtual memory allocation on the order of 20TB. Only a fraction of memory is not actually in use, but it means that if there are memory limits (ulimit, or k8s limits), they will probably be hit by this.

>  I'd suggest making those changes as separate PRs though

I never said otherwise. It was just that here in the PR all the image code is plain to see and easy to comment on, so that's what I was doing.

jiridanek (approved):

scripts/launch.sh
jiridanek: -fstack-protector-all and -D_FORTIFY_SOURCE=2 have much less overhead; and afaik these are default cflags on ubuntu, and definitely when building rpm packages

.github/scripts/Dockerfile
grs: Actually at least cyrus-sasl-plain is required at present.

fgiorgetti: Do you mean in the builder? The console has been removed, so do you think it should be kept in the final image?

grs: If the console is gone then you are right, the existing use case is gone. We can always add it when needed.

scripts/launch.sh
jiridanek: follow up issue https://github.com/skupperproject/skupper-router/issues/60

scripts/auto_mesh.py
jiridanek: follow up issue https://github.com/skupperproject/skupper-router/issues/60

ganeshmurthy (approved):
Please feel free to commit to main branch

PR #49, opened 2022-02-07
Add workaround for Pytest 7.0.0 stacktrace normalization bug

jiridanek:
Workarounds https://github.com/pytest-dev/pytest/issues/9610

### DIFF
--- a/tests/http1_tests.py
+++ b/tests/http1_tests.py
@@ -148,8 +148,10 @@ class MyHTTPServer(HTTPServer):
         self._client_sockets.append(sock)
         return sock, addr
 
-    def shutdown_request(self, request: socket.socket) -> None:
+    def shutdown_request(self, request) -> None:
         """Forget the client socket"""
+        assert isinstance(request, socket.socket)
+
         self._client_sockets.remove(request)
         super().shutdown_request(request)  # type: ignore[misc]  # error: "shutdown_request" undefined in superclass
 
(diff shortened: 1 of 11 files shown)


jiridanek:
pytest 7.0.1 came out today, containing a fix

jiridanek:
This is no longer necessary. I still think that the changes in the PR are nice and helpful, but now it is not fixing anything. I'll close this in a few days if there is no feedback.

kgiusti:
The changes LGTM - one question: these changes imply that TestCases should call setUp() instead of using an __init__ method.  So going forward should we avoid using __init__ in favor of setUp?  thanks.

kgiusti (approved):
The changes do improve the code regardless.

jiridanek:
> The changes LGTM - one question: these changes imply that TestCases should call setUp() instead of using an **init** method. So going forward should we avoid using **init** in favor of setUp? thanks.

I think so. Defining `__init__` in TestCase instances is unusual. I wouldn't do it, unless there is a compelling reason. Using `setUp` when possible avoids having to deal with the `__init__`'s optional argument, which is a distraction (and scause of the original Pytest problems here).

Going further forward, I'd consider using Pytest ways of doing things, exclusively, with the proviso that some custom work is needed to make integration/system testing convenient (https://github.com/pytest-dev/pytest/issues/9141, but current python unittest suffers from the same problem, or lack of feature, to be precise.)

PR #50, opened 2022-02-07
Initial checkin for shutdown doc.

mgoulish:


docs/notes/shutdown.txt
```diff
@@ -0,0 +1,68 @@
...
+
+qdrouterd shutdown procedure
+============================
+
+
+How does the router shut down when interrupted with ctrl-C ?
+
+
+  1. ctrl-c is caught by server.c::handle().
+     it comes packaged by proton as PN_PROACTOR_INTERRUPT.
```
jiridanek, line 30: Ctrl-c, i.e. SIGINT, is caught in main.c::signal_handler(). The handler calls qd_server_stop(). That results in delivering an interrupt into the Proton proactor. When proactor gets around to noticing that happened, it will emit `PN_PROACTOR_INTERRUPT`, which gets handled in server.c::handle().

docs/notes/shutdown.txt
```diff
@@ -0,0 +1,68 @@
...
+
+  2. The false return from handle() is received by thread_run().
+     If there is a container, thread_run() calls 
+     qd_conn_event_batch_complete() to let the container know 
+     that the work batch is finished.
+
+      2.1 qd_conn_event_batch_complete() loops through the 
+          free_link_session_list for this container. 
+          If the connection is not already closed, it frees 
+          each session's pn_link and pn_session, if they exist.
```
jiridanek, line 44: I don't think this is strictly related to shutdown. This happens every time the loop runs through, during regular operation.

docs/notes/shutdown.txt
```diff
@@ -0,0 +1,68 @@
...
+
+
+  3. Once all threads have exited from thread_run(), 
+     qd_server_run() logs a "Shut Down" message, and returns to 
+     main_process() in main.c .
+
+
+  4. main_process() sets to NULL the dispatch pointer that 
+     everybody was using, and then frees the structure.
```
jiridanek, line 53: Not everybody. The pointer has only a few users. Mostly, the pointer value is passed through parameters or in different data structures.

jiridanek, line 53: > and then frees the structure

This is the most "interesting" part. Do tell more!

jiridanek, line 53: > and then frees the structure

This is the most "interesting" part. Do tell more!

docs/notes/shutdown.txt
mgoulish: excellent -- thanks!

docs/notes/shutdown.txt
```diff
@@ -0,0 +1,68 @@
...
+# under the License.
+#
+
+
+
+qdrouterd shutdown procedure
+============================
+
+
+How does the router shut down when interrupted with ctrl-C ?
```
jiridanek, line 26: This probably needs some kind of overview section: what are the things that actually need to get shut down? In what order?

How do you shut down a Proton proactor loop? Why can't `main.c::signal_handler()` simply call `pn_proactor_interrupt()` in a loop as many times as there are threads? Why doesn't Proactor have a way to kill all threads at once?

How do you shut down memory pool? How do you shut down Python interpreter. How do you shut down libwebsockets and libnghttp2 listeners?

Can you do graceful shutdown, quiescing? Router will stop receiving new connections and will stop when last connection disconnects? Afaik no, the router shutdown will forcibly kill all connections.

docs/notes/shutdown.txt
mgoulish: There are two calls to that fn -- in two nested loops that are both conditioned on the variable 'running'.   I was looking at the call that is only in the outer loop, which only happens if 'running' comes back false when handle() is called.
The only difference in the calls is that the final arg,  'conn_closed', is false in the second call.

docs/notes/shutdown.txt
mgoulish: Yes, I will change this.
It is 'everybody' for very small values of 'everybody'.
Like when I say, "Hey everybody, listen up!"   I don't actually mean *everybody*.   :-)

mgoulish: You think there should be more detail here?

docs/notes/shutdown.txt
jiridanek: Yeah. Four calls, actually. Two in server.c (of which I only noticed the one), and also two more in container.c.

docs/notes/shutdown.txt
mgoulish: I bet it's easy to see bugs with those big glasses.

docs/notes/shutdown.txt
jiridanek: Well, yeah. `qd_dispatch_free` does a lot, and it calls functions that do a lot.

docs/notes/shutdown.txt
jiridanek: I meant you should have something high-level about what needs to happen so that the router shutdowns.

* stop proactor,
* close connections,
* stop adaptors (yeah, you have nothing about adaptors here, and they do have their own lifecycle),
* and so on

Maybe also _why_ is it happening. But that can be another docs, further on the route towards fixing, yeah.

> In this document, I would rather only describe what is happening.

+1, I am all for gradual staged approach

docs/notes/shutdown.txt
```diff
@@ -0,0 +1,68 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+
+
+qdrouterd shutdown procedure
+============================
```
jiridanek, line 23: File suffix should be .adoc, because we all like Asciidoc! (The other PR from Ken uses that format.)

docs/notes/shutdown.txt
mgoulish: OK, will do.  Oops. I just assumed it was trivial.  Bad assumption.

docs/notes/shutdown.txt
mgoulish: asciidoctored

Issue #54, opened 2022-02-08
system_tests_delivery_counts failing


ganeshmurthy:
```
44: =================================== FAILURES ===================================
44: __________ OneRouterLinkCountersTest.test_06_large_message_released ___________
44: 44: self = <system_tests_delivery_counts.OneRouterLinkCountersTest testMethod=test_06_large_message_released>
44: 
44:     def test_06_large_message_released(self):
44: >       self.verify_released(True)
44: 44: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_delivery_counts.py:1198: 
44: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
44: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_delivery_counts.py:1059: in verify_released
44:     self.assertEqual(test.receiver_stats['releasedCount'], self.COUNT)
44: E   AssertionError: 29 != 40
44: - generated xml file: /__w/skupper-router/skupper-router/skupper-router/build/tests/junitxmls/system_tests_delivery_counts.xml -
44: =========================== short test summary info ============================
44: FAILED ::OneRouterLinkCountersTest::test_06_large_message_released - Assertio...
44: ======================== 1 failed, 26 passed in 59.43s =========================
23/33 Test #44: system_tests_delivery_counts ...................***Failed   59.76 sec
============================= test session starts ==============================

```

https://github.com/ganeshmurthy/skupper-router/runs/5109959880?check_suite_focus=true#step:27:1023

jiridanek:
DISPATCH issue for porting the fix to dispatch was https://issues.apache.org/jira/browse/DISPATCH-2329

PR #55, opened 2022-02-08
#54: Explicity check for released counts instead of delivery counts o…

ganeshmurthy:
…f deliveries are being released

### DIFF
--- a/tests/system_tests_delivery_counts.py
+++ b/tests/system_tests_delivery_counts.py
@@ -860,12 +860,18 @@ class OneRouterLinkCountersTest(TestCase):
                 if rx_limit is None else rx_limit
 
             self.sent = 0
+            self.received = 0
             self.timer = 0
             self.poll_timer = None
             self.conn = None
             self.sender_stats = None
             self.receiver_stats = None
             self.large_message = large_message
+            self.reactor = None
+            self.sender = None
+            self.receiver = None
+            self.max_attempts = 10
+            self.num_attempts = 0
 
         def timeout(self):
             self._cleanup()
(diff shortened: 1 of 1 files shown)


tests/system_tests_delivery_counts.py
```diff
@@ -1001,24 +1007,35 @@ def _cleanup(self):
...
             Periodically check the deliveryCount on the receiver.  Once it
             reaches rx_limit the test is complete: gather link statistics
             before closing the clients
             """
             li = get_link_info("Rx_Test01", self.router_addr)
-            if li and li['deliveryCount'] == self.rx_limit:
-                self.receiver_stats = li
-                self.sender_stats = get_link_info("Tx_Test01", self.router_addr)
-                self._cleanup()
+            if li and self.outcome == Delivery.RELEASED and li['releasedCount'] == self.rx_limit:
```
kgiusti, line 1022: Ah - yes this is better... but...
If presettled == false, then the last counter to be updated would be the sender's counters, since it takes time to propagate the outcome back to the sender the sender's outcomes would lag behind the receivers.

Perhaps it would be best to simply keep polling the statistics for both sender and receiver and exit the poll loop once the counters haven't changed since the last poll?

kgiusti (changes requested):
See inline, thanks

kgiusti:
Oopsie - one gotcha: if comparing the statistics maps you'll probably want to filter out all the non-statistics map entries before the compare.  
list_o_statistics = ["deliveryCount", etc, etc]
stats_dict  = {key: value for key, value in link_counters.items() if key in list_o_statistics}  would do the trick I think....

kgiusti (approved):

PR #56, opened 2022-02-09
ISSUE-45 Updated MAU format to carry multiple data points per address.

ted-ross:
This breaking change to the routing protocol enables future expansion of routing capabilities.

### DIFF
--- a/src/router_core/modules/mobile_sync/mobile.c
+++ b/src/router_core/modules/mobile_sync/mobile.c
@@ -31,7 +31,7 @@
 #include <inttypes.h>
 #include <stdio.h>
 
-#define PROTOCOL_VERSION 1
+#define PROTOCOL_VERSION 2
 static const char *OPCODE     = "opcode";
 static const char *MAR        = "MAR";
 static const char *MAU        = "MAU";
@@ -39,7 +39,6 @@ static const char *ID         = "id";
 static const char *PV         = "pv";
 static const char *AREA       = "area";
 static const char *MOBILE_SEQ = "mobile_seq";
-static const char *HINTS      = "hints";
 static const char *ADD        = "add";
 static const char *DEL        = "del";
 static const char *EXIST      = "exist";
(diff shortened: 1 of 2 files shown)


src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -191,15 +190,66 @@ static qd_composed_field_t *qcm_mobile_sync_message_headers(const char *address,
...
+        qd_compose_insert_int(field, addr->treatment);
+        qd_compose_insert_int(field, DEQ_SIZE(addr->inlinks));
+    }
+    qd_compose_end_list(field);
+}
+
+
+static qd_iterator_t *qcm_mobile_sync_parse_addr_descriptor(qd_parsed_field_t *field, int *treatment, int *inlink_count)
+{
+    qd_iterator_t *iter = 0;
```
ganeshmurthy, line 218: As a first line of code, can we please check for a non-zero field like we do here - https://github.com/apache/qpid-dispatch/blob/main/src/parse.c#L359 or like we do in the qcm_mobile_sync_router_by_id() function

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -191,15 +190,66 @@ static qd_composed_field_t *qcm_mobile_sync_message_headers(const char *address,
...
+
+    if (!qd_parse_is_list(field)) {
+        return iter;
```
ganeshmurthy, line 221: Also, return zero here instead of iter and declare the iter after this check instead of at the beginning of the function

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -191,15 +190,66 @@ static qd_composed_field_t *qcm_mobile_sync_message_headers(const char *address,
...
+ */
+static void qcm_mobile_sync_compose_addr_descriptor(const qdr_address_t *addr, qd_composed_field_t *field, bool is_added)
+{
+    const char *hash_key = (const char*) qd_hash_key_by_handle(addr->hash_handle);
+
+    qd_compose_start_list(field);
+    qd_compose_insert_string(field, hash_key);
+    if (is_added) {
+        qd_compose_insert_int(field, addr->treatment);
+        qd_compose_insert_int(field, DEQ_SIZE(addr->inlinks));
```
ganeshmurthy, line 210: What happens if the number of addr->in_links changes frequently?

src/router_core/modules/mobile_sync/mobile.c
jiridanek: so, `assert` then?

ganeshmurthy (approved):

PR #61, opened 2022-02-11
DISPATCH-2323 Try binding to a port before handing it out as "free"

jiridanek:


tests/system_test.py
```diff
@@ -173,19 +173,37 @@ def get_local_host_socket(socket_address_family='IPv4'):
     return s, host
 
 
-def port_available(port, socket_address_family='IPv4'):
+def port_refuses_connection(port, socket_address_family='IPv4'):
```
ganeshmurthy, line 176: I would call this function -  check_port_refuses_connection

tests/system_test.py
```diff
@@ -173,19 +173,37 @@ def get_local_host_socket(socket_address_family='IPv4'):
...
+        return e.errno == errno.ECONNREFUSED
+    finally:
+        s.close()
+
+    return False
+
 
-    s.close()
-    return available
+def port_permits_binding(port, socket_address_family='IPv4'):
```
ganeshmurthy, line 189: I would call this is_port_permits_binding

ganeshmurthy (approved):

tests/system_test.py
```diff
@@ -173,19 +173,37 @@ def get_local_host_socket(socket_address_family='IPv4'):
...
+        s.bind((host, port))
+    except OSError:
+        return False
+    finally:
+        s.close()
+
+    return True
+
+
+def port_available(port, socket_address_family='IPv4'):
```
ganeshmurthy, line 204: I really like the old school convention of is_port_available() but I guess PEP-8 does not lay down any specific rules

jiridanek: I guess the is/can/has convention is a good one. I'll rename the methods here to use it.

tests/system_test.py
jiridanek: I did not go with `check_port_permits_connection` in the end, because the purpose of the function is to check for `errno.ECONNREFUSED`.

tests/system_test.py
jiridanek: renamed

ganeshmurthy: Please feel free to merge. Approved !

PR #64, opened 2022-02-13
Speed up CI by compiling with -j6 and run tests with -j100 or -j12

jiridanek:
So, @ted-ross says `-j12`? Phew! :100: 

I am only half-serious about this, but there actually is some merit to these crazy parallel settings.

The downsides are obvious: tests fail more and logs are a mess.

But the upside is that the same tests tend to fail more, over and over again, so they show themselves as the least reliable tests from the whole bunch, and can be fixed. Also, the CI does run a little bit faster, though not by all that much.

The -j6 for compilations is interesting, though. Most of the compilation is essentially fetching stuff from ccache, so there it does make sense to initiate more fetches in parallel. When an actual compilation needs to happen, it will be arrived at faster. Dispatch will "compile" this way in about a second!.

Edit: looking at the durations also shows my folly in doing the split compile and test jobs for Ubuntu. The compressing of the artifacts takes longer than ccache restore and the actual build.

Edit Edit: the natural continuation in this direction would be to run every test multiple times

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -123,7 +123,7 @@ jobs:
             ${ProtonCMakeExtraArgs}
 
       - name: qpid-proton cmake build/install
-        run: cmake --build "${ProtonBuildDir}" --config ${BuildType} -t install
+        run: cmake --build "${ProtonBuildDir}" --config ${BuildType} -t install --parallel 6
 
       - name: Display ccache stats
         run: ccache -s
@@ -139,7 +139,7 @@ jobs:
             ${RouterCMakeExtraArgs}
 
       - name: skupper-router cmake build/install
-        run: cmake --build "${RouterBuildDir}" --config ${BuildType} -t install
+        run: cmake --build "${RouterBuildDir}" --config ${BuildType} -t install --parallel 6
 
       - name: Display ccache stats
         run: ccache -s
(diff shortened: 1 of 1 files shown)


kgiusti (commented):
I think the motivations behind this patch are very valid - it obviously seems to be flushing out timing-related bugs/test issues.

However I wouldn't check this in until _after_ those tests/bugs are fixed and this patch can run successfully.  Otherwise ctest will always just fail which will make Ganesh's head explode.

To be fair those test failures/router bugs should be considered high priority so we *can* get this patch in sooner rather than later.

Just my $0.02 - I'd like to hear @ted-ross and @ganeshmurthy opinions on this as well.

ganeshmurthy:
@jiridanek can you please rebase this PR against master ? Thanks.

ted-ross:
I test using -j12 because that's a good match for the number of CPUs available on my system (8).  If you're looking for speed, you need to tune to the available CPUs.  I agree that using parallelism to flush out bugs is also valuable.

.github/workflows/build.yaml
```diff
@@ -222,7 +222,7 @@ jobs:
         working-directory: ${{env.RouterBuildDir}}
         run: |
           ulimit -c unlimited
-          ctest --timeout 1200 -C ${BuildType} -V -T Test --no-compress-output -I ${{matrix.shard}},,${{matrix.shards}} -j2 ${{env.RouterCTestExtraArgs}}
+          ctest --timeout 1200 -C ${BuildType} -V -T Test --no-compress-output -I ${{matrix.shard}},,${{matrix.shards}} -j100 ${{env.RouterCTestExtraArgs}}
```
ganeshmurthy, line 225: I think we should set this to j12, not j100

jiridanek: Ubuntu GHA jobs seem significantly more stable than the Fedora ones. Partially it is because Fedora runs TSan, which check has high overhead.

Starting all ctest targets at once (which this in effect does) results only in small number of test failures.

ganeshmurthy (approved):

PR #82, opened 2022-02-17
DISPATCH-2272 Mark all functions called from Python with QD_EXPORT

jiridanek:


kgiusti (approved):

jiridanek:
I've (accidentally) found another context where this is necessary and I've found that there are some unmarked function that I'll have to add to the PR

The uses for `QD_EXPORT` are

* when compiling in context where not all symbols are exported, which is the default situation on Windows, and can be configured in GCC
* when compiling the code with g++, because then function names get mangled using the c++ conventions to support (possible) function overloading (that is the new context)

Testing the PR in one of these two ways is necessary to reliably reveal missing `QD_EXPORT`s.

PR #84, opened 2022-02-17
Base the Docker image on ubi8 image instead of fedora

jiridanek:
The only change I needed is that I had to do

    dnf -y --setopt=tsflags=nodocs update

as the first thing. Other than that, I only change base image and add some centos repositories for packages that ubi lacks.


fgiorgetti (approved):
+1

Issue #60, opened 2022-02-10
Make CI builds (especially skupper-router image) more production-like


jiridanek:
This is a follow up to (my) comments on https://github.com/skupperproject/skupper-router/pull/44. This ticket suggests to consider doing (and eventually do) one or more of the following

* [x] base the skupper-router image on CentOS Stream 8 or UBI 8 https://github.com/skupperproject/skupper-router/pull/84
* [x] build the router with production-like flags, such as the various flavors of `-fstack-protector` and with `-D_FORTIFY_SOURCE=2`
   * [x] https://github.com/skupperproject/skupper-router/issues/444
* [ ] use RPM .spec file to build the router in the builder image and then install it from that RPM in the deployed image
   * [x] https://github.com/skupperproject/skupper-router/issues/186
   * [x] WON'T DO, the production image is not using rpm either
* [x] build qpid-proton as a static library and link it into dispatch; possibly use `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` and `-fno-semantic-interposition` (https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup)
   * https://github.com/skupperproject/skupper-router/pull/96
* use catchsegv or `-lSegFault` in production build to always dump stacktrace on crash (use `addr2line` to later decrypt the trace)
* `LIBC_FATAL_STDERR_=1`
* why our libwebsockets is built with `LWS_WITH_LIBUV=ON` and `LWS_WITH_HTTP2=OFF`?
   * [x] fixed by @ajssmith in https://github.com/skupperproject/skupper-router/pull/369
* `set -Eeuo pipefail`
* `QDROUTERD_AUTO_MESH_DISCOVERY` as core skupper-router feature, instead of hacky config file editing during image startup
* some tests for the container image, to be reasonably sure Skupper was not broken by changes
   * [ ] https://github.com/skupperproject/skupper-router/pull/169
   * [x] https://github.com/skupperproject/skupper-router/pull/381

ganeshmurthy:
Most tasks completed

PR #96, opened 2022-02-23
Issue #60: Build skupper-router image using LTO and with static proton libs

jiridanek:
@fgiorgetti Is there a good way I can run skupper tests using this image, before I unmark the PR as a draft? Thanks.


.github/scripts/compile.sh
```diff
@@ -42,13 +42,19 @@ tar -zxf qpid-proton.tar.gz -C qpid-proton-src --strip-components 1
...
+    && VERBOSE=1 make DESTDIR=$WORKING/proton_install install \
     && tar -z -C $WORKING/proton_install -cf /qpid-proton-image.tar.gz usr \
-    && make install
+    && VERBOSE=1 make install
 cd $WORKING/build
-cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_LIBWEBSOCKETS=ON -DCMAKE_INSTALL_PREFIX=/usr $WORKING/ \
-    && make  \
-    && make DESTDIR=$WORKING/staging/ install \
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+  -DCMAKE_C_FLAGS='-fno-semantic-interposition' -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON \
```
jiridanek, line 55: This might be premature to consider for production build, but I want to wave it around for general consideration. Works for me.

jiridanek: `-fno-semantic-interposition` afaik precludes doing `LD_PRELOAD=/usr/lib64/libjemalloc.so`, which is another performance-enhancing option (with significant performance impact on my laptop)... Yeah, probably not touch it until the best combination of these options is settled.

.github/scripts/compile.sh
```diff
@@ -42,13 +42,19 @@ tar -zxf qpid-proton.tar.gz -C qpid-proton-src --strip-components 1
 do_patch "patches/proton" qpid-proton-src
 
 cd proton_build
-cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_BINDINGS=python -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON -DSSL_IMPL=openssl $WORKING/qpid-proton-src/ \
-    && make \
-    && make DESTDIR=$WORKING/proton_install install \
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+  -DENABLE_LINKTIME_OPTIMIZATION=ON \
+  -DSSL_IMPL=openssl -DBUILD_STATIC_LIBS=ON -DBUILD_BINDINGS=python -DSYSINSTALL_PYTHON=ON \
```
jiridanek, line 47: Static proton libs don't get compiled with LTO. I think this is a deliberate decision (by @astitcher, or maybe Alan, back then) and not an omission. The problem with LTO-ing static libraries is that you have to replace `ar` and `runlib` with `gcc-ar` or `gcc-runlib`, and there is something equivalent for clang... Or you can try to use CMake facilities if you have new-enough CMake (https://cmake.org/cmake/help/latest/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html#variable:CMAKE_INTERPROCEDURAL_OPTIMIZATION).

In any case, it is not exactly simple, then.

jiridanek: @astitcher Do you maybe remember the details about ^^^? If the `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` behaves better, would you be for enabling LTO that way?

Btw, new Fedora (and maybe RHEL 9) enable LTO by default when compiling RPM packages, e.g.

```
+ CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64  -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'
```

from https://download.copr.fedorainfracloud.org/results/jdanek/skupper-router/fedora-36-x86_64/03946420-skupper-router/builder-live.log.gz (This is Fedora 36)

jiridanek:
Performance comparison. On my laptop, after doing

```
echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo
cpupower frequency-set -u 1000mhz
cpupower frequency-set -d 1000mhz
tuned-adm profile latency-performance
```

and running quiver with default settings for 30 s, repeated 9 times

# Without LTO on Router

```
Estimated message rate is 53938.7 +- 507 msgs/second at 95% confidence
```

# With LTO on Router (`-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`)

```
Estimated message rate is 57101.0 +- 630 msgs/second at 95% confidence
```

When repeated 500 times to get histogram,

image

In short, seems to me that LTO is the future and should be enabled. The disadvantage is a longer build time for the image, which can be compensated for by skipping test targets build.

jiridanek:
I got an informal review on the first commit in this two-commit PR from Fernando in chat. My take is that he was not offended by anything in there, but he was not eager on getting this merged right away either.

.github/scripts/compile.sh
jiridanek: > `-fno-semantic-interposition` afaik precludes doing `LD_PRELOAD=/usr/lib64/libjemalloc.so`

That's nonsense, the option works completely differently from what I thought. In any case, I saw no improvement from `-fno-semantic-interposition`, so I removed it from PR.

ganeshmurthy:
@kgiusti Can you please take a look at this when you have a few mins. I don't see any issues with this as far as I can tell.

.github/scripts/compile.sh
jiridanek: `CMAKE_INTERPROCEDURAL_OPTIMIZATION` does indeed behave better, the winning compilation options are

```
-DCMAKE_POLICY_DEFAULT_CMP0069=NEW -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON 
```

PR #99, opened 2022-02-23
Fixed #98: Reduced sleep times so test executes faster

### LINKED ISSUE #98: Reduce sleep times in system_tests_http2
Do this so the test executes faster

ganeshmurthy:


tests/system_tests_http2.py
```diff
@@ -255,7 +255,7 @@ def check_listener_delete(self, client_addr, server_addr):
         # Add back the listener and run a curl command to make sure that the newly added listener is
         # back up and running.
         create_result = qd_manager.create("org.apache.qpid.dispatch.httpListener", self.http_listener_props)
-        sleep(2)
+        sleep(1)
```
jiridanek, line 258: I really don't like magic timeouts. What about something like this. Some retry method, not sure about precise name, which retries the assertion until either time runs out (then it raises the last error) or until assertion eventually succeeds, in which case it will just move on

```python
def complex_assertion():
    out = self.run_curl(client_addr)
    self.assertIn(ret_string, out)

system_tests.repeat_until(timeout=10, wait=1, f = test(), tolerated_exception=AssertionError)
```

tests/system_tests_http2.py
```diff
@@ -548,7 +548,7 @@ def setUpClass(cls):
         cls.router_qdra.wait_router_connected('QDR.B')
         cls.router_qdrb.wait_router_connected('QDR.A')
 
-        sleep(2)
+        sleep(1)
```
jiridanek, line 551: I recall some tests were sending a probe message through the entire router network to check it is established... or they were checking qdstat. Afaik the preferred way was the probe message.

ganeshmurthy: This sleep is there because there is really no way to find out if the connector made a connection to the http2 server. But now that I think about it, the http2 server connection is the first connection that the router attempts and then it moves to the establish the inter-router connector. I am taking out these sleeps.

tests/system_tests_http2.py
```diff
@@ -297,15 +296,16 @@ def check_connector_delete(self, client_addr, server_addr):
...
-        sleep(2)
-
         # Now, run a curl client GET request with a timeout
         request_timed_out = False
-        try:
-            out = self.run_curl(client_addr, timeout=5)
-            print(out)
-        except Exception as e:
-            request_timed_out = True
+        for n in range(3):
```
jiridanek, line 301: looks good, but I think you should first merge https://github.com/skupperproject/skupper-router/pull/101 and then rebase the pr, just to be sure

kgiusti:
IIUC the HTTP2 GHA CI tests are not running, making it unpossible to tell if this patch affects the tests it intends to modify.
@ganeshmurthy can you please mark this PR as a draft until HTTP2 GHA CI tests are running?  Thanks.

ganeshmurthy:
Marked this as draft. Thanks.

PR #101, opened 2022-02-24
Test: Added libnghttp2-devel

ganeshmurthy:


.github/workflows/build.yaml
```diff
@@ -353,7 +353,7 @@ jobs:
       - name: Install Linux build dependencies
         if: ${{ runner.os == 'Linux' }}
         run: |
-          dnf install -y gcc gcc-c++ cmake libuuid-devel openssl-devel cyrus-sasl-devel cyrus-sasl-plain swig python3-devel python3-pip make libwebsockets-devel libnghttp2-devel ccache libasan libubsan libtsan
+          dnf install -y gcc gcc-c++ cmake libuuid-devel openssl-devel cyrus-sasl-devel cyrus-sasl-plain swig python3-devel python3-pip make libwebsockets-devel libnghttp2-devel ccache libasan libubsan libtsan curl
```
jiridanek, line 356: Curl is already installed on line 435, afaik.

src/message.c
```diff
@@ -2631,8 +2636,22 @@ void qd_message_stream_data_release(qd_message_stream_data_t *stream_data)
...
     //
+    qd_buffer_t *start_buf;
+    if (stream_data->free_prev) {
+        size_t protected_buffer_count = content->protected_buffers;
+        start_buf = DEQ_HEAD(content->buffers);
+        while (protected_buffer_count > 0) {
+            start_buf = DEQ_NEXT(start_buf);
+            protected_buffer_count -= 1;
+            if (start_buf == 0)
+                break;
```
kgiusti, line 2647: Can you log the values of start_buf, stream_data->location.buffer and stream_data->location.buffer->prev?  And stream_data->prev?  All pointers.

ganeshmurthy: Did you mean start_buf, stream_data->section.buffer and stream_data->section.buffer->prev ?

kgiusti: Ah, I don't think this will work in the case of multicast (fanout > 1).
Imagine the following: there are two outgoing messages sharing the same content.  The first outgoing message is serviced first.  It parses a body_data, frees it, then parses the next body data and frees that.  The second free will result in decrementing the refcounts for the previously freed data_body, which is wrong because those refcounts belong to the second outgoing message!

PR #106, opened 2022-02-27
DISPATCH-2321 Add type annotations divined by MonkeyType tool

jiridanek:


python/qpid_dispatch_internal/management/config.py
```diff
@@ -40,7 +45,12 @@
 class Config:
     """Load config entities from qdrouterd.conf and validated against L{QdSchema}."""
 
-    def __init__(self, filename=None, schema=QdSchema(), raw_json=False):
+    def __init__(
+            self,
+            filename: Optional[str] = None,
+            schema: QdSchema = QdSchema(),
```
kgiusti, line 51: @jiridanek Not a comment on the patch since the original had this behavior, but shouldn't we avoid instantiating QdSchema() in the argument list?  IIUC this causes the schema to be instantiated at import time rather than on the call.  I ask because better pythonistas than me have told me that doing stuff like foo(listarg=list()) is bad.

Would it be better to mark schema as Optional[QdSchema] = None and instantiate the schema in the method body itself?

python/qpid_dispatch_internal/management/config.py
```diff
@@ -260,14 +274,19 @@ def dump_json(self, title, js_text):
 
 
 class PolicyConfig(Config):
-    def __init__(self, filename=None, schema=QdSchema(), raw_json=False):
+    def __init__(
+            self,
+            filename: Optional[str] = None,
+            schema: QdSchema = QdSchema(),
```
kgiusti, line 280: Same comment re: schema here.

kgiusti (approved):
+1 All hail type annotations!

python/qpid_dispatch_internal/management/config.py
jiridanek: Yes, all true. Furthermore, if anything modifies that instance of QdSchema, then, because there is only one instance, everybody sees the change, past and future calles alike. There should be a pylint check that checks for this. I'll create issue to re-enable check and fix this.

```python
"""
The below prints:

import time
instance of A created
everything defined
42
24
"""

print("import time")


class A:
    def __init__(self):
        self.data = 42
        print("instance of A created")


def get_a(a: A = A()):
    return a


print("everything defined")

a1 = get_a()
print(a1.data)  # prints 42
a1.data = 24

a2 = get_a()
print(a2.data)  ## prints 24
```

jiridanek: https://github.com/skupperproject/skupper-router/issues/142

PR #110, opened 2022-03-01
DISPATCH-2039 Add option -DQD_DISABLE_MEMORY_POOL for CMake build

jiridanek:


cmake/RuntimeChecks.cmake
```diff
@@ -31,6 +31,16 @@
 # This file updates the QDROUTERD_RUNNER and CMAKE_C_FLAGS
 # appropriately for use when running the ctest suite.
 
+# Disabling memory pool turns use-after-poison asan warnings into use-after-free warnings. The latter come with
+# a freeing stacktrace that makes them easier to triage.
+#
+# Safe pointers require memory pool to work (memory may never be relinquished to the OS, otherwise safe pointers may
+# break). Therefore, only disable memory pool for special debugging purposes. Sanitizers minimize memory reuse and
+# make it significantly less likely that a safe pointer breaks.
+set(QD_DISABLE_MEMORY_POOL OFF CACHE STRING "Disables memory pool. Should be only used with asan or msan RUNTIME_CHECK")
```
kgiusti, line 40: @jiridanek any chance this patch can for the 'cmake' command to fail if QD_DISABLE_MEMORY_POOL is set but RUNTIME_CHECK is not asan/msan?

Issue a message(FATAL_ERROR... like done elsewhere in this file.

kgiusti (approved):
See comment (optional suggestion), otherwise LGTM

cmake/RuntimeChecks.cmake
jiridanek: I guess so, this should thwart the unwarry who might want to compile production build with the option, and put them back on the right path. Not that they would not be taught differently soon enough, the safe pointers break a lot, without sanitizers. (increasing the `quarantine_size_mb` sanitizer option helps make the router more stable).

PR #117, opened 2022-03-03
Factor out configuring signal handler into a separate function

jiridanek:
All comments I have are in source in comments

### DIFF
--- a/router/src/main.c
+++ b/router/src/main.c
@@ -39,6 +39,16 @@ static qd_dispatch_t *dispatch = 0;
 static qd_log_source_t *log_source = 0;
 static const char* argv0 = 0;
 
+/**
+ * Configures the handler function. Specify SIG_IGN to ignore incoming signals.
+ */
+static void install_signal_handler(void (*handler)(int))
+{
+    signal(SIGHUP, handler);
+    signal(SIGQUIT, handler);
+    signal(SIGTERM, handler);
+    signal(SIGINT, handler);
+}
 
 /**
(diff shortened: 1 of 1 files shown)


jiridanek:
See https://github.com/skupperproject/skupper-router/pull/140 for more reasonable attempt to fix the same issue.

router/src/main.c
```diff
@@ -39,18 +39,26 @@ static qd_dispatch_t *dispatch = 0;
...
+{
+    signal(SIGHUP, handler);
+    signal(SIGQUIT, handler);
+    signal(SIGTERM, handler);
+    signal(SIGINT, handler);
+}
 
 /**
  * This is the OS signal handler, invoked on an undetermined thread at a completely
  * arbitrary point of time.
```
jiridanek, line 55: Not necessarily. It seems to be possible, with pthreads, to ensure that there is only one thread available for handling signals. The OS will then be forced to deliver signals to it.

Alternative solution is this: the signal handler submits an action to some queue, and the action queue consumer then takes care of handling the signal.

jiridanek:
Fancy signal masking is interesting, and it might be relevant for orderly shutdown, even. See below. But the issue in question has been already resolved by adjusting the test, and that fix is better than what's being done here.

Regarding shutdown, this thing allows us to force the OS to deliver signals to a specific thread, so that we then always know where the signal handler is running. I'm wondering if there is a risk to this: imagine the special thread somehow died without taking the rest of skrouterd down with it; then the other surviving threads will be ignoring SIGTERM?

PR #131, opened 2022-03-05
Fixed #129: Configure coredumping in GHA and print and archive backtraces

### LINKED ISSUE #129: Capture core files from GHA
There is a feeble attempt to upload files with `core` in their name, but that does not actually ever get any core files.

https://github.com/skupperproject/skupper-router/runs/5428604821?check_suite_focus=true#step:9:2719

```
67: c-benchmarks: ../src/posix/threading.c:58: sys_mutex_lock: Assertion `result == 0' failed.
22/34 Test #67: c-benchmarks ............................***Failed   82.45 sec
```

It is likely that `coredumpctl` has to be used. That might interact with Docker in some way... Also, ASAN and TSAN binaries reserve massive amounts of virtual memory (20 TB or so) which may end up disabling core file dumping.

Would be best to create a branch, on the branch intentionally create a crashing binary that produces coredump quickly and then investigate methods of getting the coredump uploaded.

jiridanek:
See the dummy build https://github.com/jiridanek/skupper-router/actions/runs/1939196539 for what this looks like. Fedora stacktraces are the best.

Here is a branch running a genuinely failing qdrouterd. TODO

I dislike writing "special cases" like this in CI scripts because they get rarely triggered and it is possible they will stop working when nothing tests them. Also, it is difficult to edit the script because the turnaround time when any change is only tested by running CI is long.

Hopefully it will work out.


.github/workflows/build.yaml
```diff
@@ -325,7 +334,6 @@ jobs:
       CCACHE_COMPRESS: 'true'
       CCACHE_MAXSIZE: '400MB'
       QPID_SYSTEM_TEST_TIMEOUT: 300
-      QPID_SYSTEM_TEST_SKIP_FALLBACK_SWITCHOVER_TEST: True
```
jiridanek, line 328: todo, did not mean to delete this here, although this env var is obsolete, for previously deleted test

jiridanek:
The example job did fail on assert, but no core was dumped. More work needed. https://github.com/jiridanek/skupper-router/runs/5435028258?check_suite_focus=true#step:9:5723

.github/workflows/build.yaml
jiridanek: fixed

jiridanek:
> The example job did fail on assert, but no core was dumped. More work needed. https://github.com/jiridanek/skupper-router/runs/5435028258?check_suite_focus=true#step:9:5723

The problem was that sanitizers disable coredumping by default on 64bit architectures. I had to explicitly enable it.

kgiusti (approved):
Yes, please.

PR #132, opened 2022-03-06
Fixed #130: Archive only logs from tests that failed in GHA

### LINKED ISSUE #130: Archive only logs from tests that failed in GHA
This is not too hard to do and it will make the process of uploading artifacts more reliable, when there is less to upload. GHA enforces a limit on total size of all artifacts by deleting old artifacts when the limit is exceeded; this task will allow keeping history further back.

jiridanek:


kgiusti (approved):

jiridanek:
This changed the output from CentOS 8 tests. Previously, in the GitHub actions console log, it outputted only from failed CTests, now it shows output from all tests (same as the other CI jobs).

Do we have preference for either one of those ways?

PR #140, opened 2022-03-08
DISPATCH-1689 Wait for startup log line to ensure router is fully started in tests

jiridanek:
I seem to recall that @ChugR had some log-checking router startup PR at some point, for qpid-dispatch. I was not able to find it, so here is my attempt from scratch.

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -30,7 +30,7 @@ Features:
 
 import fcntl
 import pathlib
-from typing import Callable
+from typing import Callable, TextIO
 
 import errno
 import logging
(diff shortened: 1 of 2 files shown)


tests/system_test.py
```diff
@@ -715,12 +715,40 @@ def wait_connectors(self, **retry_kwargs):
...
+    def wait_startup_message(self, **retry_kwargs):
+        """Wait for router startup message to be printed into logfile
+
+        This ensures that the router installs its signal handlers, avoiding
+        a router failure with return code -15 upon premature SIGTERM (DISPATCH-1689)
+
+        e.g. 2022-03-03 19:08:13.608655 +0100 SERVER (notice) Operational, 4 Threads Running (process ID 2190110)
+        """
+        def _is_startup_line_present(f: TextIO) -> bool:
+            for line in f:
```
ganeshmurthy, line 727: Instead of going line by line, is it possible to do a read() on the file and pass the entire contents of the file to re.search() ?

jiridanek: True, but I don't see any real advantage in the suggested approach.

Reading line by line uses less memory ;P

Either approach would fail if router somehow logged only half of the relevant file; the file was flushed, and the test read the incomplete line in one retry attempt, and the other half in the next retry attempt.

Maybe I should try to add `assert line[-1] == '\n'` to ensure this is not happening... I think it is actually never happening.

kgiusti (approved):

tests/system_test.py
jiridanek: @ganeshmurthy lets discuss this on a new issue or something; i'm going to merge this and hope for GREEN! Thanks.

image

PR #115, opened 2022-03-03
Speed-up python-checker GHA job

jiridanek:
Use Ubuntu directly, and install minimal necessary amount of packages.

PR #141, opened 2022-03-08
Fixed #139: Rename all artifacts to avoid collisions with Qpid Dispatch

### LINKED ISSUE #139: Rename qpid-dispatch artifacts/binaries to skupper
Goal: avoid naming collisions with Apache Qpid Dispatch.  It should be possible to install both skupper-router and qpid-dispatch-router on the same system without overlap.

### Router Name changes

- Binary
  - qdrouterd -> skrouterd
- Tooling
  - qdstat -> skstat; qdmanage->skmanage
- Python
  - python paths: qpid-dispatch -> skupper-router
- Config file path (?)
  - /etc/qpid-dispatch/qdrouterd.conf -> /etc/skupper-router/skrouterd.conf
  - /etc/sasl2/qdrouterd.conf -> /etc/sasl2/skrouterd.conf
- Other pathnames created during install:
  - /share/qpid-dispatch/
  - /include/qpid/dispatch/
  - /share/doc/qpid-dispatch/
  - /lib/python3.10/site-packages/qpid_dispatch[_site.py]/
  - /lib/qpid-dispatch/

kgiusti:
This renames all executables, paths, and other installables to use
skupper oriented names.


.github/workflows/build.yaml
```diff
@@ -496,15 +496,15 @@ jobs:
       - name: Install build dependencies
         run: |
           dnf install -y 'dnf-command(builddep)'
-          dnf builddep -y qpid-dispatch-router --setopt=install_weak_deps=False
+          dnf builddep -y skupper-router --setopt=install_weak_deps=False
```
jiridanek, line 499: This won't work; the Fedora package is still named `qpd-dispatch-router`. Using Fedora package for the purpose seemed expedient at the time. We should IMO merge https://github.com/skupperproject/skupper-router/pull/115 which removes dependence on the Fedora package.

CMakeLists.txt
```diff
@@ -18,7 +18,7 @@
 ##
 
 cmake_minimum_required(VERSION 3.20)
-project(qpid-dispatch C CXX)
+project(skupper-router C CXX)
```
jiridanek, line 21: :+1: thx

jiridanek (changes requested):

jiridanek:
This will break the skupper-router image, esp. the scripts in https://github.com/skupperproject/skupper-router/tree/main/scripts. They will need to be adjusted for renamed binary and config files, in the least. Maybe there is something else I don't see needs to be done in the respect.

.github/workflows/build.yaml
kgiusti: @jiridanek ah - thanks, I was a little lost there.

I've approved #115 and I'll mark this PR as a draft until I can rebase on that change, thanks.

kgiusti:
> This will break the skupper-router image, esp. the scripts in https://github.com/skupperproject/skupper-router/tree/main/scripts. They will need to be adjusted for renamed binary and config files, in the least. Maybe there is something else I don't see needs to be done in the respect.

@jiridanek Yeah unfortunately this change is an API breaking change so we can expect other projects that consumer skupper-router to break when this change lands.  It's also a release blocker since once the release ships the API is pretty much set in stone.  All the more reason to land this sooner rather than later so we can find and fix anything that this change breaks.

Unless there's something in this repo that the patch has missed, this patch is ready to go.

@fgiorgetti @grs @ted-ross See Jiri's comment above and please be aware of this breaking change and let me know if there's anything I can add to this patch to make life a little less miserable for consumers of skupper-router before it lands. thanks.

jiridanek (approved):
Merging this soon and asking Skupper to adapt and overcome is probably the right call.

Btw, I tried to rebase the skupper-integration-tests on top of this PR and it died almost silently on timeouts. There were probably some useful messages in pod logs, but the CI does not allow easy access to that.

```
2022-03-15T12:35:03.6321895Z 2022/03/15 12:35:03 connection not ready yet, current pods state:
2022-03-15T12:35:03.6693782Z KUBECONFIG=/home/runner/.kube/config kubectl -n  public-acceptance-1 get pods -o wide
2022-03-15T12:35:03.9741504Z The connection to the server 10.1.0.187:16443 was refused - did you specify the right host or port?
2022-03-15T12:35:03.9771101Z 
2022-03-15T12:35:03.9798180Z 2022/03/15 12:35:03 timed out waiting for skupper sites to be connected: timed out waiting for sites to be connected
2022-03-15T12:35:04.0194079Z FAIL	github.com/skupperproject/skupper/test/integration/acceptance	926.109s
2022-03-15T12:35:04.0749324Z ?   	github.com/skupperproject/skupper/test/integration/acceptance/annotation	[no test files]
2022-03-15T12:35:04.0852078Z ?   	github.com/skupperproject/skupper/test/integration/acceptance/console	[no test files]
2022-03-15T12:35:18.0387632Z === RUN   TestBasic
2022-03-15T12:35:18.2825753Z === RUN   TestBasic/interiors-ingress-none
2022-03-15T12:35:18.2826147Z     basic.go:244: Testing: Connecting two interiors with ingress=none
```

https://github.com/jiridanek/skupper-router/runs/5553258995?check_suite_focus=true

PR #160, opened 2022-03-10
Fixes #153: Fix leak in message body buffer handling

### LINKED ISSUE #153: qd_message_stream_data_release() sometimes misses out freeing qd_buffers thus leaking them and causing a q2 stall
When I print the sizes of content buffers in _Q2_holdoff_should_unblock_LH buff_ct before and after the buffer frees in qd_message_stream_data_release(), I see a slow leak of buffers

_Q2_holdoff_should_unblock_LH buff_ct=34
_Q2_holdoff_should_unblock_LH buff_ct=2

_Q2_holdoff_should_unblock_LH buff_ct=35
_Q2_holdoff_should_unblock_LH buff_ct=3

_Q2_holdoff_should_unblock_LH buff_ct=68
_Q2_holdoff_should_unblock_LH buff_ct=36

_Q2_holdoff_should_unblock_LH buff_ct=36
_Q2_holdoff_should_unblock_LH buff_ct=4

Notice that the buffer count in the content buffer list is slowly increasing from 2 to 3 to 4 and so on. Once this buffer count reaches q2 lower (128), the router stops receiving and the request stalls.

This slow buffer leak needs to be fixed.

kgiusti:


src/message.c
```diff
@@ -2625,59 +2625,53 @@ void qd_message_stream_data_release(qd_message_stream_data_t *stream_data)
...
-            DEQ_REMOVE(pvt->stream_data_list, stream_data);
-            free_qd_message_stream_data_t(stream_data);
-            return;
-        }
+    qd_buffer_t *stop_buf = stream_data->last_buffer;
+    qd_buffer_t *end_boundary = DEQ_NEXT(stream_data) ? DEQ_NEXT(stream_data)->first_buffer : pvt->body_buffer;
+    if (stop_buf != end_boundary)
+        // safe to free last_buffer since this message no longer references it
+        stop_buf = DEQ_NEXT(stop_buf);
+    assert(stop_buf);
```
ganeshmurthy, line 2639: Should this assert check for stop_buf == end_boundary ?

ganeshmurthy (approved):

src/message.c
kgiusti: No, stop_buf never gets free (it's where we "stop" freeing buffers) so stop_buf == end_boundary is legit and will happen often.

PR #165, opened 2022-03-10
Fixed #163: Added libnghttp2-devel and also added fix to a failing im…

### LINKED ISSUE #163: http2 and grpc tests not running in CI
omission of nghttp2 library is causing the http2 and grpc tests to not run in CI

ganeshmurthy:
…age upload test

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -366,7 +366,7 @@ jobs:
       - name: Install Linux build dependencies
         if: ${{ runner.os == 'Linux' }}
         run: |
-          dnf install -y gcc gcc-c++ cmake libuuid-devel openssl-devel cyrus-sasl-devel cyrus-sasl-plain swig python3-devel python3-pip make libwebsockets-devel ccache libasan libubsan libtsan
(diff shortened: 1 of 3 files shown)


jiridanek:
```
>       self.assertIn('Success', out)
E       AssertionError: 'Success' not found in '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>408 Request Timeout</title>\n<h1>Request Timeout</h1>\n<p>The server closed the network connection because the browser didn&#x27;t finish the request within the specified time.</p>\n'
```

What timeout is that? It looks like something that is either in qdrouterd or the http2 server. It does not look like something imposed by the test itself. Probably it is the http2 server. I suggest increasing that timeout, possibly in another PR.

The failing http2 tests took each approx 115 seconds more than passing run does. That means little less than 40s extra for each of the three large image transfer tests. The image is not that large, actually. 40 s should be more than enough. 

Is it significant this only happened with Proton 0.36 in the CI?

ganeshmurthy:
I believe this failure is caused due to qd_buffer leak which creates a q2 stall. I thought @kgiusti fixed it in https://github.com/skupperproject/skupper-router/commit/604fe4ebb440ae9869627e0cf6a09a9fe7dfc875 but unfortunately not. I will add more logging to confirm that it is indeed the leaking buffer problem.

PR #179, opened 2022-03-13
Use Ninja to build container image, because it is faster than make

jiridanek:
There are three ways in which ninja is faster:

1. ninja runs in parallel by default
2. rerunning ninja on unchanged sources more efficiently detects no work needs to be done (not relevant in Dockerfile)
3. the ninja has been explicitly designed to be fast, overall, which make was not

Image works just fine for me.


jiridanek:
Actually, no. This is not a good idea, because the actual production build is always going to be done with make. That is what rpm builds will do. So there is no point to this

(although it works just fine, see, https://github.com/jiridanek/skupper-router/runs/5574564065?check_suite_focus=true#step:4:758)

Issue #157, opened 2022-03-10
remove dispatch router docs


pwright:
as discussed with group, remove all books and plan to re-introduce them as required

PR #180, opened 2022-03-13
remove docs

pwright:
addresses #157


pwright:
@kgiusti  I wonder whether I should also remove the man page docs, given that we don't intend users to ever start the router from the cli?

kgiusti:
> @kgiusti I wonder whether I should also remove the man page docs, given that we don't intend users to ever start the router from the cli?

I'd keep them, certainly the "qdmanage/qdstat" stuff still applies - but that's just my $0.02 so I'm happy to go along with the general consensus

kgiusti (approved):
LGTM.  The conflict github is complaining about is where I changed the version to 2.0.0-SNAPSHOT in the common/attributes file.  This patch removes that file which is The Right Thing To Do....

pwright:
@kgiusti  resolved conflict, ok to merge?

kgiusti:
> @kgiusti resolved conflict, ok to merge?

@pwright merge man, merge!

PR #181, opened 2022-03-14
Fixed #168: Prevent the policy tests from running in CI

### LINKED ISSUE #168: Prevent policy, policy oversize basic and compound tests from running only on CI
These tests are failing intermittently on CI and cannot be reproduced locally. I want to comment these tests out for now. Skupper has its own policy and the skupper-router's policy might never be used again. If policy is going to be used, we should start running these tests in CI again.

ganeshmurthy:


ganeshmurthy:
@jiridanek it is not totally clear to me why the policy tests are still being run on Ubuntu. It looks like the build is being downloaded from an earlier step - https://github.com/skupperproject/skupper-router/runs/5539179536?check_suite_focus=true#step:3:14  ??

jiridanek (approved):
It's green. Now, don't touch it, or it won't last :P

Issue #48, opened 2022-02-07
The unbearable optionality of dependencies and features


jiridanek:
Qdrouterd build is very adaptable to the circumstances, switching off features and their tests depending on what's available. This is all well for a generic library, but may not be so great for an end-user application where uniformity can simplify things greatly.

For example, should libwebsockets remain to be optional dependency? (I think not.) What about libnghttp2 (There I am not so sure, but I wouldn't leave it optional either.) What about `SKIP_DELETE_HTTP_LISTENER` config option? (I'd decide if that should be ON or OFF and then made it that way; even if it requires messing with libwebsocket version, because the CentOS one does not suit, then so be it.)

What about Python version? RHEL8/CentOS Stream 8 ship with all versions of Python from 3.6 to 3.9. Is there a reason to use the oldest 3.6 version, which means some tests for TLS (that require 3.7) don't run?

kgiusti:
This is probably best converted to a discussion rather than an issue as it raises several points to... discuss.  Issues can then be opened as needed from decisions made.

I agree there's many options that really no longer apply in the skupper-router case, so in general if a feature is required for skupper it should not be optional (HTTP_LISTENER is required, so SKIP_DELETE_HTTP_LISTENER is Not Good, for example).  libnghttp2 is a must have as well and should no longer be optional.  libwebsockets is a requirement as well.

In most (all?) skupper deployments skupper-router will be containerized so we have the liberty to pick whatever version of python is available in the base image that suites our needs.  In the case of Centos8 MHO would be to use 3.7 so we can ensure that the TLS feature is deployed against a version of python that we used to test it in CI.

jiridanek:
@kgiusti I think you've resolved pretty much all the questions. Only thing I am not sure about is this: if we make all system_tests mandatory (not auto-skippable), then the RPM %check will have hard time. So my thinking there is that either some good solution is devised (such as, don't run %check in RPMs, or only run it on Fedora where there are RPM-packaged python dependencies available, or run `pip install` as part of RPM build, and disable network sandbox for the build to permit that, ...) or the tests remain optional, and we somehow better monitor what got skipped (the Jenkins UI helps with that... as much as I am not a fan of Jenkins)

`SKIP_DELETE_HTTP_LISTENER` is not good? But we need it, because some libwebsocket versions don't permit socket deletion

[...]

PR #189, opened 2022-03-15
Fixed #186: Add rpm .spec to skupper-router project

### LINKED ISSUE #186: Add rpm .spec to skupper-router project

jiridanek:
This is a very basic rpm .spec file, which nonetheless can be compiled and installed.

Tested with

```bash
rpkg local --nocheck
sudo dnf install /tmp/rpkg/skupper-router-31-8gtb5wwt/x86_64/skupper-router-0.0.git.3483.7cc6a1d8.dirty.1i20wb-2.fc35.x86_64.rpm
skrouterd
```


jiridanek:
@mcressman @ajssmith Do you think such basic rpm .spec is sufficient for upstream uses, or is there something more that has to be done?

Do we want to also build qpid-proton static libraries as part of this spec file?

skupper-router.spec
```diff
@@ -0,0 +1,99 @@
...
+
+%check
+%ctest
+
+%files router
+/etc/qpid-dispatch/qdrouterd.conf
+/etc/sasl2/qdrouterd.conf
+/usr/bin/qdmanage
+/usr/bin/qdstat
+/usr/sbin/qdrouterd
```
jiridanek, line 83: feedback: the files should be split into multiple output rpm packages, the way qpid-dispatch spec does it in Fedora. That is, separate `-tools` package for skmanage/skstat, and possibly scrapper(?), then another package (not sure how to name it) for skrouterd and the config files. Each package should carry man-pages for their respective content. Not sure where to put LICENSE file, probably with skrouterd.

skupper-router.spec
```diff
@@ -0,0 +1,99 @@
...
+
+%build
+%cmake -DPython_EXECUTABLE=%python3
+%cmake_build
+
+%install
+%cmake_install
+
+%check
+%ctest
```
jiridanek, line 76: this may not be a good idea, I heard multiple people talking against this. Easy to skip with `--nocheck`, though.

skupper-router.spec
jiridanek: I did the split, LICENSE is in -docs package.

.github/workflows/build.yaml
```diff
@@ -584,6 +584,60 @@ jobs:
...
+          mkdir /tmp/skupper-rpms
+          rpkg local --nocheck --outdir /tmp/skupper-rpms
+
+      - name: Install built packages
+        run: |
+          dnf install -y /tmp/skupper-rpms/*/*.rpm
+
+      - name: Check that skrouterd works
+        run: |
+          skrouterd -c /dev/empty |& grep "Configuration file could not be opened"
```
jiridanek, line 631: Ideally, this should test each package independently (Does -server package work without -tools being installed?) and more in depth. Sadly, the -tests package is actually broken and the system tests don't run against installed skrouterd (this is known ENTMQIC issue). For now, this should do. It checks that Python internal module can be loaded, because that happens before router accesses config.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,144 @@
...
+%cmake_build --target all --target man
+
+%install
+%cmake_install
+
+%check
+%ctest
+
+# TODO: name?
+%package server
```
jiridanek, line 70: Previously, the packages were called `qpid-dispatch-router`, `qpid-dispatch-tools`. If the base name is to be `skupper-router`, we can't have `skupper-router-router`. I resolved this by calling it `-server`.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,144 @@
...
+%config /etc/skupper-router/skrouterd.conf
+%config /etc/sasl2/skrouterd.conf
+
+%{python3_sitelib}/skupper_router/
+%{python3_sitelib}/skupper_router_site.py
+%{python3_sitelib}/__pycache__/skupper_router_site.*.pyc
+%{python3_sitelib}/skupper_router-*.egg-info
+
+# TODO: skupper-router-tools depend on tools/command.py, etc. in skupper_router_internal/
+/usr/lib/skupper-router/python/skupper_router_internal/
```
jiridanek, line 92: The qpid-dispatch packages have the same problem. It is permitted to install `qpid-dispatch-tools` without `qpid-dispatch-router`, but then qdmanage/qdstat will fail with

```
$ qdmanage
  File "/usr/bin/qdmanage", line 27, in <module>
    import qpid_dispatch_site  # noqa: F401 # imported for side-effects (inserts to sys.path)
ModuleNotFoundError: No module named 'qpid_dispatch_site'
```

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,144 @@
...
+
+/usr/share/man/man5/skrouterd.conf.5.gz
+/usr/share/man/man8/skrouterd.8.gz
+
+/usr/share/skupper-router/html/index.html
+
+# TODO: package -tools in fact depends on -server, see TODO above
+# TODO: what to do with scrapper tool?
```
jiridanek, line 100: scrapper?

skupper-router.spec.rpkg
jiridanek: @kgiusti @mcressman This is a bug, ^^^, I'd say. I am not sure how to best resolve this. Either make the `-tools` package depend on the server package, or split `qpid-dispatch-site` between tools and the router. The dependency is weird, but the server package is relatively small, and trying to do the python module split is somehow even weirder.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,144 @@
...
+#
+
+# This .spec file uses `rpkg` (https://pagure.io/rpkg-util) to provide
+# pleasant user experience to developers.
+#
+#  Command                  Description
+# `rpkg srpm`              Creates a *.src.rpm file in /tmp/rpkg (exact path is printed)
+# `rpkg local --nocheck`   Builds a *.rpm for your system in /tmp/rpkg (exact path is printed)
+#
+# See `man rpkg` for more commands.
```
jiridanek, line 27: Alternate solution, https://cmake.org/cmake/help/latest/cpack_gen/rpm.html. I only found about it accidentally after I did the rpkg work.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,147 @@
...
+Source: {{{ git_dir_pack }}}
+
+%description
+A lightweight message router, written in C and built on Qpid Proton, that provides flexible and scalable interconnect backend for Skupper.io Level 7 Virtual Application Network.
+
+%prep
+{{{ git_dir_setup_macro }}}
+
+%build
+%cmake -DUSE_LIBWEBSOCKETS=ON -DUSE_LIBNGHTTP2=ON -DPython_EXECUTABLE=%{python3}
```
jiridanek, line 63: On CentOS8, this expands to `-DPython_EXECUTABLE=/usr/libexec/platform-python`. I am not sure this is desired. (Instead of `/usr/bin/python3`). The reason I added this is that on my laptop, CMake would detect Python 3.11 beta and try to build against that.

.github/workflows/build.yaml
```diff
@@ -584,6 +584,60 @@ jobs:
...
+
+      - name: Check that manpages work
+        run: |
+          dnf install -y man
+
+          man skrouterd
+          man skstat
+          man skmanage
```
jiridanek, line 639: The man pages are somewhat outdated regarding skupper-router changes. The synopsis talk about skrouter being good for AMQP, disregarding the differences from old dispatch.

skupper-router.spec.rpkg
jiridanek: The thing looks like the usual sort of dark CMake magic that nobody understands. Given that it internally generates (I think) a spec file, which is another set of magics, it is the kind of tech that is best avoided, me thinks.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,149 @@
...
+%check
+%ctest
+
+# TODO: name?
+%package server
+Summary:  The skrouterd deamon
+Requires: python3
+Requires: qpid-proton-c >= %{proton_minimum_version}
+Requires: libwebsockets >= %{libwebsockets_minimum_version}
+Requires: libnghttp2 >= %{libnghttp2_minimum_version}
```
jiridanek, line 80: Here (and for -tools) there probably should be a Requires: for `cyrus-sasl-plain`. That is needed to pass `%check` (due to `system_tests_sasl_plain`) and actually I think users should have this package ready since they are very likely to need it.

If you disagree this should be runtime dep, it can be made a build dep, maybe condition it on whether %check is required?

skupper-router.spec.rpkg
jiridanek: (Packaging `qpid-dispatch-site` in a separate package that both -server and -tools depend on is certainly also possible, but that would result in a lot of packages)

skupper-router.spec.rpkg
jiridanek, line 80: On a related note, what about depending on openssl?

skupper-router.spec
jiridanek: done

skupper-router.spec.rpkg
jiridanek: Maybe put the dep on -tests package! That should be the smallest possible change. Still, I think that users should not be allowed to install the router without plain sasl mechanism. It is usually needed.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,151 @@
...
+BuildRequires: python3-devel
+BuildRequires: openssl-devel
+BuildRequires: qpid-proton-c-devel >= %{proton_minimum_version}
+BuildRequires: libwebsockets-devel >= %{libwebsockets_minimum_version}
+BuildRequires: libnghttp2-devel >= %{libnghttp2_minimum_version}
+# man pages --help
+BuildRequires: asciidoc
+BuildRequires: python3-qpid-proton >= %{proton_minimum_version}
+# ctest
+BuildRequires: cyrus-sasl-plain
```
jiridanek, line 56: I'm going to add this, so I can run (some) tests in Fedora COPR. The results look ok, with some flakes here and there

https://copr.fedorainfracloud.org/coprs/jdanek/skupper-router/build/3821395/

image

I reported missing Qpid Proton RPM on EPEL 9 at https://issues.apache.org/jira/browse/PROTON-2523

skupper-router.spec.rpkg
kgiusti: Agreed - avoid!

skupper-router.spec.rpkg
kgiusti: I'll go on-record and say we should split out the python bits that are used for the client tools from the router-internal python code.  Ideally the client python stuff could be a stand-alone python package (available via pypi) that the tools/tests/etc would depend upon.

Not advocating that split should be done as part of this patch of course, but a separate future enhancement for some plucky python developer to run with :) ...

skupper-router.spec.rpkg
kgiusti: Darn - is it too late to rename the scraper tool, 'cause "scrapper" is so much better ;)

kgiusti (commented):
I like the idea but to be honest all this RPM-fu is way beyond my expertise.  We should get @fgiorgetti @mcressman involved for their approval.

.github/workflows/build.yaml
kgiusti: Can you open a corresponding GH skupper-router issue for this -tests package breakage?

skupper-router.spec.rpkg
kgiusti: @mcressman @fgiorgetti thoughts?

.github/workflows/build.yaml
jiridanek: https://github.com/skupperproject/skupper-router/issues/267

skupper-router.spec.rpkg
jiridanek: In any case, scraper should not be packaged, because it never before was packaged.

.github/workflows/build.yaml
jiridanek: https://github.com/skupperproject/skupper-router/issues/268

skupper-router.spec.rpkg
jiridanek: Yes, putting it on -tests package. Hopefully this will work the way I expect.

jiridanek: Transitive is fine, IMO.

jiridanek: It does not work as expected, it has to be a `BuildRequires`. There is an enhancement proposal for rpmbuild which got stuck https://bugzilla.redhat.com/show_bug.cgi?id=1134397.

skupper-router.spec.rpkg
```diff
@@ -0,0 +1,156 @@
...
+URL:           https://skupper.io
+VCS:           {{{ git_dir_vcs }}}
+
+BuildRequires: gcc
+BuildRequires: gcc-c++
+BuildRequires: cmake
+# https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_installing-and-using-python_configuring-basic-system-settings#proc_installing-python-3_assembly_installing-and-using-python
+# https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/packaging_and_distributing_software/index#con_common-macros-for-python-3-rpms_assembly_packaging-python-3-rpms
+%{?fedora:BuildRequires: python3.9}
+%{?rhel:BuildRequires: python39-devel}
```
jiridanek, line 53: This is attempt to use Python 3.9 everywhere, as per https://github.com/skupperproject/skupper-router/issues/48.

```
ModuleNotFoundError: No module named 'proton'
```

That happens because `python3-qpid-proton` RPM installs Proton only for the system-default Python. I should've thought of that! Using static embedded build of Proton will not completely help, because the spec that @mcressman has now still needs the RPM package to do, ..., things.

Furthermore, I am not completely positive on the exact procedure to use the 3.9 Python in Fedora and CentOS. I asked on StackOverflow ;P https://stackoverflow.com/questions/71782123/how-do-i-install-python39-rpm-macros-for-fedora-35

skupper-router.spec.rpkg
jiridanek: Fixed now.

skupper-router.spec.rpkg
jiridanek: This can be resolved. We will build Qpid Proton statically as part of this RPM build anyways, which means that the python proton part can be built as well. I am not sure where to put it, though, so that it is out of the way of the user of the system, but still findable for the router. Seems to me that nonstandard location then requires some modifications in the router so that it will look there. (It can be small changes to PYTHONPATH setting, or note the `-I` switch for skrouterd... feels messy

skupper-router.spec.rpkg
jiridanek: do new rpm package, call it `-common`

skupper-router.spec.rpkg
jiridanek: It probably is, since the RPM macro set it this way.

skupper-router.spec.rpkg
jiridanek: Attempt to use Python 3.9 everywhere is considered abandoned, because it was decided so.

skupper-router.spec.rpkg
jiridanek: done

skupper-router.spec.rpkg
jiridanek: (I'll bring it up again when I'll know how to implement it cleanly; It should become easier when Proton build is embedded in building this RPM)

Issue #152, opened 2022-03-10
timeout in system_tests_http1_adaptor


jiridanek:
https://github.com/skupperproject/skupper-router/runs/5488621887?check_suite_focus=true#step:26:1804

```
58: ::Http1AdaptorBadEndpointsTest::test_01_unsolicited_response PASSED
58: ::Http1AdaptorBadEndpointsTest::test_02_bad_request_message FAILED
58: ::Http1AdaptorBadEndpointsTest::test_03_bad_response_message PASSED

[... some time later ...]

31/31 Test #58: system_tests_http1_adaptor ...........***Timeout 1200.13 sec
```

Router log (100MB uncompressed) [TestBadEndpoints.zip](https://github.com/skupperproject/skupper-router/files/8219166/TestBadEndpoints.zip)

jiridanek:
This happened because `TestServer` instance leaked in the failed test. It has to be disposed even if test fails with AssertionError or some other exception.

jiridanek:
https://github.com/skupperproject/skupper-router/runs/5870421205?check_suite_focus=true#step:26:2070

just a timeout (maybe the artifact zips do have something, I did not look)

jiridanek:
http://coprbe.devel.redhat.com/results/jdanek/skupper-router/rhel-8-x86_64/00084043-skupper-router/builder-live.log.gz

```
67/67 Test #61: system_tests_http1_adaptor ..............***Timeout 600.01 sec
test_01_unsolicited_response (system_tests_http1_adaptor.Http1AdaptorBadEndpointsTest) ... ok
test_02_bad_request_message (system_tests_http1_adaptor.Http1AdaptorBadEndpointsTest) ... ok
test_03_bad_response_message (system_tests_http1_adaptor.Http1AdaptorBadEndpointsTest) ... ok
test_04_client_request_close (system_tests_http1_adaptor.Http1AdaptorBadEndpointsTest) ... ok
test_05_client_response_close (system_tests_http1_adaptor.Http1AdaptorBadEndpointsTest) ... ok
test_01_concurrent_requests (system_tests_http1_adaptor.Http1AdaptorEdge2EdgeTest) ... ok
test_02_credit_replenish (system_tests_http1_adaptor.Http1AdaptorEdge2EdgeTest) ... ok
test_03_server_reconnect (system_tests_http1_adaptor.Http1AdaptorEdge2EdgeTest) ... ok
[...]

PR #202, opened 2022-03-17
Fixed #197: Wait for routers in test to exit before checking log for raised exceptions

### LINKED ISSUE #197: `StandaloneEdgeRouterConfigTest.test_48_router_in_error` does not give router enough time
On a busy machine, the test may start reading the router log before the router finished starting up, parsing the config, and getting a chance to log an error.

https://github.com/skupperproject/skupper-router/runs/5575764409?check_suite_focus=true#step:9:4848

```
18:         test_pass = False
18:         with open(self.router_5.outfile + '.out', 'r') as out_file:
18:             for line in out_file:
18:                 if "Exception: Cannot load configuration file test-router-5.conf: role='standalone' not allowed to connect to or accept connections from other routers." in line:
18:                     test_pass = True
18:                     break
18: >       self.assertTrue(test_pass)
18: E       AssertionError: False is not true
18: 
18: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_one_router.py:299: AssertionError
18: - generated xml file: /home/runner/work/skupper-router/skupper-router/skupper-router/build/tests/junitxmls/system_tests_one_router.xml -
18: =========================== short test summary info ============================
18: FAILED ::StandaloneEdgeRouterConfigTest::test_48_router_in_error - AssertionE...
18: =================== 1 failed, 50 passed in 187.94s (0:03:07) ===================
29/31 Test #18: system_tests_one_router ...............***Failed  193.94 sec
```

The following code in the test is insufficient protection against this situation

```python
        # Give some time for the test to write to the .out file. Without
        # this sleep, the tests execute too
        # fast and find that nothing has yet been written to the .out files.
        sleep(3)
```

jiridanek:
This is the smallest-delta fix that I can think of. I don't like the test on the whole, but I overcame my distaste and urge to rewrite. I am proud of myself.

### DIFF
--- a/tests/system_tests_one_router.py
+++ b/tests/system_tests_one_router.py
@@ -21,6 +21,7 @@ import json
 import os
 from subprocess import PIPE, STDOUT
 from time import sleep
+from typing import Sequence
 
 from proton import Condition, Delivery, Message, Timeout, symbol
 from proton.handlers import MessagingHandler
@@ -197,7 +198,7 @@ class StandaloneEdgeRouterConfigTest(TestCase):
             ('router', {'mode': 'standalone', 'id': 'QDR'}),
             ('listener', {'port': cls.tester.get_port(), 'role': 'edge', 'host': '0.0.0.0'})
         ])
-        cls.router = cls.tester.qdrouterd(name, config, wait=False, perform_teardown=False)
+        cls.router = cls.tester.qdrouterd(name, config, wait=False, expect=Process.EXIT_FAIL)
 
         # A standalone router cannot have inter-router connectors.
         name = "test-router-1"
@@ -205,7 +206,7 @@ class StandaloneEdgeRouterConfigTest(TestCase):
             ('router', {'mode': 'standalone', 'id': 'QDR'}),
             ('connector', {'port': cls.tester.get_port(), 'role': 'inter-router', 'host': '0.0.0.0'})
         ])
-        cls.router_1 = cls.tester.qdrouterd(name, config_1, wait=False, perform_teardown=False)
+        cls.router_1 = cls.tester.qdrouterd(name, config_1, wait=False, expect=Process.EXIT_FAIL)
 
         # An edge router cannot have edge listeners.
         # Edge routers can have connectors that connect to interior routers
@@ -215,7 +216,7 @@ class StandaloneEdgeRouterConfigTest(TestCase):
             ('router', {'mode': 'edge', 'id': 'QDR'}),
             ('listener', {'port': cls.tester.get_port(), 'role': 'edge', 'host': '0.0.0.0'})
         ])
-        cls.router_2 = cls.tester.qdrouterd(name, config_2, wait=False, perform_teardown=False)
+        cls.router_2 = cls.tester.qdrouterd(name, config_2, wait=False, expect=Process.EXIT_FAIL)
 
         # Edge routers cannot have inter-router listeners. Only interior
         # routers can have inter-router listeners.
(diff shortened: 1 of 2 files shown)


tests/system_tests_one_router.py
```diff
@@ -242,12 +243,13 @@ def setUpClass(cls):
...
+        cls.router_5 = cls.tester.qdrouterd(name, config_5, wait=False, expect=Process.EXIT_FAIL)
 
-        # Give some time for the test to write to the .out file. Without
-        # this sleep, the tests execute too
+        # Give some time for the test to write to the .out file. Without this, the tests execute too
         # fast and find that nothing has yet been written to the .out files.
-        sleep(3)
+        routers: Sequence[Qdrouterd] = (cls.router, cls.router_1, cls.router_2, cls.router_3, cls.router_4, cls.router_5)
+        for router in routers:
+            router.wait(timeout=TIMEOUT)
```
ganeshmurthy, line 252: This is a good fix. I was wondering if we can catch the [TimeoutExpired](https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired) exception and print the return value of router.wait

jiridanek: I was thinking that anything more is not necessary. If the `wait` timeouts, that means the test fails because of it, and the tearDown will print router log that hopefully contains something explanatory. Also, if `router.wait` timeouts, then there is no return value, only exception, no?

```python
>>> import subprocess
>>> subprocess.Popen(["sleep", "2"]).wait(timeout=1)
  File "/usr/lib64/python3.10/subprocess.py", line 1204, in wait
    return self._wait(timeout=timeout)
  File "/usr/lib64/python3.10/subprocess.py", line 1930, in _wait
    raise TimeoutExpired(self.args, timeout)
subprocess.TimeoutExpired: Command '['sleep', '2']' timed out after 1 seconds
```

ganeshmurthy: Ah yes. I wanted to know why the process would not terminate but alas that we cannot figure out when the exception is thrown. Approved.

ganeshmurthy (approved):

tests/system_tests_one_router.py
jiridanek: thanks, hopefully that will never happen; and if it does, we cross the bridge then!

jiridanek: @ganeshmurthy Actually, if this was `communicate` and not `wait`, in that case it would make sense to print the data carried by the exception. For example, this should be done in https://github.com/skupperproject/skupper-router/issues/226 where it is possible to print the stdout and stderr that the exception carries.

ganeshmurthy: > @ganeshmurthy Actually, if this was `communicate` and not `wait`, in that case it would make sense to print the data carried by the exception. For example, this should be done in #226 where it is possible to print the stdout and stderr that the exception carries.

That would be cool. I just saw the following code in https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

```
proc = subprocess.Popen(...)
try:
    outs, errs = proc.communicate(timeout=15)
except TimeoutExpired:
    proc.kill()
    outs, errs = proc.communicate()
```

We could use the above code in system_tests_http2 which passes every test but occasionally fails to terminate and times out at the end. I am not sure why that test fails to terminate. Some process is not terminating is my guess.

jiridanek: That's even better than what I had in mind. i had in mind that the `TimeoutExpired` gives access to the stdout/stderr that was fetched so far. It makes better sense to SIGKILL the process first, and then get it all.

In system_tests_http1, there are sometimes threads that are not terminating. For that, I've been thinking that the FakeServer's should be registered with Tester#cleanup, so that they are disposed when the test does tearDown.

ganeshmurthy: > 
> In system_tests_http1, there are sometimes threads that are not terminating. For that, I've been thinking that the FakeServer's should be registered with Tester#cleanup, so that they are disposed when the test does tearDown.
Good point. The http2 tests also start a http2 server to test all requests and it should be registered with Tester#cleanup as well. Here is the timeout I am talking about.
https://github.com/skupperproject/skupper-router/runs/5665348873?check_suite_focus=true#step:26:4114

jiridanek: communicate in bad_configuration test issue: https://github.com/skupperproject/skupper-router/issues/226
Http1 issue: https://github.com/skupperproject/skupper-router/issues/152
Http2 issue: https://github.com/skupperproject/skupper-router/issues/233

PR #209, opened 2022-03-19
Fixed #208: Initialize atomic variables without synchronization

### LINKED ISSUE #208: `sys_atomic_init` should use `atomic_init` instead of `atomic_store`
The [`atomic_init`](https://en.cppreference.com/w/c/atomic/atomic_init) function or its deprecated equivalent [`ATOMIC_VAR_INIT`](https://en.cppreference.com/w/c/atomic/ATOMIC_VAR_INIT) macro, provide for initializing an atomic variable when there are no other concurrent accesses possible.

For example,

https://github.com/skupperproject/skupper-router/blob/8147b62df40a0167af489750b7f6d6bd3f135696/src/buffer.c#L40-L48

When new `qd_buffer_t` is being initialized, nobody else has reference to it, so it is pointless to use slow(er) atomic access to set `&buf->bfanout` to zero. Of course, for correctness, this requires that there is some synchronization when the new qd_buffer_t value is made available to the other threads.

This was discussed at a Proton meeting.

jiridanek:


kgiusti (approved):

ganeshmurthy (approved):

jiridanek:
Since everybody seems to think this is a good idea, let's merge!

PR #211, opened 2022-03-19
Remove the GCC `__sync` atomic function implementations

jiridanek:
Given that skupper-router is to only support Linux with -std=gnu11, there is no reason to provide alternate atomic implementations.

Currently, the `__sync` operations were only used by `c_unittests` (surprisingly to me, although I should've realized). Therefore, after removing them here, a provision needs to be made for this C++ code. Which is why the `#ifdef __cplusplus` is used. It is a hack, but if it ever misfires, it can only possibly break c_uniitests, and not production code, so I think this is good enough to both improve router code and also preserve c_uniitests.

Given that there is now only a single atomic implementation, it would be possible to go further and simply use `atomic_uint` and other types directly in the router code and call `atomic_fetch_add` and other such functions on them directly. The `atomic.h` header can then simply hold the C++ compatibility layer, that is, the conditional `using` directives that are needed to make things work in C++.

Alternatively, C11 offers `_Generic` functions, so the `sys_atomic_*` functions could be made to support various atomic variable types... On a second thought, this one is a bad suggestion, seems to me.

### DIFF
--- a/include/qpid/dispatch/atomic.h
+++ b/include/qpid/dispatch/atomic.h
@@ -25,12 +25,17 @@
 
 #include <stdint.h>
 
-/******************************************************************************
- * C11 atomics                                                                *
- ******************************************************************************/
-#if defined(__STDC__) && (__STDC_VERSION__ >= 201112L) && !defined(__STDC_NO_ATOMICS__)
-
+#ifdef __cplusplus
+// Compatibility maneuver required due to c_unittests written in C++.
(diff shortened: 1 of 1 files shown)


kgiusti (approved):
We're requiring C11 now, so I think this cleanup is fine.  @ganeshmurthy @ted-ross agree?

ganeshmurthy (approved):

PR #212, opened 2022-03-22
Add Clang Thread Safety Annotations

jiridanek:
Proposal, aiming to partially provide the features of https://github.com/apache/qpid-dispatch/pull/805, at compile time.

Uses https://clang.llvm.org/docs/ThreadSafetyAnalysis.html (with all the limitations that stem from it, namely, works only in clang or IDE based on clangd, and the GUARDED_BY annotation cannot be used in C structs, https://github.com/llvm/llvm-project/issues/20777)

This is really a draft, but I'd nevertheless appreciate somebody looking at it and giving a preliminary review

## Calling a `_LH` function without holding the lock

warning_should_hold_lock

## Unlocking something that is not locked

releasing mutex not held

## Returning without unlocking

return_without_unlock


decisions/lock checing.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Decision
+
+Annotate all functions that acquire, release, or require locks with the Clang Thread Safety Analysis annotations.
+Run a CI job that will check these annotations.
+
+## Consequences
+
+Header file defining the annotations needs to be included.
+There are multiple possibilities, either the Clang one(mutex.h), or the Zircon one (thread_annotations.h).
```
jiridanek, line 29: This one is licensed under Apache, https://github.com/google/google-input-tools/blob/master/client/base/thread_annotations.h

kgiusti (commented):
I'm in favor of the idea, but would like others to weigh in since it will require all contributors to be aware of these constructs while working on lock-related code.

I'm going to request reviews from some of the proton devs as well since we've involved them in the past regarding how best to interact with the proactor threading model.

jiridanek:
Issues I discovered so far

## suppressing the warning for a single line

Here is one way. The main problem is that GCC will not accept `#pragma clang` annotations (although that warning can be suppressed, this may be actually the cleanest way to deal with this). The clang is happy to accept `#pragma GCC`, so that's what can be used. The suppression for the thread safety warning is clang specific, so there has to be a conditional; thanks to c11, it is possible to put a pragma into a macro

```c
#ifdef __clang__
#define TA_SUPPRESS _Pragma("clang diagnostic ignored \"-Wthread-safety-analysis\"")
#else
#define TA_SUPPRESS
#endif
```

usage is

```c
    // initialize the Q2 flag. Note that we do not need to hold the content
    // lock when we make this call since no other threads are able to access
    // the new message until this function returns.
#pragma GCC diagnostic push
    TA_SUPPRESS;
    if (_Q2_holdoff_should_block_LH(content))
        content->q2_input_holdoff = true;
#pragma GCC diagnostic pop
```

(Suppressing for entire function is easy, just annotate with `TA_NO_THREAD_SAFETY_ANALYSIS`.)

## annotating a function which does not have its lock in the scope is impossible

In the following, there are two functions which must be protected by `LOCK(content->lock);`, but they don't take `content` as an argument, so there is no way to annotate them in the function declaration.

* line 2752: `find_last_buffer_LH(&stream_data->section, &msg->body_cursor, &msg->body_buffer);`
* line 2754: `trim_stream_data_headers_LH(stream_data, !is_footer);`

In case of `trim_stream_data_headers_LH` It is possible to get at the lock through `stream_data`, but that is not good enough for `-Wthread-safety-precise` (see https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#warning-flags)

https://github.com/skupperproject/skupper-router/blob/cfb31ce764e0146693b9f55c32213829634689ba/src/message.c#L2724-L2758

## limitations

there are limitations described here and at the clang doc site; one example is that calls through function pointers are not checked

## alternatives

* dynamic checking, some custom macro, but then the annotation would have to be in function body, not before the `{`, so one annotation cannot serve both for clang and this.
* coverity, see the section "C/C++ Concurrency Primitives" in manual at https://url.corp.redhat.com/cov-customizing; so far I haven't found a way there how to assert lock must be held ;(

jiridanek:
Found the coverity usage, it is in that page, I knew I saw it there before, I just did not realize the second time

```c
#ifdef __COVERITY__
    extern void __coverity_recursive_lock_acquire__(void*);
    extern void __coverity_recursive_lock_release__(void*);
    extern void __coverity_assert_locked__(void*);
#endif

// ...

sal_Bool SAL_CALL osl_releaseMutex(oslMutex pMutex)
{
#ifdef __COVERITY__
    __coverity_assert_locked__(pMutex);
#endif
```

Example from https://dev-builds.libreoffice.org/cppcheck_reports/master/3633.html

jiridanek:
# Calling `_CT` function without being on the core_thread

image

PR #218, opened 2022-03-23
Fixed #217: pass null rather than empty string when testing for ipv6

### LINKED ISSUE #217: http listener with no host emits error in log
The documentation of the `host` field of a `listener` states that 'The empty string listens on all local addresses.'. However for a listener where 'http' is true, not specifying the host results in an error being logged, and ipv6 being disabled, e.g. 'SERVER (error) getaddrinfo(, 9090) failed with Name or service not known'.

This is a result of the check for ipv6 added to fix https://issues.apache.org/jira/browse/DISPATCH-2257

grs:


ganeshmurthy (approved):

jiridanek:
All the test failures seem to be either known, or were just introduced by #2 and are filled in https://github.com/skupperproject/skupper-router/issues. The CI for the PR is therefore a success, of sorts.

grs:
Thanks @jiridanek !

PR #254, opened 2022-03-29
Fixed typo on auto_mesh.py

fgiorgetti:


jiridanek (approved):
I'd also rename the C variable in

https://github.com/skupperproject/skupper-router/blob/cfb31ce764e0146693b9f55c32213829634689ba/src/connection_manager.c#L326

https://github.com/skupperproject/skupper-router/blob/cfb31ce764e0146693b9f55c32213829634689ba/src/connection_manager.c#L445

Just to completely eliminate the wrong capitalization from the codebase.

jiridanek (approved):

PR #255, opened 2022-03-29
Fixed #188: Added a loop to give some time for the data structures in…

### LINKED ISSUE #188: system_tests_qdmanage failure on CI
```
23: =================================== FAILURES ===================================
23: ___________________________ QdmanageTest.test_query ____________________________
23: 
23: self = <system_tests_qdmanage.QdmanageTest testMethod=test_query>
23: 
23:     def test_query(self):
23:     
23:         def long_type(name):
23:             return 'org.apache.qpid.dispatch.' + name
23:     
23:         types = ['listener', 'log', 'router']
23:         long_types = [long_type(name) for name in types]
23:     
23:         qall = json.loads(self.run_qdmanage('query'))
23:         qall_types = {e['type'] for e in qall}
23:         for t in long_types:
23:             self.assertIn(t, qall_types)
23:     
23:         qlistener = json.loads(self.run_qdmanage('query --type=listener'))
23:         self.assertEqual([long_type('listener')] * 4, [e['type'] for e in qlistener])
23:         self.assertEqual(self.router_1.ports[0], int(qlistener[0]['port']))
23:     
23:         qattr = json.loads(self.run_qdmanage('query type name'))
23:         for e in qattr:
23:             self.assertEqual(2, len(e))
23:     
23:         def name_type(entities):
23:             ignore_types = [long_type(t) for t in ['router.link', 'connection', 'router.address']]
23:             return set((e['name'], e['type']) for e in entities
23:                        if e['type'] not in ignore_types)
23: >       self.assertEqual(name_type(qall), name_type(qattr))
23: E       AssertionError: Items in the first set but not the second:
23: E       ('router.node/R2', 'org.apache.qpid.dispatch.router.node')
23: 
23: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_qdmanage.py:190: AssertionError
```
https://github.com/skupperproject/skupper-router/runs/5557333167?check_suite_focus=true#step:9:4219

ganeshmurthy:
… Router 1 to update

### DIFF
--- a/tests/system_tests_skmanage.py
+++ b/tests/system_tests_skmanage.py
@@ -179,15 +179,24 @@ class SkmanageTest(TestCase):
         self.assertEqual([long_type('listener')] * 4, [e['type'] for e in qlistener])
         self.assertEqual(self.router_1.ports[0], int(qlistener[0]['port']))
 
-        qattr = json.loads(self.run_skmanage('query type name'))
-        for e in qattr:
-            self.assertEqual(2, len(e))
-
         def name_type(entities):
             ignore_types = [long_type(t) for t in ['router.link', 'connection', 'router.address']]
             return set((e['name'], e['type']) for e in entities
                        if e['type'] not in ignore_types)
-        self.assertEqual(name_type(qall), name_type(qattr))
+
+        test_passed = False
+        for i in range(10):
+            qattr = json.loads(self.run_skmanage('query type name'))
+            try:
+                for e in qattr:
+                    self.assertEqual(2, len(e))
+                self.assertEqual(name_type(qall), name_type(qattr))
+                test_passed = True
+                break
+            except AssertionError as ae:
+                # Give some time for the Router 1's data structures to update.
+                sleep(0.5)
+        self.assertTrue(test_passed)
 
     def test_get_schema(self):
         schema = dictify(QdSchema().dump())
--- a/tests/system_tests_skmanage.py
+++ b/tests/system_tests_skmanage.py
@@ -28,7 +28,7 @@ from proton.utils import BlockingConnection
 from skupper_router_internal.compat import dictify
 from skupper_router_internal.management.qdrouter import QdSchema
 
-from system_test import unittest
+from system_test import unittest, retry_exception
 from system_test import Logger, TestCase, Process, Qdrouterd, main_module, TIMEOUT, DIR
 from system_test import QdManager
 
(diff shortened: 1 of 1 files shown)


tests/system_tests_skmanage.py
```diff
@@ -179,15 +179,25 @@ def long_type(name):
...
+                self.assertEqual(name_type(qall), name_type(qattr))
+                test_passed = True
+                break
+            except AssertionError as ae:
+                # Give some time for the Router 1's data structures to update.
+                sleep(0.5)
+        self.assertTrue(test_passed)
+
 
     def test_get_schema(self):
```
jiridanek, line 202:
```
62: /home/runner/work/skupper-router/skupper-router/tests/system_tests_skmanage.py:202:5: E303 too many blank lines (2)
62:     def test_get_schema(self):
62:     ^
62: 1
```

tests/system_tests_skmanage.py
```diff
@@ -179,15 +179,24 @@ def long_type(name):
...
+            qattr = json.loads(self.run_skmanage('query type name'))
+            try:
+                for e in qattr:
+                    self.assertEqual(2, len(e))
+                self.assertEqual(name_type(qall), name_type(qattr))
+                test_passed = True
+                break
+            except AssertionError as ae:
+                # Give some time for the Router 1's data structures to update.
+                sleep(0.5)
```
astitcher, line 198: Just passing through! But this sort of code is just asking for a flaky test, In a year or under some heavy load or on a different system 5s (10*0.5s) will not be long enough and the test will fail.
Please, please, pretty please architect systems so that you can test them without arbitrary sleeps!
I expect this is not going to be an issue here, but I'm triggered! /rant

jiridanek: There will always be some arbitrariness in this. The current arbitrary (total) timeout of choice for waiting (the `TIMEOUT` constant) is 60 seconds. That's why I like `system_tests#retry_exception`, because it defaults to using this timeout.

ganeshmurthy: The repeated querying is necessary in this case to find out if the router has the expected state. The setupClass function makes sure that both routers are connected to each other by executing a specific management query on both routers but this not guarantee that everything is up-to-date in each router. We need a good way for the router(s) to tell the world (if asked) that the entire router network is completely ready to go so the clients would not have to retry to find out if the routers have reached steady state.

jiridanek: Nobody's disputing the need for a retry with a timeout, IIUC. The only objection is that 5 second timeout might be too short.

I think you should wait up to `TIMEOUT` seconds in total, as sum of all waits of all iterations.

kgiusti (approved):

PR #259, opened 2022-03-29
Fixed #253: asan leak in qd_policy_c_counts_alloc

### LINKED ISSUE #253: asan leak in qd_policy_c_counts_alloc

mgoulish:
This one is less trivial that the other recent ones I did.

Since the leaked pointers are only being used in Python -- and never freed -- I keep track of them in the dispatch.c file, and free them all just before the router itself is freed.

### DIFF
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -68,6 +68,34 @@ sys_atomic_t global_delivery_id;
 
 qd_dispatch_t *qd = 0;
 
+int    n_policy_longs   = 0;
+int    max_policy_longs = 0;
+long * policy_longs;
+
+void alloc_policy_longs()
+{
+    max_policy_longs = 100;
+    policy_longs = qd_malloc(sizeof(long) * max_policy_longs);
+}
+
+void append_policy_long(long val)
+{
+    if (n_policy_longs >= max_policy_longs) {
+        policy_longs = qd_realloc(policy_longs, 2*max_policy_longs);
+        max_policy_longs *= 2;
+    }
+    policy_longs[n_policy_longs] = val;
+    ++ n_policy_longs;
+}
+
+void free_policy_longs()
+{
+    for (int i = 0; i < n_policy_longs; ++i) {
+        qd_policy_c_counts_free(policy_longs[i]);
+    }
+    free(policy_longs);
+}
+
 qd_dispatch_t *qd_dispatch_get_dispatch()
 {
     return qd;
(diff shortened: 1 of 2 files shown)


jiridanek (commented):
Quick first impression: is this change worth it? I am inclined to say not. The code is not too convoluted, after I got over the initial state of shock and incomprehension, but, it deals with shutdown leak. The OS will free the memory at shutdown anyways. 

Balancing the suppresion effect achieved by two lines in lsan.supp with about 30 new lines of C that achieve more or less the same, I think I prefer lsan.supp.

Regarding why Python is not freeing, I think it could be the cyclical reference in IoAdaptor, https://github.com/apache/qpid-dispatch/pull/1052. But, when Python actually starts freeing things, that results in more problems (shutdown crashes). I spent some time trying to resolve that on the PR, but I never did.

jiridanek:
Uh, sorry, this has nothing to do with IoAdaptor (probably, I still spent much less time investigating than you, so of course I can't know what's really going on).

If this thing causes some memory to be allocated, then I think it should be responsible for deallocating it

https://github.com/skupperproject/skupper-router/blob/662be06f345081a84c2f48d57fcaf6101e3886f5/python/skupper_router_internal/policy/policy_local.py#L532-L550

In Python, classes can have destructor, that would be 

```python
def __del__():
    ctypes.something.free(self.pointer)
```

or there could be some such methods that would be called explicitly. The usual problem with implementing destructor is that you as programmer don't have control when it gets called. Because Python calls it when GC runs. Which in this case is probably fine, because the resource being freed actually _is_ memory. (If the resource was open file, or socket, then linking this to __del__ would not be so great.)

https://docs.python.org/3/reference/datamodel.html?highlight=__del__#object.__del__

jiridanek:
I tried to quickly figure out if there would be some benefit in wrapping the pointer in a Capsule, https://docs.python.org/3/c-api/capsule.html, because that seems to be a way to manage a C pointer with associated destructor. The problem with Capsule is that the Python code is not supposed to use the pointer value, only to pass it to other C functions, and they have to take PyCapsule as parameter, not the raw pointer (the API does not provide the Python side with access to the raw pointer value; though it can be circumvented, of course)... Just a thought.

kgiusti:
IMHO this seems to be trying to compensate for a design that never considered a lifecycle for this counter object.  I think we'd be better off fixing the lifecycle model in the first place - or better yet avoid it entirely:

One question that comes to mind is why are these counters stored in a C allocated memory object in the first place rather than just keeping them in the AppStats class as native python counters?  The c alloc and free functions can then go away entirely, and the c refresh function could take the four long counters as individual parameters  instead of passing a memory pointer.

jiridanek:
This was fixed differently, see the attached issue. If you want to update this and reopen, feel free, but I think that this PR should be closed.

PR #265, opened 2022-03-31
Fixed #263: do not wait for settlement once a non-PN_ACCEPTED termina…

### LINKED ISSUE #263: TCP client connections hang when a non PN_ACCEPTED disposition is received.
Reproducer:  deploy two routers.  On one router provision a tcpListener.  On the other provision a tcpConnector.  Do not provide a service for the connector - we don't want it to connect for this reproducer.

Fire up a curl request on the listener.   Hang.  Instantiate a service for the tcpConnector.  Still hang.  Such sadness.

Reason:  when the curl client connects it sends an HTTP GET request.  This request is put into a delivery and sent to the router with the connector.  The connector activates and attempts to connect to the non-existent service.  This fails, and the delivery is PN_RELEASED and settled.  Back on the listener router the disposition arrives with PN_RELEASED but NOT SETTLED!  

Now feast your eyes on this:
https://github.com/skupperproject/skupper-router/blob/aa3c8358ea1697d592e006064bb0cb0ec6e5f511/src/adaptors/tcp_adaptor.c#L1730

But also this:
https://github.com/skupperproject/skupper-router/blob/aa3c8358ea1697d592e006064bb0cb0ec6e5f511/src/router_node.c#L2022

Now think a bit....

kgiusti:
…l outcome is received

### DIFF
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -1727,21 +1727,25 @@ static void qdr_tcp_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t
                DLV_FMT" qdr_tcp_delivery_update: disp: %"PRIu64", settled: %s",
                DLV_ARGS(dlv), disp, settled ? "true" : "false");
 
-        if (settled && disp == PN_RELEASED) {
-            // When the connector is unable to connect to a tcp endpoint it will
-            // release the message. We handle that here by closing the connection.
-            // Half-closed status is signalled by read_eos_seen and is not
-            // sufficient by itself to force a connection closure.
+        const bool final_outcome = qd_delivery_state_is_terminal(disp);
+        if (final_outcome && disp != PN_ACCEPTED) {
+            // Oopsie! The message failed to be delivered due to error. Since
+            // this is a streaming message do not bother waiting for settlement
+            // because downstream routers will not propagate settlement until
+            // after rx complete is set.  Note this is not half-closed, that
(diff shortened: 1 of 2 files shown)


src/adaptors/tcp_adaptor.c
```diff
@@ -1727,21 +1727,25 @@ static void qdr_tcp_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t
                DLV_FMT" qdr_tcp_delivery_update: disp: %"PRIu64", settled: %s",
                DLV_ARGS(dlv), disp, settled ? "true" : "false");
 
-        if (settled && disp == PN_RELEASED) {
-            // When the connector is unable to connect to a tcp endpoint it will
-            // release the message. We handle that here by closing the connection.
-            // Half-closed status is signalled by read_eos_seen and is not
-            // sufficient by itself to force a connection closure.
+        const bool final_outcome = qd_delivery_state_is_terminal(disp);
```
grs, line 1730: Out of curiousity, is it possible there has been a change in proton behaviour here?

grs (approved):

src/adaptors/tcp_adaptor.c
kgiusti: @grs @ssorj - good question.  I've gone back as far as 0.34.0 proton (min required version for skupper-router) and the issue is present: outcome Released, settled=False... hang.

I've added a simple TCP test that verifies the fix in this PR so going forward we'll be sure this fixed for good.

PR #271, opened 2022-04-01
Fixed #270: Created a manual github action that releases skupper-router

### LINKED ISSUE #270: Create a manual github workflow for releasing skupper-router
Create a new manual workflow where you can select a git tag and release the skupper-router

ganeshmurthy:


ganeshmurthy:
I found this "release" event - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
I tried using it in this yml thinking that when I add the following to the yml - 
```
on:
  release:
    types: [published]
```
The job would trigger when we published the release but that did not happen. Not sure what I was missing

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: Release skupper-router
+#on:
+#  release:
+#    types: [published]
```
jiridanek, line 23: Not sure why this does not work. One alternative would be to do

```
on:
  push:
    tags:
      - 'v**'
```

to trigger this when version tag (starting with `v`) is pushed.

jiridanek: In any case, it might be best to leave this step manual. Just so that weird things don't happen on their own.

fgiorgetti (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build and publish skupper-router image
+    runs-on: ubuntu-latest
+    steps:
```
fgiorgetti, line 31: @ganeshmurthy one thing that might be interesting to evaluate would be the
use of create-release action, see: [](https://github.com/actions/create-release)

This manual task works ok, from the POV of releasing images, but there won't be
any release related information on the repository.

ganeshmurthy: > @ganeshmurthy one thing that might be interesting to evaluate would be the use of create-release action, see: https://github.com/actions/create-release
> 
> This manual task works ok, from the POV of releasing images, but there won't be any release related information on the repository.

Yes, Please see my earlier comment in this PR. I tried using the on: release which did not work. @jiridanek suggested a fix to my approach but then we agreed to keep this process manual.

Issue #148, opened 2022-03-09
system_tests_topology failure


ganeshmurthy:
```
38: self = <system_tests_topology.TopologyTests testMethod=test_01_topology_failover>
38: 
38:     def test_01_topology_failover(self):
38:         name = 'test_01'
38:         if self.skip[name]:
38:             self.skipTest("Test skipped during development.")
38:         test = TopologyFailover(name,
38:                                 self.client_addrs,
38:                                 "closest/01"
38:                                 )
38:         test.run()
38: >       self.assertIsNone(test.error)
38: E       AssertionError: "expected trace ['0/A', '0/D', '0/C', '0/B'] but got ['0/A', '0/D', '0/B']" is not None
```
https://github.com/skupperproject/skupper-router/runs/5481493366?check_suite_focus=true

jiridanek:
Known issue from Dispatch, https://issues.apache.org/jira/browse/DISPATCH-2021

jiridanek:
still present, https://github.com/skupperproject/skupper-router/runs/5692504231?check_suite_focus=true#step:26:114

kgiusti:
This test has a long history of being flakey.  I propose we disable it until we have a chance to fix it.

Issue #143, opened 2022-03-09
Flaky test QdmanageTest::test_query


jiridanek:
https://github.com/skupperproject/skupper-router/runs/5482417212?check_suite_focus=true#step:27:1927

```
>       self.assertEqual(name_type(qall), name_type(qattr))
E       AssertionError: Items in the first set but not the second:
E       ('router.node/R2', 'org.apache.qpid.dispatch.router.node')
```

jiridanek:
Ubuntu GHA fail here, https://github.com/skupperproject/skupper-router/runs/5503069758?check_suite_focus=true#step:9:4179, again with Proton 0.36.0.

kgiusti:
I've seen this as well but have had no luck reproducing it reliably.

PR #299, opened 2022-04-08
Create `flakey` decorator for retrying flaky tests and apply it to #143 and #148

jiridanek:


tests/system_test.py
```diff
@@ -113,6 +114,39 @@ def _check_requirements():
...
+            # Python 3 clears the exception upon leaving the `except` block
+            # https://docs.python.org/3/reference/compound_stmts.html#the-try-statement
+            preserved_exc = None
+            for _ in range(repeats):
+                try:
+                    return f(*args, **kwargs)
+                except AssertionError as exc:
+                    preserved_exc = exc
+                    logging.info("Test marked as flaky has failed: %s", exc)
+            raise AssertionError("Flaky test has failed too many times") from preserved_exc
```
jiridanek, line 143: Ken seems to use `flakey`, I am used to seeing `flaky`. I'm going to use the Ken's terminology thoughout.

ganeshmurthy:
A test that is failing intermittently (a flakey test) either is badly written and needs to be fixed or it is it is exposing a bug in the router. I would rather fix the test bug or the router bug than marking the test as flakey.

There is usually already an issue associated with a flakey test. If a test is marked as flakey and say the test does not fail for a month or so, do we consider that issue fixed  and close it ? Or, are we going to come back to the issue and fix it when we have some time ?

tests/system_test.py
```diff
@@ -113,6 +114,39 @@ def _check_requirements():
...
+        def inner(*args, **kwargs):
+            # Python 3 clears the exception upon leaving the `except` block
+            # https://docs.python.org/3/reference/compound_stmts.html#the-try-statement
+            preserved_exc = None
+            for _ in range(repeats):
+                try:
+                    return f(*args, **kwargs)
+                except AssertionError as exc:
+                    preserved_exc = exc
+                    logging.info("Test marked as flakey has failed: %s", exc)
```
jiridanek, line 142: Maybe increase the log level of that.

jiridanek:
all good points; I knew, going in, that `@flakey` is controversial

> If a test is marked as flakey and say the test does not fail for a month or so, do we consider that issue fixed and close it?

That's how it worked in Dispatch, I'd say; Cannot reproduce -> Closed

jiridanek:
> mark a test as flakey and forget about it

We should not forget. But we should not be nagged by them all the time. I think the idea is to spend less time on CI review revisiting tests we know they are unreliable, again and again. Having the unreliable tests tracked in the Issues should be sufficient. Contributors should not be forced to investigate random flaky failures when they want to propose a PR.

Doing a `grep '@flakey' -A3` will quickly show what tests need fixing.

jiridanek:
> The end result is we are going to ignore fixing the test which for me has the same effect as skipping a test.

Largely yes, but not quite. A `@flakey`-decorated test still has to be capable of passing at least sometimes, because the CI checks for that. So we know it is either flakey, or possibly fixed. It may never become completely broken (=incapable of passing).

Furthermore, the `@flakey` decorator only catches `AssertionError`, so if the test starts failing somehow unexpectedly, with some other exception, that will be also revealed in CI.

The router has been disabling tests in the past. Marking them as flakey would've been better solution, IMO. (Fixing is of course best, but that's matter of prioritization. The `@flakey` decorator gives you that flexibility of postponing the investigation/fix and still having usable CI that reveals problems in all the other reliable tests, and does not waste people's time having to deal with the flaky fails over and over.

(The last time I suggested doing something like this (autoretry) the folks did mostly dislike that; lets see what other people think now, ~3 years later)

tests/system_test.py
```diff
@@ -114,6 +115,39 @@ def _check_requirements():
...
+    If applied, executes the test up to three times, marking it as failed only if it fails each time.
+    Note, that use of this decorator is generally discouraged---tests should pass reliably when their assertions are upheld.
+
+    Example usage
+
+        @flakey(issue='#123')
+        def test_feature_in_fragile_manner():
+            self.assertTrue(...)
+    """
+    del issue  # unused
```
kgiusti, line 130: One thing to consider is to force the use of the issue parameter so we're certain there is a corresponding GH issue for the test.  Keeps us honest. And probably worth including the issue in the logging/assertion error text.

kgiusti (commented):
It's actually "flaky" according to a quick dictionary lookup.  Not saying we should change it (unless you want to) but truth be known I'm a lousy speller...

Ganesh has a good point that this decorator can easily be abused by over-use.

I'm okay with this as long as we have solid evidence that the test is truly flaky and if so we should probably make fixing/removing them release blockers.

But I'll delegate approval to Mr. Murthy since he's the CI Czar...

jiridanek:
Closed due to the lack of enthusiasm.

PR #307, opened 2022-04-09
Fix `QDROUTERD_RUNNER` CMake option parsing to be shell-like

jiridanek:
Previously, the option value was parsed by splitting the string at every space character, even within quoted substrings


kgiusti (approved):
Nice.

PR #308, opened 2022-04-09
Fixed #306: Build and use static Proton library in RPM build

### LINKED ISSUE #306: Make rpm .spec in skupper-router project resemble the Containerfile build more
The Containerfile builds static Qpid Proton libraries and links them into Skupper Router. (Not right now, yet, but after when https://github.com/skupperproject/skupper-router/pull/96 is merged.) This procedure should be replicated in the .spec file.

jiridanek:


ganeshmurthy:
Would be good if Mike Cressman had a look at this once.

jiridanek:
@ganeshmurthy You'd need this to get green build for your TLS PR. The Proton TLS library is not available in RPM-packaged Proton, it is necessary to build our own.

jiridanek:
Got some help from @clime with the COPR builds from Proton tar.gz. COPR is running again https://copr.fedorainfracloud.org/coprs/jdanek/skupper-router/build/4220877/ With that, I think this is reasonable .spec, ready for review.

jiridanek:
@ganeshmurthy Thanks, I think I will merge this. @mcressman pretty much saw this; it is not dissimilar from his .spec for the TP3. I think I did the Proton compilation slightly better than what his PR does, without requiring any patches, but that's probably just because I have newer Proton to work with. And I have the `-common` package that Justin suggested.

PR #309, opened 2022-04-09
Add linebreaks to Containerfile RUN commands

jiridanek:
I'm hoping this will help make the Containerfile more approachable in the future. For example, now it is much easier to see that libuuid is installed in the image, and I can now start asking relevant people whether Proton still needs it. (AFAIK it does not.)

Btw, maybe I figured out why `cyrus-sasl-plain` was originally present in the build image: because it is required to run `ctest`!

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -26,7 +26,13 @@ RUN dnf -y --setopt=tsflags=nodocs install \
  && dnf config-manager --set-enabled powertools \
  && dnf clean all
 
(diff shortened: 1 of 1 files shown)


ganeshmurthy:
If I understand correctly, this is just to improve human readability ?

jiridanek:
Yes, only that. Shorter lines so it is easier to work with.

ganeshmurthy (approved):

PR #314, opened 2022-04-10
Fixed #142: Do not use mutable default values in function parameter lists in Python code

### LINKED ISSUE #142: Do not use mutable default values in function parameter lists in Python code
See https://github.com/skupperproject/skupper-router/pull/106 where @kgiusti pointed out the issue.

jiridanek:
I haven't found any other instances of this, but it is quite hard to grep for this because the function signature does not have to be on a single line. This needs a better (syntax-aware) tool than grep.

### DIFF
--- a/python/skupper_router_internal/management/config.py
+++ b/python/skupper_router_internal/management/config.py
@@ -48,12 +48,11 @@ class Config:
     def __init__(
             self,
             filename: Optional[str] = None,
-            schema: QdSchema = QdSchema(),
             raw_json: bool = False
     ) -> None:
-        self.schema = schema
-        self.config_types = [et for et in schema.entity_types.values()
-                             if schema.is_configuration(et)]
+        self.schema = QdSchema()
+        self.config_types = [et for et in self.schema.entity_types.values()
+                             if self.schema.is_configuration(et)]
         self._log_adapter = LogAdapter("AGENT") if _log_imported else None
 
         if filename:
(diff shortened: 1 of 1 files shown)


python/skupper_router_internal/management/config.py
```diff
@@ -48,9 +48,11 @@ class Config:
     def __init__(
             self,
             filename: Optional[str] = None,
-            schema: QdSchema = QdSchema(),
+            schema: QdSchema = None,
             raw_json: bool = False
     ) -> None:
+        schema = schema or QdSchema()
```
ganeshmurthy, line 54: I want to understand the problem here. This class is instantiated in this file and the init is called once.  Is there a fear that QdSchema would be somehow manipulated and passed back in again. Could we just remove the schema parameter  and instantiate QdSchema inside the function.

jiridanek: The fear is the usual "Mutable default arguments" gotcha, https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments, yes.

> Could we just remove the schema parameter and instantiate QdSchema inside the function.

That would work too. The constructor is never called with the `schema` parameter, the value is always defaulted.

ganeshmurthy: > The fear is the usual "Mutable default arguments" gotcha, https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments, yes.
> 
> > Could we just remove the schema parameter and instantiate QdSchema inside the function.
> 
> That would work too. The constructor is never called with the `schema` parameter, the value is always defaulted.

I prefer to just remove the schema parameter altogether and instantiate the schema inside the function. It is clean and easy to understand.

ganeshmurthy (approved):

PR #318, opened 2022-04-11
Workaround Proton build issues by building with `make` and `--parallel 1`

jiridanek:
Let's see how this goes. I have high hopes.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -122,7 +122,6 @@ jobs:
           cmake "${{github.workspace}}/qpid-proton" \
             "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" \
             "-DCMAKE_BUILD_TYPE=${BuildType}" \
-            "-GNinja" \
             ${ProtonCMakeExtraArgs}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -122,11 +122,10 @@ jobs:
           cmake "${{github.workspace}}/qpid-proton" \
             "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" \
             "-DCMAKE_BUILD_TYPE=${BuildType}" \
-            "-GNinja" \
             ${ProtonCMakeExtraArgs}
 
       - name: qpid-proton cmake build/install
-        run: cmake --build "${ProtonBuildDir}" --config ${BuildType} -t install --parallel 6
```
ganeshmurthy, line 129: if we leave out the --parallel command line option, is it the same as --parallel 1 ?

jiridanek: Depends on the generator. For `make`, yes, for `ninja`, no. I think it is better to be explicit.

ganeshmurthy (approved):

Issue #335, opened 2022-04-14
Leaked termini and qdr_link_t at router shutdown


ganeshmurthy:
When a attach comes in just before router shutdown, the qdr_link_first_attach() is called which adds an action to the core thread with the qdr_link_inbound_first_attach_CT callback. When discard is true, qdr_link_inbound_first_attach_CT does not free the termini. It cannot free the qdr_link_t object since someone is holding its reference and that person might try to use it. 
This leak can be reproduced intermittently by running the system_tests_http2_tls.py system test.
```
56: ERROR: Aborted due to unexpected alloc pool leak of type 'qdr_link_t'
56: 
56: <<<<
56: 
56: Router QDR debug dump file:
56: >>>>
56: alloc.c: Items of type 'qd_iterator_t' remain allocated at shutdown: 1 (SUPPRESSED)
56: alloc.c: Items of type 'qdr_field_t' remain allocated at shutdown: 1 (SUPPRESSED)
56: alloc.c: Items of type 'qd_buffer_t' remain allocated at shutdown: 1 (SUPPRESSED)
56: alloc.c: Items of type 'qdr_terminus_t' remain allocated at shutdown: 2
56: Leak: 2022-04-06 13:48:04.996099 +0000 type: qdr_terminus_t address: 0x612000035f50
56: /lib64/libasan.so.6(+0x44701) [0x7fba53ea6701]
56: skrouterd(qd_alloc+0x1fe) [0x4c064e]
56: skrouterd(qdr_terminus+0x2e) [0x6226ae]
56: skrouterd() [0x6c6f67]
56: skrouterd() [0x6d27f2]
56: skrouterd() [0x67aa85]
56: skrouterd() [0x682156]
56: /lib64/libc.so.6(+0x8db1a) [0x7fba528b4b1a]
56: /lib64/libc.so.6(+0x112660) [0x7fba5[2939](https://github.com/skupperproject/skupper-router/runs/5851875396?check_suite_focus=true#step:26:2939)660]
56: 
56: Leak: 2022-04-06 13:48:04.996123 +0000 type: qdr_terminus_t address: 0x612000035dd0
56: /lib64/libasan.so.6(+0x44701) [0x7fba53ea6701]
56: skrouterd(qd_alloc+0x1fe) [0x4c064e]
56: skrouterd(qdr_terminus+0x2e) [0x6226ae]
56: skrouterd() [0x6c7009]
56: skrouterd() [0x6d27f2]
[...]

PR #336, opened 2022-04-14
#335: Partially address the leak of qdr_link_t by adding it to the su…

ganeshmurthy:
…ppression list. The termini objects can be freed since we have ownership of those objects

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -1711,26 +1711,24 @@ static void qdr_link_inbound_first_attach_CT(qdr_core_t *core, qdr_action_t *act
             qdr_delivery_decref(core, initial_dlv,
                                 "qdr_link_inbound_first_attach_CT - discarding action");
 
-        if (discard) {
-            //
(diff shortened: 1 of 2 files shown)


src/router_core/connections.c
```diff
@@ -1702,16 +1702,40 @@ static void qdr_link_inbound_first_attach_CT(qdr_core_t *core, qdr_action_t *act
...
+
+    qdr_terminus_t *source = action->args.connection.source;
+    qdr_terminus_t *target = action->args.connection.target;
+
     if (discard || !conn || !link) {
         if (initial_dlv)
             qdr_delivery_decref(core, initial_dlv,
                                 "qdr_link_inbound_first_attach_CT - discarding action");
+
+        if (discard) {
```
kgiusti, line 1714: Remove this conditional - you want to free these terminus if !conn or !link as well since in all those cases the source/target are leaked.

kgiusti (approved):

Issue #344, opened 2022-04-17
[gcc-12] src/iterator.c:596:20: error: 'octet' may be used uninitialized [-Werror=maybe-uninitialized]


jiridanek:
https://github.com/jiridanek/skupper-router/runs/6053922543?check_suite_focus=true#step:19:170

```
/home/runner/work/skupper-router/skupper-router/skupper-router/src/iterator.c:596:20: error: 'octet' may be used uninitialized [-Werror=maybe-uninitialized]
  596 |             result = (unsigned char) octet;
      |             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/skupper-router/skupper-router/skupper-router/src/iterator.c:594:21: note: 'octet' declared here
  594 |             uint8_t octet;
      |                     ^~~~~
cc1: all warnings being treated as errors
```

jiridanek:
This issue can be workarounded if `-flto` compiler switch is provided. Either with LTO the compiler has more information to determine value is not actually uninitialized, or there is a compiler bug that hides the warning.

PR #345, opened 2022-04-17
Use Fedora 36 in GitHub Actions CI builds

jiridanek:
* [x] Workaround [PROTON-2473: Deprecation warnings when compiling Proton with OpenSSL 3](https://issues.apache.org/jira/browse/PROTON-2473) 
* [x] currently fails due to https://github.com/skupperproject/skupper-router/issues/344
* [x] and https://github.com/skupperproject/skupper-router/issues/346
* [x] https://github.com/skupperproject/skupper-router/issues/636
* [x] https://github.com/skupperproject/skupper-router/issues/543

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -307,7 +307,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         container: ['fedora']
-        containerTag: ['35']
+        containerTag: ['36']
         buildType: [RelWithDebInfo]
         runtimeCheck: [asan, tsan]
         protonGitRef:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -464,10 +464,12 @@ jobs:
       - name: Zero ccache stats
         run: ccache -z
 
+      # PROTON-2473: either install `openssl1.1-devel` compat version, or build with `-Wno-error=deprecated-declarations`
       - name: qpid-proton cmake configure
         working-directory: ${{env.ProtonBuildDir}}
         run: >
           cmake "${{github.workspace}}/qpid-proton" \
+            '-DCMAKE_C_FLAGS="-Wno-error=deprecated-declarations"' \
             "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" \
             "-DCMAKE_BUILD_TYPE=${BuildType}" \
             ${ProtonCMakeExtraArgs}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -348,7 +348,7 @@ jobs:
           # coverage
           - os: ubuntu-20.04
             container: 'fedora'
-            containerTag: 35
+            containerTag: 36
             buildType: Coverage
             runtimeCheck: OFF
             protonGitRef: 0.37.0
@@ -429,10 +429,6 @@ jobs:
         run: |
           dnf -y install epel-release
           dnf config-manager --set-enabled powertools
-
-      - name: Install dmesg on Fedora 35
-        if: ${{ matrix.container == 'fedora' }}
-        run: dnf install -y util-linux-core
 
       - name: Install Linux build dependencies
         run: |


kgiusti (approved):
Path looks good to me but should probably land after the new failures have been fixed....

kgiusti:
@jiridanek Thinking out loud... I wonder if these failures are due to the deprecated interface no longer working on F36?  In other words maybe ignoring the warning isn't going to work?

jiridanek:
> @jiridanek Thinking out loud... I wonder if these failures are due to the deprecated interface no longer working on F36? In other words maybe ignoring the warning isn't going to work?

You mean the OpenSSL 3.0? I don't think it stopped working. There would be a good way to try: install the openssl compat package instead and see what happens there! Let me try.

jiridanek:
Did not work ;(

```
Error: 
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
 Problem: conflicting requests
  - package libwebsockets-devel-4.3.1-1.fc[36](https://github.com/skupperproject/skupper-router/runs/8096011353?check_suite_focus=true#step:8:37).i686 requires openssl-devel, but none of the providers can be installed
  - package libwebsockets-devel-4.3.1-1.fc36.x86_64 requires openssl-devel, but none of the providers can be installed
  - package libwebsockets-devel-4.3.2-1.fc36.i686 requires openssl-devel, but none of the providers can be installed
  - package libwebsockets-devel-4.3.2-1.fc36.x86_64 requires openssl-devel, but none of the providers can be installed
  - package openssl1.1-devel-1:1.1.1n-1.fc36.i686 conflicts with openssl-devel provided by openssl-devel-1:3.0.2-4.fc36.i686
```

jiridanek:
@kgiusti It seems that with openssl 1.1 there is one additional failure on Fedora 36

https://github.com/skupperproject/skupper-router/runs/8096468704?check_suite_focus=true

```
The following tests FAILED:
	 40 - system_tests_ssl (Failed)
	 54 - system_tests_http2 (Failed)
	 56 - system_tests_http2_tls (Failed)
```

kgiusti:
@jiridanek yes - @ganeshmurthy is looking into the http2 stuff but I tried hacking a bit in a f36 container with proton main.  From what I can tell, that system_tests_ssl test only fails if we build proton against openssl1.1-devel, then remove openssl1.1-devel and install openssl-devel (due to exactly those lws/nghttp2 deps).

The system_tests_ssl test only passes if both main proton and skupper-dispatch are build with the default openssl-devel.

Is the use of openssl1.1-devel required for proton 0.37 support only?  If so, maybe it make sense to just test main on f36 and leave 0.37 to centos/ubuntu?  What do you think?

jiridanek:
> From what I can tell, that system_tests_ssl test only fails if we build proton against openssl1.1-devel, then remove openssl1.1-devel and install openssl-devel (due to exactly those lws/nghttp2 deps)

Thanks for investigating this more! I was thinking that maybe the mismatch in openssl version used by Python, Proton-Python, and Proton C in the router could be responsible... The default config could've changed, causing that particular failure...

> Is the use of openssl1.1-devel required for proton 0.37 support only? If so, maybe it make sense to just test main on f36 and leave 0.37 to centos/ubuntu? What do you think?

I would not use openssl1.1. It is a crutch to make old sw work that does not support current openssl version. Proton should not need it. It would be best to figure out the cause of the system_tests_http2 / system_tests_http2_tls failures on fedora 36, which seems to be failing regardless of openssl version used, and use latest openssl.

The only difference between proton 0.37 and proton main is afaik that main has the warning suppression pragma for the openssl3.0 deprecations; https://github.com/apache/qpid-proton/commit/5aaa655cb

kgiusti:
@jiridanek @ganeshmurthy The http2 failures may not be related to ssl - Ganesh is trying to root cause those now.

From local testing it appears that the system_tests_ssl failure is due to building proton with openssl1.1-devel, then upgrading to the default openssl-devel to build skupper router.  Building proton main without openssl1.1 causes system_tests_ssl to pass for me.

Is openssl1.1-devel required to support 0.37.0?

jiridanek:
> @jiridanek @ganeshmurthy The http2 failures may not be related to ssl - Ganesh is trying to root cause those now.
> 
> From local testing it appears that the system_tests_ssl failure is due to building proton with openssl1.1-devel, then upgrading to the default openssl-devel to build skupper router. Building proton main without openssl1.1 causes system_tests_ssl to pass for me.
> 
> Is openssl1.1-devel required to support 0.37.0?

@kgiusti I was probably extremely unclear in my comment https://github.com/skupperproject/skupper-router/pull/345#issuecomment-1232150615. I thought I addressed all of ^^^ in it. ;(

kgiusti:
> 
> @kgiusti I was probably extremely unclear in my comment [#345 (comment)](https://github.com/skupperproject/skupper-router/pull/345#issuecomment-1232150615). I thought I addressed all of ^^^ in it. ;(

@jiridanek Sorry - you were completely clear but for some reason my brain wasn't fully engaged!

@ganeshmurthy has found a problem with our use of the hypercorn python package v0.14.1 that appears to be responsible for those http2 failures.  He's trying a temporary fix to pin that package to a working version. If that works it would make sense to rebase this patch to include his changes and see if the http2 tests pass.

kgiusti:
FYI: I'm working on a fix to #636 
system_tests_ssl needs to be refactored a bit...

jiridanek:
In the last rebase, I saw only

* https://github.com/skupperproject/skupper-router/issues/807

to happen, which is a known issue on main already. So, it is time to merge this, I think. And, we can think about Fedora 37 upgrade already ;P

jiridanek:
> Path looks good to me but should probably land after the new failures have been fixed....

@kgiusti @ganeshmurthy heh, the user_id test failure (https://github.com/skupperproject/skupper-router/issues/636) looks to be still present. I guess I should revert this, then... or feel free to push the revert button on the PR yourself

PR #348, opened 2022-04-18
Fixes #327: Removed VERSION.txt

### LINKED ISSUE #327: Replace `VERSION.txt` with a solution not requiring manual updates
## Current status

1. When running a build from the main branch `skrouterd --version` with respond with `2.0.0-SNAPSHOT`.
2. Released version is without the `-SNAPSHOT` suffix.
3. Packagers (Fedora rpm, and others) can modify `VERSION.txt` and add information about their specific build.

This works with either git checkout or with exported sources .tar.gz archive.

## Desired status

1. We would like `skrouterd --version` to respond with the commit sha with say the branch name like so `master-9875634fdd` or something like that. We want the user to know that the software is not released yet.
2. Released version should print the release version, which is given in the git tag
3. Packagers should still be able to add a suffix to the upstream version to modify version information of skrouterd in their package

All this should work with both sources checkout and with exported sources .tar.gz archive.

ganeshmurthy:
Added a create release workflow on tag push
Modified CMakeLists.txt to use the supplied VERSION build paramter which defaults to UNKNOWN


.github/workflows/release.yml
```diff
@@ -18,26 +18,42 @@
...
-#  release:
-#    types: [published]
 on:
-  workflow_dispatch:
+  push:
+    # Sequence of patterns matched against tags
+    # Right now, we run this job automatically when a semantically versioned
+    # tag is pushed.
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
```
jiridanek, line 27: Maybe escape the period in the regular expression? Afaik what's there currently would match non-semantic version strings such as `99999` or `0x0y0`.

RELEASE.adoc
```diff
@@ -0,0 +1,43 @@
...
+  specified in the https://github.com/skupperproject/skupper-router/blob/main/Containerfile[Containerfile,window=_blank]
+  and push the image to quay.io
+  *** The skupper-router image tag name will be obtained from the name of the tag that was used to initiate the release workflow.
+  ** Creates a *draft release* which can be seen here - https://github.com/skupperproject/skupper-router/releases/
+* Edit the draft release by clicking the pencil link and verify if the details in the draft release are accurate.
+  ** Add more details to the text box if necessary. The text box should already contain a link to the issues that
+     were fixed in this milestone.
+* When everything has been verified click the *Publish release* button to publish the release.
+* Create the next milestone if necessary.
+* There is no need to advance any versions on the master branch.
```
jiridanek, line 43:
```suggestion
* There is no need to advance any versions on the main branch.
```

RELEASE.adoc
```diff
@@ -0,0 +1,43 @@
...
+specific language governing permissions and limitations
+under the License
+////
+
+# Use the following steps to generate a skupper-router release
+
+* Create a branch (off of the main branch commit) you want to release
+** Name the branch with the patch part set to 'x'. For example 2.0.x
+* Test the branch to make sure there are no obvious bugs. If bugs are found, create relevant github issues, fix the bugs,
+  push the fixes to the master branch and cherry-pick those same fixes to the newly created branch.
```
jiridanek, line 25:
```suggestion
  push the fixes to the main branch and cherry-pick those same fixes to the newly created branch.
```

jiridanek (commented):
I'd put brief documentation for the VERSION CMake parameter to the table at the end of `README.adoc`

As a future improvement, I think it would be worth it to have CMake default to running `git describe` to obtain a version (assuming git is present and build is happening from a git-controlled directory.

I am wondering if it is correct to exclude `.git` directory in .dockerignore file.

fgiorgetti:
> I am wondering if it is correct to exclude `.git` directory in .dockerignore file

It helps with: `COPY . .` which is done in the builder container

.github/workflows/release.yml
fgiorgetti: I believe it is not a real regular expression. There are a few characters that must be escaped, but `.` in this case 
is not one of them, see:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

fgiorgetti (commented):
Only issue I have is when I run the image.sh script locally. The version is set as an empty string.

jiridanek (approved):
The `git describe` feature works for me; now I have nice version for local builds as well :+1:

jiridanek:
> > I am wondering if it is correct to exclude `.git` directory in .dockerignore file
> 
> It helps with: `COPY . .` which is done in the builder container

Yeah, but it means that CMake inside container build cannot learn version from git, it has to have the version given from outside in container build arguments. No matter, it is what it is; we can provide version as necessary.

CMakeLists.txt
```diff
@@ -58,7 +59,28 @@ else()
     option(QD_ENABLE_ASSERTIONS "Enable assertions" OFF)
 endif()
 
-file(STRINGS "${CMAKE_SOURCE_DIR}/VERSION.txt" QPID_DISPATCH_VERSION)
+if(NOT DEFINED VERSION)
+  # Version was not provided. First check to see if git is avaialble
```
jiridanek, line 63:
```suggestion
  # Version was not provided. First check to see if git is available
```

CMakeLists.txt
```diff
@@ -58,7 +59,35 @@ else()
...
+      if ('${POS}' STREQUAL '-1')
+         # 'fatal' was not found in the output, we will use the DEFAULT_VERSION as the QPID_DISPATCH_VERSION
+         set(QPID_DISPATCH_VERSION ${DEFAULT_VERSION})
+       else()
+         set(QPID_DISPATCH_VERSION "UNKNOWN")
+       endif('${POS}' STREQUAL '-1')
+   else(Git_FOUND)
+       # Git executable was not available, we will not be able to determine the version, just set it to "UNKNOWN"
+       set(QPID_DISPATCH_VERSION "UNKNOWN")
+   endif(Git_FOUND)
```
jiridanek, line 78: I looked through CI and I saw, in the RPM job, this

```
fatal: not a git repository (or any of the parent directories): .git
-- Found Git: /usr/bin/git (found version "2.31.1") 
CMake Error at CMakeLists.txt:68 (string):
  string sub-command FIND requires 3 or 4 parameters.
```

At the same time, i was considering maybe it would be better to look at the exit code of git, instead of looking for `fatal` in stderr

```suggestion
      execute_process(
              COMMAND ${GIT_EXECUTABLE} describe --tags --dirty=-modified --always
              OUTPUT_VARIABLE DEFAULT_VERSION
              RESULT_VARIABLE GIT_RESULT
              OUTPUT_STRIP_TRAILING_WHITESPACE)
      # You might sometimes get a fatal error when running the above command (when this is not a git repo).
      if (GIT_RESULT EQUAL 0)
         # Git succeeded, we will use the DEFAULT_VERSION as the QPID_DISPATCH_VERSION
         set(QPID_DISPATCH_VERSION ${DEFAULT_VERSION})
       else()
         set(QPID_DISPATCH_VERSION "UNKNOWN")
       endif(GIT_RESULT EQUAL 0)
   else(Git_FOUND)
       # Git executable was not available, we will not be able to determine the version, just set it to "UNKNOWN"
       set(QPID_DISPATCH_VERSION "UNKNOWN")
   endif(Git_FOUND)
```

jiridanek: RPM job seems to be doing the right thing now,

```
-- Found Git: /usr/bin/git (found version "2.31.1") 
fatal: not a git repository (or any of the parent directories): .git
-- Setting skupper-router version to UNKNOWN
```

ganeshmurthy: Sounds good. Thanks for testing.

fgiorgetti (approved):
LGTM

Issue #321, opened 2022-04-11
system_tests_two_routers KeyError: 'adminStatus'


jiridanek:
https://github.com/jiridanek/skupper-router/runs/5978454315?check_suite_focus=true#step:26:851

looks like random flaky fail

```
28:                 if event.message.properties['statusDescription'] != 'OK':
28:                     error = "Expected statusDescription to be OK but instead got %s" % \
28:                             event.message.properties['statusDescription']
28: >               if event.message.body['adminStatus'] != self.deleted_admin_status:
28: E               KeyError: 'adminStatus'
28: 
28: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_two_routers.py:484: KeyError
```

jiridanek:
https://github.com/jiridanek/skupper-router/runs/5976545992?check_suite_focus=true#step:26:849

ganeshmurthy:
It looks like the `event.message.properties['statusDescription']` was not OK. There was a problem with the request but I am not sure what the problem is. That is why `event.message` did not have a body. The following if statement should be an elif statement. That way we can see the actual error

ganeshmurthy:
I am going to close this issue. When the real error shows up, we will open a new issue for it.

PR #358, opened 2022-04-20
#321: Fixed test code so we can see the statusDescription error more …

ganeshmurthy:
…clearly


ganeshmurthy:
This is not a fix for the problem but rather a fix to the code so we can see the `statusDescription` error

kgiusti (approved):

PR #366, opened 2022-04-21
Fixes #365 - Replace valgrind with sanitizers in Containerfile

### LINKED ISSUE #365: Remove Valgrind dependency from Containerfile
The proposed solution is to build everything two (or three) times: normal build, `-fsanitize=address,leak,undefined` build and maybe also `-fsanitize=thread`. Each build will produce `skrouterd` binary. The first build will be normally installed, and the other two binaries will be renamed and copied to some convenient location.

The container startup script will be updated to run the debug binary when DEBUG is requested. This way, valgrind is not needed. GDB is present in UBI, so it can be permitted to stay. The `-fsanitize=address,leak,undefined` binary cannot be run under GDB because leak sanitizer in a ptraced program it will crash on exit (when it tries to find the leaks).

This change is not approved yet, strictly speaking, but I think there are no valgrind lovers who would oppose. It means build will run longer, and you don't have the convenience of single binary. On the other hand, sanitizers are significantly faster than valgrind at runtime.

jiridanek:


Containerfile
```diff
@@ -74,5 +74,10 @@ ARG version=latest
 ENV VERSION=${version}
 ENV QDROUTERD_HOME=/home/skrouterd
 
+ENV ASAN_OPTIONS="disable_coredump=0 detect_odr_violation=0 strict_string_checks=1 detect_stack_use_after_return=1 check_initialization_order=1 strict_init_order=1 detect_invalid_pointer_pairs=2"
+ENV LSAN_OPTIONS="disable_coredump=0 suppressions=/lsan.supp"
```
jiridanek, line 78: lsan needs to use `ptrace`, which requires capabilities, for which I tend to run container with `--privileged`. Afaik this is no different from gdb, and therefore is not a problem.

one solution (to have asan without lsan) would be to define here `detect_leaks=0`.

.github/scripts/compile.sh
```diff
@@ -33,30 +38,45 @@ do_patch () {
...
+    -DProton_DIR="$WORKING/proton_install${suffix}/usr/lib64/cmake/Proton" \
+    -DBUILD_TESTING=OFF \
+    -DVERSION="${VERSION}" \
+    -DCMAKE_INSTALL_PREFIX=/usr
+  cmake --build "$WORKING/build${suffix}" --parallel "$(nproc)" --verbose
+}
+
+do_build "" OFF
+do_build "_asan" asan
+do_build "_tsan" tsan
```
jiridanek, line 72: The image build used to take 6 mins on GHA, this PR ran 14 minutes. Too bad. Each of these builds is different (different gcc options) so there is no space for any caching or such ways.

scripts/launch.sh
```diff
@@ -33,7 +33,11 @@ fi
 if [[ $QDROUTERD_DEBUG = "gdb" ]]; then
     exec gdb -batch -ex "run" -ex "bt" --args skrouterd $ARGS
 elif [[ $QDROUTERD_DEBUG = "valgrind" ]]; then
-    exec valgrind skrouterd $ARGS
+    exec skrouterd_asan $ARGS
```
jiridanek, line 36: backward compatibility ftw!

jiridanek:
Contained in https://github.com/skupperproject/skupper-router/pull/372

PR #369, opened 2022-04-22
add lws to image build

ajssmith:
Built with out libuv on and verified that router starts correctly (e.g. health check is succesful)


Containerfile
```diff
@@ -22,21 +22,23 @@ FROM registry.access.redhat.com/ubi8/ubi:latest as builder
...
-    libwebsockets-devel libnghttp2-devel \
+    libnghttp2-devel \
     wget patch findutils git valgrind \
  && dnf clean all -y
+
 WORKDIR /build
 COPY . .
 ENV PROTON_VERSION=0.37.0
 ENV PROTON_SOURCE_URL=${PROTON_SOURCE_URL:-http://archive.apache.org/dist/qpid/proton/${PROTON_VERSION}/qpid-proton-${PROTON_VERSION}.tar.gz}
+ENV LWS_VERSION=v3.2.3
```
jiridanek, line 40: somewhat OFFTOPIC: I wonder about the version choice. Epel had `4.0.1-2.el8` Why not use latest here? That would be `v4.3.1`, https://github.com/warmcat/libwebsockets/releases/tag/v4.3.1

.github/scripts/compile.sh
```diff
@@ -35,13 +35,37 @@ do_patch () {
...
+    -D LWS_WITH_STATIC=OFF \
+    -D LWS_IPV6=ON \
+    -D LWS_WITH_HTTP2=OFF \
+    -D LWS_WITHOUT_CLIENT=OFF \
+    -D LWS_WITHOUT_SERVER=OFF \
+    -D LWS_WITHOUT_TESTAPPS=ON \
+    -D LWS_WITHOUT_TEST_SERVER=ON \
+    -D LWS_WITHOUT_TEST_SERVER_EXTPOLL=ON \
+    -D LWS_WITHOUT_TEST_PING=ON \
+    -D LWS_WITHOUT_TEST_CLIENT=ON \
```
jiridanek, line 62: What is the logic for choosing what options to enable? Is it a list of all switches currently available, or is it carefully chosen?

Containerfile
jiridanek: Table of all versions in the opensource RH world, https://src.fedoraproject.org/rpms/libwebsockets

.github/scripts/compile.sh
ganeshmurthy: I am guessing these options were chosen based off off the build log of LWS in brew.

ganeshmurthy:
I am guessing that the health check is still working after LWS_WITH_LIBUV=OFF ?

.github/scripts/compile.sh
```diff
@@ -35,13 +35,37 @@ do_patch () {
...
+mkdir -p qpid-proton-src build staging proton_build proton_install lws-src lws_build lws_install
 tar -zxf qpid-proton.tar.gz -C qpid-proton-src --strip-components 1
+tar -zxf libwebsockets.tar.gz -C lws-src --strip-components 1
 
 do_patch "patches/proton" qpid-proton-src
 
-cd proton_build
+cd $WORKING/lws_build
+cmake \
+    -D LWS_LINK_TESTAPPS_DYNAMIC=ON \
```
ganeshmurthy, line 48: This is minor but just for consistency sake let's remove the space between the -D and the options.  
E.g. instead of `-D LWS_WITH_LIBUV=OFF`, let's do `-DLWS_WITH_LIBUV=OFF`

.github/scripts/compile.sh
jiridanek: Could be, the Rawhide version is https://src.fedoraproject.org/rpms/libwebsockets/blob/rawhide/f/libwebsockets.spec#_58

.github/scripts/compile.sh
jiridanek: Is it more valuable to be consistent with libwebsockets.spec file or with other cmake builds in this dockerfile? :P

.github/scripts/compile.sh
ganeshmurthy: Looking at that rawhide link, it seems to me that the `LWS_WITH_LIBUV` option by default is `OFF`
You can see that in the LWS CMakeLists.txt
https://github.com/warmcat/libwebsockets/blob/main/CMakeLists.txt#L215

But in brew, this option is set to ON

.github/scripts/compile.sh
ganeshmurthy: I would like it if it were consistent with other builds in this file.

jiridanek: /me quietly withdraws from the thread

.github/scripts/compile.sh
ajssmith: I used what was here https://dist-git.host.prod.eng.bos.redhat.com/cgit/rpms/libwebsockets/ as a starting point and then turned LIBUV off

Containerfile
jiridanek: @ganeshmurthy wdyt, also taking https://github.com/skupperproject/skupper-router/issues/48 into consideration. We don't have to resolve this on the PR right away, of course. Just something to keep in mind and eventually take care of. 

Depending also on what @mcressman thinks, of course.

.github/scripts/compile.sh
ajssmith: agree that we should make it consistent with other formats in the project

Containerfile
ganeshmurthy: I am all for using the latest version of LWS

.github/scripts/compile.sh
jiridanek: makes sense, for lws 3.x, I think; for latest lws the reasonable defaults would be the fedora rawhide

ganeshmurthy (approved):

PR #372, opened 2022-04-22
Remove usage of CentOS 8 repos from Containerfile

jiridanek:
This combines the LWS PR from @ajssmith (https://github.com/skupperproject/skupper-router/pull/369) with my https://github.com/skupperproject/skupper-router/pull/366 and achieves a build that uses only ubi8 repos, without requiring any other repositories.

One concession had to be made, despite original plan, in not building tsan variant of skrouterd, as libtsan is not in ubi8. Bug requesting tsan is filled at https://bugzilla.redhat.com/show_bug.cgi?id=2078034.


ganeshmurthy:
Should this PR land first - https://github.com/skupperproject/skupper-router/pull/369/files - and then this PR should be rebased ? Both are modifying the same content in compile.sh

jiridanek:
@ganeshmurthy This PR contains the commit from #396 (with exception of raising lws version, and it has lws compile options one per line, which is IMO better than what ansmith has there now, ...) I wanted to show that it can be all made to work together.

The order does not really matter. I agree #369 should be merged first, because 1) it is already approved 2) lets not waste manager's time on doing rebases :P

jiridanek:
> Should this PR land first - https://github.com/skupperproject/skupper-router/pull/369/files - and then this PR should be rebased ? Both are modifying the same content in compile.sh

@ganeshmurthy Rebased.

.github/scripts/compile.sh
```diff
@@ -21,6 +21,32 @@
 
 set -euxo pipefail
 
+WORKING=$(pwd)
+
+#region libwebsockets
+wget ${LWS_SOURCE_URL} -O libwebsockets.tar.gz
+tar -zxf libwebsockets.tar.gz --one-top-level=lws-src --strip-components 1
+
+cmake -S "$WORKING/lws-src" -B "$WORKING/lws_build" \
```
ganeshmurthy, line 30: You wanted to put each -D build option in its own line for clarity ?

jiridanek: If I was adding the code for the first time, then I'd do it this way. Now I am trying to minimize the number of changes in the PR. I am pretty much rewriting the whole script, but still, what can be preserved should be preserved.

I don't expect we'll be touching the lws options much.

ganeshmurthy: I do think that putting each -D option in its own line increases clarity and also makes it look uniform within the file

jiridanek: done

jiridanek: @ganeshmurthy done

Containerfile
```diff
@@ -46,20 +40,13 @@ RUN .github/scripts/compile.sh
...
-
-# gdb and valgrind are part of final image as they can be used as debug options for Skupper
-RUN dnf -y --setopt=tsflags=nodocs install \
+# gdb and sanitizers are part of final image as they can be used as debug options for Skupper
+RUN dnf -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs install \
     glibc \
     cyrus-sasl-lib cyrus-sasl-plain cyrus-sasl-gssapi libuuid openssl \
     python3 \
     libnghttp2 \
-    gdb valgrind \
```
fgiorgetti, line 62: @jiridanek is proposing the removal of valgrind, so that we can get
rid of the CentOS 8 repos, we must also remove it as a valid debug option.
@grs Are you ok with that?

jiridanek: Actually, replaced valgrind; instead of `valgrind skrouterd` I propose you do `skrouterd_asan`. That is instrumented binary running with the address sanitizer. The Containerfile builds the router twice, first to get skrouterd, then to get skrouterd_asan binary.

grs: Yes, fine with me, but we should update the skupper command as well. At present it enforces only valgrind or gdb as valid options. We can change that to whatever is recommended, just make it a boolean or even remove the option entirely and provide some other way to enable it.

jiridanek: was the valgrind debug option part of Red Hat's TP3? I don't recall seeing valgrind in the image there, but I might remember wrong. Was valgrind and gdb expected to be available in the Red Hat's GA?

scripts/launch.sh
```diff
@@ -33,7 +33,9 @@ fi
 if [[ $QDROUTERD_DEBUG = "gdb" ]]; then
     exec gdb -batch -ex "run" -ex "bt" --args skrouterd $ARGS
 elif [[ $QDROUTERD_DEBUG = "valgrind" ]]; then
```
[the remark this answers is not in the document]
jiridanek: done, valgrind elif removed

ganeshmurthy:
This looks good to me. It would be good to get either @fgiorgetti or @grs to approve this. Thanks.

jiridanek:
> This looks good to me. It would be good to get either @fgiorgetti or @grs to approve this. Thanks.

I'll consider it sufficiently approved on the skupper side of the change, https://github.com/skupperproject/skupper/pull/742

PR #377, opened 2022-04-25
Fixes #233: Make sure the process is killed using the kill() call if …

### LINKED ISSUE #233: timeout in system_tests_http2
All tests in system_tests_http2 pass but there seems to be a problem with the test teardown.

```
97% tests passed, 1 tests failed out of 32

Total Test time (real) = 1229.13 sec

The following tests FAILED:
	 55 - system_tests_http2 (Timeout)
Errors while running CTest
```

https://github.com/skupperproject/skupper-router/runs/5665348873?check_suite_focus=true#step:26:4114

ganeshmurthy:
…the terminate() call did not do it

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -47,7 +47,7 @@ import unittest
 import uuid
 from copy import copy
 from datetime import datetime
-from subprocess import PIPE, STDOUT
+from subprocess import PIPE, STDOUT, TimeoutExpired
 from threading import Event
 from threading import Thread
 from typing import Callable, TextIO, List, Optional, Tuple
@@ -282,17 +282,22 @@ class Process(subprocess.Popen):
         @param expect: Raise error if process status not as expected at end of test:
             L{RUNNING} - expect still running.
             L{EXIT_OK} - expect process to have terminated with 0 exit status.
-            L{EXIT_FAIL} - expect process to have terminated with exit status 1.
+            L{EXIT_FAIL} - expect process to have terminated with exit status >= 1.
             integer    - expected return code
         @keyword stdout: Defaults to the file name+".out"
         @keyword stderr: Defaults to be the same as stdout
         """
         self.name = name or os.path.basename(args[0])
         self.args = args
+        if expect is None:
+            raise RuntimeError("Process (name=%s) specified a None for except" % self.name)
         self.expect = expect
+        self.actual = None
         self.outdir = os.getcwd()
         self.outfile = os.path.abspath(self.unique(self.name))
         self.torndown = False
+        self._logger = Logger(title="Process logger for name=%s" % self.name,
+                              print_to_console=True)
         kwargs.setdefault('stdin', subprocess.PIPE)
         with open(self.outfile + '.out', 'w') as out:
             kwargs.setdefault('stdout', out)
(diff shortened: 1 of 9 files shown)


tests/system_test.py
```diff
@@ -282,17 +282,22 @@ def __init__(self, args, name=None, expect=EXIT_OK, **kwargs):
         @param expect: Raise error if process status not as expected at end of test:
             L{RUNNING} - expect still running.
             L{EXIT_OK} - expect process to have terminated with 0 exit status.
-            L{EXIT_FAIL} - expect process to have terminated with exit status 1.
+            L{EXIT_FAIL} - expect process to have terminated with exit status >= 1.
```
jiridanek, line 285: and also `<0`, to be extra sure? Remember the exit codes of `-11` and so on in the report... Sometimes the exit code is `uint8` and sometimes it is `int8`. I never found what's the logic behind it.

tests/system_test.py
```diff
@@ -317,19 +322,55 @@ def teardown(self):
...
+                                 "process %s (name=%s). Trying to kill process now" % (self.pid, self.name))
+                self.kill()
+                try:
+                    self.wait(timeout=TIMEOUT)
+                    self._logger.log("Process %s killed successfully" % self.pid)
+                except TimeoutExpired:
+                    # There is something seriously wrong if kill() did not kill the process.
+                    # But should the test suite fail because of this ? Right now, it will
+                    # but we need to revisit this.
+                    error_msg = "Error: Could not kill process "
```
jiridanek, line 366: I can't imagine what could go wrong. Kill could fail due to permission errors (user cannot kill other user's process), but I don't think there is anything that can mess up the permissions while the test is running. SIGKILL cannot be blocked by the program. Maybe the process might die slowly on a slow machine?

There are zombie processes, but that means the parent is slow to collect; yet the parent is the test here. Apparently, processes can enter "uninterruptible sleep" when waiting for IO on NFS...

I'd leave the handling here with the presumption that this will never happen.

tests/system_test.py
```diff
@@ -282,17 +282,22 @@ def __init__(self, args, name=None, expect=EXIT_OK, **kwargs):
         @param expect: Raise error if process status not as expected at end of test:
             L{RUNNING} - expect still running.
             L{EXIT_OK} - expect process to have terminated with 0 exit status.
-            L{EXIT_FAIL} - expect process to have terminated with exit status 1.
+            L{EXIT_FAIL} - expect process to have terminated with exit status >= 1.
             integer    - expected return code
```
jiridanek, line 286: What does `None` mean?

tests/system_test.py
```diff
@@ -317,19 +322,55 @@ def teardown(self):
...
+        if status is None:
+            # The process is still running.
+            # Call process.terminate() which will try to stop the child by sending it a SIGTERM
+            # but is not guaranteed to stop it
             self.terminate()
-            if self.expect is not None and self.expect != Process.RUNNING:
-                error("still running")
-            self.expect = 0     # Expect clean exit after terminate
-            status = self.wait()
-        if self.expect is not None and self.expect != status:
```
jiridanek, line 331: This is sufficiently complicated logic that I am tempted to suggest there should be a test for it :P

(I'm actually almost serious; writing tests for tests is nonsense, but writing tests for test helpers like this is IMO not otter nonsense.)

tests/system_test.py
jiridanek: According to https://tldp.org/LDP/abs/html/exitcodes.html, codes for death after signal are `128 + signumber`, which, when represented as signed integer, produces those negative numbers.

jiridanek: I'd leave this the way it was, that is, EXIT_FAIL means exit code == 1. If people want to check different codes, let them be explicit.

tests/system_test.py
```diff
@@ -317,19 +322,55 @@ def teardown(self):
...
+                    # There is something seriously wrong if kill() did not kill the process.
+                    # But should the test suite fail because of this ? Right now, it will
+                    # but we need to revisit this.
+                    error_msg = "Error: Could not kill process "
+                    self._logger.log(error_msg + self.pid)
+                    # This will raise a RuntimeError and fail the test
+                    error(error_msg)
+
+        # At this time, we want to check if self.expect and self.actual are the same
+        if self.expect != self.actual:
```
jiridanek, line 372: What if `self.expect` was None? Then we don't care what self. actual is, IMO.

tests/system_test.py
ganeshmurthy: This teardown() function handles expect codes like `SIGTERM (<0)` as seen here - https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_edge_router.py#L2700

tests/system_test.py
ganeshmurthy: None is not allowed anymore,

tests/system_test.py
ganeshmurthy: The hope is that this piece of code will never be executed. But it is there just in case.

tests/system_test.py
ganeshmurthy: `None` is not allowed for `self.expect`

tests/system_test.py
```diff
@@ -282,17 +282,22 @@ def __init__(self, args, name=None, expect=EXIT_OK, **kwargs):
...
             L{EXIT_OK} - expect process to have terminated with 0 exit status.
-            L{EXIT_FAIL} - expect process to have terminated with exit status 1.
+            L{EXIT_FAIL} - expect process to have terminated with exit status >= 1.
             integer    - expected return code
         @keyword stdout: Defaults to the file name+".out"
         @keyword stderr: Defaults to be the same as stdout
         """
         self.name = name or os.path.basename(args[0])
         self.args = args
+        if expect is None:
```
ganeshmurthy, line 292: A `RuntimeError` is raised if except is `None`

tests/system_test.py
ganeshmurthy: The existing tests are testing the expect and actual (<0, 0, >0) cases. As far as the process termination part is concerned, every teardown() will go thru it, so bugs in this function will bubble up immediately.

jiridanek:
Thanks for explaining the intention with `None`.

tests/system_test.py
kgiusti: Getting rid of None is a good idea - the expect parameter has defined an explicit set of valid values that does not include None.

Personally I would consider any test passing None for expect a coding error, and rather than raising an exception I'd just:

assert expect is not None, "Process argument expect cannot be None"

kgiusti:
Stepping back a bit - and not to criticize this patch at all, it's doing the best job for a complicated subject - but maybe we can simplify this.

I think the __init__() call for the Process() class may be too complicated with respect to how it deals with expected process state.

I think the root of the complication is that the expect argument is overloaded.  It trying to reflect both the "expected state" of the process (running vs. not running), and if the expected_state is "not running", the expected exit status.

What if we separated those two concerns into two separate arguments, for example:

expected_state = [RUNNING or EXITED]
expected_exit_code = <integer> (default=0 for success)

Further enforce the following: if expected_state==RUNNING then expected_exit_code MUST BE zero.  Which makes total sense, because if it's running at this point it should exit cleanly when terminate() is called.  

Also passing in an expected_exit_code of None is a coding error, since a valid exit code will never be None.

Splitting these two attributes can make the code a bit easier to understand (IMHO).  Something like this (totally untested):


    if (expected_state == RUNNING)
        assert expected_exit_code == 0, "Bad coder! No commit for you!"
        terminate()
        try:
           actual_exit_code = wait(TIMEOUT)
        except timeout:
           actual_exit_code = None # it is still running, this is bad

    else (expected_state == EXITED)
        assert expected_exit_code is not None, "must specify exit code!"
        actual_exit_code = poll()

    # If still running at this point that is an error: kill and fail test.
    # No need to check exit code at this point because it is pointless.

    if actual_exit_code == None:
       # still running
       kill()
       try:
           wait(TIMEOUT)
       finally:
           raise Runtime Error, fail test

    # if we get here the process is done and we have an exit code
    # validate it:
    if actual_exit_code != expected_exit_code:
       raise RuntimeError blah blah wrong exit code blah blah blah

Not a small amount of code, but I think it more accurately reflects the goal of testing the exit state/code. IMHO.

ganeshmurthy:
It is true that the code we currently have is combining the expected exit code (0, <0, >0) and expect (RUNNING, EXIT_OK, EXIT_FAIL). This PR is attempting to *translate* the actual exit code (the output of process.poll()) into one of RUNNING, EXIT_OK, EXIT_FAIL.

A real simplification would be to remove the expect parameter altogether and replace it with an expected_exit_code parameter. The new  expected_exit_code parameter will have to match the output of the process,poll() function (None means process still running and will be the default) 

That being said, I think what we have in this PR keeps it simple stupid and will work for all cases and is an improvement over what we currently have. If in the future we need more flexibility with checking the exact exit codes, we could move to a newer model.

jiridanek (approved):

jiridanek:
It is an improvement. lgtm

PR #378, opened 2022-04-26
Fixes #376 - Address watch doesn't work on edge routers

### LINKED ISSUE #376: Address watch doesn't work on Edge routers
Address watches established on edge routers do not detect reachability in other parts of the network.  The tests don't address these cases.

ted-ross:


src/router_core/address_watch.c
```diff
@@ -127,15 +129,43 @@ static void qdr_core_watch_address_CT(qdr_core_t *core, qdr_action_t *action, bo
...
+                watch->watch_handle = action->args.io.value32_1;
+                watch->addr         = addr;
+                watch->handler      = action->args.io.watch_handler;
+                watch->context      = action->args.io.context;
+                DEQ_INSERT_TAIL(core->addr_watches, watch);
+
+                addr->watch = watch;
+                addr->ref_count++;
+
+                qdrc_event_addr_raise(core, QDRC_EVENT_ADDR_WATCH_ON, addr);
```
kgiusti, line 164: Would it be possible to trigger the newly watched address now so the watcher gets the current counters?

ted-ross: This sounds like a good idea.

ted-ross: I added the requested feature.

kgiusti (approved):

Issue #383, opened 2022-04-27
Optimizations for Flow-Log Events


ted-ross:
This issue includes two necessary optimizations for Flow Logging:

1. Allow record events to be batched such that one message delivery may hold multiple (up to 50) record updates.  This will significantly reduce the processing and network overhead involved in event transmission.
2. Rate-limit the emission of events to one update per second per record.  This will greatly reduce the number of events transmitted in the case of high-throughput and long-lived protocol streams.

PR #384, opened 2022-04-27
Flow logging optimizations outlined in issue #383

ted-ross:


src/protocol_log.c
```diff
@@ -1038,10 +1091,12 @@ static void *_plog_thread(void *context)
 
     _plog_flush_TH(core);
```
ted-ross, line 1092: This flush will only push out one slot.  We need to flush all slots here.

ganeshmurthy (approved):

PR #397, opened 2022-04-29
Issue #48 - Make libwebsockets and libnghttp2 mandatory

jiridanek:
Router does not make sense without websockets or http2 support, so let's not permit such build.

This does not (yet) help with one remaining special libwebsockets case, but I believe this is just a matter of time.

https://github.com/skupperproject/skupper-router/blob/96730e81fdefb6002e224cb9e0363cdcd4674694/python/skupper_router_site.py.in#L50-L51

Thanks to embedding libwebsockets in the container build, the only thing holding it back are distribution packages, and Ubuntu 22.04 is on the way to CIs near you!

* See the https://github.com/skupperproject/skupper-router/issues/48 issue for detailed discussion of the proposal.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -46,6 +46,7 @@ set(qpid_dispatch_SOURCES
   adaptors/http1/http1_client.c
   adaptors/http1/http1_server.c
   adaptors/http1/http1_request_info.c
+  adaptors/http2/http2_adaptor.c
   adaptors/tcp_adaptor.c
   adaptors/test_adaptor.c
   alloc.c
@@ -68,6 +69,7 @@ set(qpid_dispatch_SOURCES
   entity_cache.c
   failoverlist.c
   hash.c
+  http-libwebsockets.c
   iterator.c
   log.c
   message.c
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -532,7 +532,7 @@ jobs:
       - name: Install Linux build dependencies
         if: ${{ runner.os == 'Linux' }}
         run: |
-          sudo apt update; sudo apt install -y python3-qpid-proton libpython3-dev ninja-build
+          sudo apt update; sudo apt install -y python3-qpid-proton libpython3-dev ninja-build libwebsockets-dev libnghttp2-dev
```
jiridanek, line 535: this is a disadvantage of the change; cmake creates some `py` files (from `py.in`) files and there is now no way to have that part of the build run without providing all dependencies for the rest of the build

tests/system_tests_http.py
```diff
@@ -315,13 +315,10 @@ def address():
...
 
-        # DISPATCH-1513: libwebsockets versions 3.2.0 introduces a new flag called
+        # DISPATCH-1513: libwebsockets version 3.2.0 introduced a new flag called
         # LWS_SERVER_OPTION_ALLOW_HTTP_ON_HTTPS_LISTENER
         # The new flag allows (as the flag says) HTTP over HTTPS listeners.
-        # Since this flag is not available before lws 3.2.0 we need
-        # to selectively disable this check
-        if skupper_router_site.LIBWEBSOCKETS_VERSION >= (3, 2, 0):
-            self.assert_get("http://localhost:%s" % r.ports[0])
+        self.assert_get("http://localhost:%s" % r.ports[0])
```
jiridanek, line 321: the feature seems a bit unsafe, thinking about it; like, https downgrade attack?

jiridanek: This should be mentioned to @grs for definitive judgement.

ganeshmurthy: There is a requireSsl flag on the listener entity as seen here - https://github.com/skupperproject/skupper-router/blob/main/python/skupper_router/management/skrouter.json#L731

I think this code is trying to honor the value of that flag. If requireSsl is set to false, then a we are allowed to make a http request over https.

This begs the question, why would someone associate an sslProfile with a listener and then go out of the way to set requireSsl to false ? I don't know the exact answer but see here on the amqp side - https://github.com/skupperproject/skupper-router/blob/main/src/server.c#L416

On the bright side, the adaptors don't have such a flag. In the case of adaptors, if an sslProfile is associated with listener, then ssl is required by default.

jiridanek:
> Please take a look at my other comments in this PR. Once those are addressed, this is ready to merge

I don't see any other comments from you besides this one, on https://github.com/skupperproject/skupper-router/pull/397

ganeshmurthy (changes requested):

ganeshmurthy (approved):

tests/system_tests_http.py
grs: I think we could get rid of the option, I agree it seems likely to result in unsafe configuration if used.

jiridanek: https://github.com/skupperproject/skupper-router/issues/572

PR #398, opened 2022-04-29
Issue #910, #48, #102 - Require Python 3.9 in RPM package

### LINKED ISSUE #102: Remove downgrade of the python version in GHA configuration
Use the default system version of python instead.

Need to remove [this step](https://github.com/skupperproject/skupper-router/blob/b96b1e792e28fee9347cc51cd3b7bb57aaa986f2/.github/workflows/build.yaml#L99) from the workflow configuration.

jiridanek:
As the linked issue explains, Python 3.6 is out-of-upstream-support. Is it possible to use a newer Python on ubi8? Yes! Is it advisable? Not sure. Can it be done for RPMs too? Yes! It would be easy if there was python-proton packaged for python39. But there is not one.

The solution needed for the RPM is a real mess (and it is only a good fortune that it plugs into an already existing mess with `skupper-router-internal` and `sys.path` modifications that the router already implements).

I think that if Python >3.6 was deemed worth the effort, it would be necessary to either have a python39-qpid-proton RPM package (which does not have file conflicts with the python3-qpid-proton package) or decide only package the skrouterd in a RPM and distribute the sktools using pip, or something along these lines.

I am not myself convinced this Python 3.9 is a good idea, overall. It is interesting, though, that CentOS/ubi has support for using non-system version of Python for RPMs. On CentOS, there is even automatic rewrite of the `#!` in scripts to use Python 3.9 (or whichever version is chosen).

* See the https://github.com/skupperproject/skupper-router/issues/48 issue for detailed discussion.

### DIFF
--- a/packaging/skupper-router.spec
+++ b/packaging/skupper-router.spec
@@ -119,9 +119,13 @@ cd %{_builddir}/qpid-proton-%{proton_vendored_version}
 %__cmake --build "%{__cmake_builddir}" %{?_smp_mflags} --verbose
 %__cmake --install "%{__cmake_builddir}"
 
+# for `import proton` when rendering `sktools --help` to manpages, and for running tests later
+# this will install all-in-one cpython .so module with proton inside
+%python3 -m pip install --target "%{buildroot}/usr/lib/skupper-router/python/" %{_builddir}/qpid-proton-%{proton_vendored_version}/%{__cmake_builddir}/python/dist/python-qpid-proton-*.tar.gz
+#%python3 -m pip install --target "%{buildroot}/usr/lib/skupper-router/python/" %{_builddir}/qpid-proton-%{proton_vendored_version}/%{__cmake_builddir}/python/dist/python_qpid_proton-*.whl
+export PYTHONPATH="%{buildroot}/usr/lib/skupper-router/python/"
+
 cd %{_builddir}/skupper-router-%{version}
-# for `import proton` when rendering sktools --help to manpages
-source %{_builddir}/qpid-proton-%{proton_vendored_version}/%{__cmake_builddir}/config.sh
 %cmake \
     -DVERSION="%{version}" \
     -DPython_EXECUTABLE=%{python3} \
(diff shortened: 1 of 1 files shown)


jiridanek:
The idea is that as we need to build skupper RPMs, we cannot depend on python39-qpid-proton (because, I am assuming, it is not yet available). Therefore, it is necessary to build the python binding for proton here, and ship it (in rpm skupper-router-common). That is the cause for most of the changes in the PR diff

Next, replacing python3 with python39 constitutes the rest of the diff.

jiridanek:
I tried building this on COPR, which showed the following repeated test fail

```
======================================================================
FAIL: test_80_stats (system_tests_tcp_adaptor.TcpAdaptor)
----------------------------------------------------------------------
  File "/builddir/build/BUILD/skupper-router/tests/system_tests_tcp_adaptor.py", line 1337, in test_80_stats
    self.assertEqual(es_inta_connections_opened, 7)
AssertionError: 6 != 7
```

e.g. https://download.copr.fedorainfracloud.org/results/jdanek/skupper-router/epel-8-aarch64/05281012-skupper-router/builder-live.log.gz

I am wanting to dismiss this as just some timing issue on lowpowered build machine, but it happens _everywhere_!

The build results are at https://copr.fedorainfracloud.org/coprs/jdanek/skupper-router/build/5281012/, the `failed` states mean failed tests, not failed compilation

packaging/skupper-router.spec.rpkg
```diff
@@ -63,19 +65,28 @@ BuildRequires: gcc-c++
...
 BuildRequires: asciidoc
-BuildRequires: python3-qpid-proton >= %{proton_minimum_version}
 # check ctest
 BuildRequires: cyrus-sasl-plain
 
 # proton-c requirements
 BuildRequires: openssl-devel
 BuildRequires: cyrus-sasl-devel
+# python-qpid-proton requirements
+BuildRequires: swig
```
jiridanek, line 89: yeah, this is now out of date, proton python is using cffi; but early this week it still worked with main

ganeshmurthy:
> * Should I close this as rejected, based on discussion in [Set proton version in CI to 0.39.0 #1115 (comment)](https://github.com/skupperproject/skupper-router/pull/1115#issuecomment-1578856334)?

Yes please, thanks.

ganeshmurthy:
@jiridanek Can this PR can be closed in favor of https://github.com/skupperproject/skupper-router/pull/1120 ?

Issue #170, opened 2022-03-12
Consider requiring pytest to run system-tests


jiridanek:
There are some (considerable) advantages to pytest, over the currently used unittest module from the Python std. lib.

1. Compatibility with unittests module. No code changes are required to start with, only a small change in CMakeLists.txt to change the runner.
2. Fixtures, https://docs.pytest.org/en/latest/explanation/fixtures.html
3. conftest.py, https://stackoverflow.com/questions/34466027/in-pytest-what-is-the-use-of-conftest-py-files, allows e.g. set up test environment, maybe could be used to simplify https://github.com/skupperproject/skupper-router/blob/main/tests/run_system_tests.py and testing against installed router
4. marks on tests and excluding tests based on mark (marks are pytests name for tags in junit)
5. plugin to enforce test timeouts (on the level of python test methods, not only on ctest targets the way it is now)
6. more, https://realpython.com/pytest-python-testing/#what-makes-pytest-so-useful

Pytest is currently running tests in GitHub Actions.

jiridanek:
For example, [pytest-instafail](https://github.com/pytest-dev/pytest-instafail) and [pytest-timeout](https://github.com/pytest-dev/pytest-timeout).

kgiusti:
Are there any gaps in features?  In other words are there features we use in ctest that are not present in pytest?  Will things break and how much work is involved fixing/implementing the gaps?

jiridanek:
@kgiusti I meant replacing `unittest` with `pytest`. The `ctest` runner would remain, in my proposal.

PR #404, opened 2022-04-30
Issue #170 - Enable the pytest-instafail plugin in CI

jiridanek:
Here's what it looks like

https://github.com/jiridanek/skupper-router/runs/6239492635?check_suite_focus=true#step:26:3050

Sadly the pytest-instafail plugin removes printing of the results at the end, which makes it harder to find the results in the CI report. There is open issue for this, https://github.com/pytest-dev/pytest-instafail/issues/21.

So far, it seems doable to search for `***` or `***F` to get to the failed test, and then search for `_ _` or `_____` to find individual failure reports.


jiridanek:
This was interesting idea, but it's something that will never go to main. Closing.

PR #403, opened 2022-04-30
Issue #48 - Collect Python dependencies in requirements-dev.txt

jiridanek:

PR #407, opened 2022-04-30
Issue #170 - Enable the pytest-timeout plugin in CI

jiridanek:
This enables a per-test timeout in pytest. Looking at the results, seems that setup for `test_connected_tls_sasl_routers` in tls system tests wants to take >300 seconds? Anyways, when https://github.com/skupperproject/skupper-router/pull/403 is merged, this is rebased on top, and the ncat test is passing, the results will be more clear. Marking as draft until then.

See plugin docs at https://pypi.org/project/pytest-timeout/

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -136,7 +136,7 @@ jobs:
           cmake "${{github.workspace}}/skupper-router" \
             "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" \
             "-DCMAKE_BUILD_TYPE=${BuildType}" \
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -92,6 +92,21 @@ $ ./run.py
 
 === Test-only dependencies
 
+.Install all package dependencies at once (Fedora)
```
ganeshmurthy, line 95: Just a couple of minor suggestions - 
Just say Install package dependencies instead of Install all package dependencies at once
and say Install Python test dependencies from requirements-dev.txt instead of Install all Python test dependencies at once

ganeshmurthy (approved):

README.adoc
jiridanek: done in https://github.com/skupperproject/skupper-router/pull/403

jiridanek:
* Reported the test timeout as https://github.com/skupperproject/skupper-router/issues/420.

kgiusti (commented):
Let's hold off on this until we fix the other issues preventing this from passing CI

Issue #412, opened 2022-04-30
Python (shutdown) leaks


jiridanek:
Brought forward from [DISPATCH-1962](https://issues.apache.org/jira/browse/DISPATCH-1962) Make LSan suppressions more targeted and specific (within reason)

= Python leaks

Leaks in CPython itself are known and there is ongoing effort to fight them: https://bugs.python.org/issue1635741 (https://bugs.python.org/issue25302) and https://www.python.org/dev/peps/pep-3121

Here's valgrind suppression file from somebody who actually investigated the Python leaks and identified the harmless ones: https://github.com/libgit2/pygit2/blob/master/misc/valgrind-python.supp

= Router leaks

One example of a hidden leak in dispatch, which is revealed by making Python suppressions more targetted:

```
9: Direct leak of 56 byte(s) in 1 object(s) allocated from:
9:     #0 0x7f78a3606e8f in __interceptor_malloc (/nix/store/g40sl3zh3nv52vj0mrl4iki5iphh5ika-gcc-10.2.0-lib/lib/libasan.so.6+0xace8f)
9:     #1 0x7f78a2d64afb in qd_malloc ../include/qpid/dispatch/ctools.h:229
9:     #2 0x7f78a2d657da in qdr_core_subscribe ../src/router_core/route_tables.c:149
9:     #3 0x7f78a2c83072 in IoAdapter_init ../src/python_embedded.c:711
9:     #4 0x7f78a2353a6c in type_call (/nix/store/r85nxfnwiv45nbmf5yb60jj8ajim4m7w-python3-3.8.5/lib/libpython3.8.so.1.0+0x165a6c)
```

The problem is in

```python
class Agent:
    ...
    def activate(self, address):
        ...
        self.io = IoAdapter(self.receive, address, 'L', '0', TREATMENT_ANYCAST_CLOSEST)
```

`IoAdapter` refers to `Agent` (through the bound method reference `self.receive`) and `Agent` refers to `IoAdapter` (through property `self.io`). Since `IoAdapter` is implemented in C and does not implement support for Python's cyclic GC, there is no way to break the cycle, collect the garbage, and run destructors.

Heap dump in attachment of the original Jira. The bound method is at the top of the picture. (Ignore the `Mock` objects, I was trying to simplify the picture while not getting crashes due to too much meddling).

== Random observations

[...]

PR #413, opened 2022-04-30
Fixes #254, #412 - Use PyCapsule to manage a C pointer's lifecycle and have it GC'd at shutdown

jiridanek:
* This is revival of https://github.com/skupperproject/skupper-router/pull/259
* It attempts to fix issue described in https://github.com/skupperproject/skupper-router/issues/412
* and by the way it seems to also fix https://github.com/skupperproject/skupper-router/issues/253

## About capsules and men

The Capsule is a container for a pointer, opaque from the Python side, and translucent from the C side. The C code can put a pointer in and Python can treat it as py_object and pass it to some other C code, but cannot unwrap the pointer (unless it really tries, see example code on Stack Overflow for how to break the encapsulation). The important thing for our purposes is that Capsules can own the pointer and can have a destructor associated. More at https://docs.python.org/3/c-api/capsule.html

The idea of the #253 fix is to use Capsule to associate destructor to the pointer, and then rely on Python to run GC eventually and free the memory. Thereby leak is prevented.

## Getting the GC to actually run

Capsule is owned by `AppStats`, which is owned by `PolicyLocal`, which is contained in `PolicyManager` which is contained in `Agent`, which is very bad, because `Agent` cannot be GC'd because there is a refcounting cycle with `IoAdapter`, which does not participate in the cyclic GC. The Dispatch PR which tried to fix the GC for IoAdapter hit into weird shutdown crashes because when the Python was actually running GC fully (which it was not before) it hit... issues.

## In conclusion

I considered writing a TODO list here, but I guess it would be best to just discuss this first... Honestly, I am not sure why I am not seeing the horrible problems that plagued the Dispatch PR. And I feel uneasy about it.

One thing I had to do in Dispatch that does not seem necessary here is this (in python_embedded.c::qd_io_rx_handler)

```diff
diff --git a/src/python_embedded.c b/src/python_embedded.c
--- a/src/python_embedded.c	(revision 922ca0ad7d43f2d838c4bbb282d69e14b8b457c3)
+++ b/src/python_embedded.c	(revision 714e506d601054eb839e5968caf01c81afa1c3f0)
@@ -639,6 +639,11 @@
     IoAdapter *self = (IoAdapter*) context;
     *error = 0;
 
+    if (self->handler == NULL) {
+        *error = qdr_error(QD_AMQP_COND_INTERNAL_ERROR, "Router is shutting down");
+        return PN_REJECTED;
+    }
+
```

I'm going to try to somehow stress-test this. And also try python-dbg variant to see if there aren't any decref issues.

### DIFF
--- a/src/policy.c
+++ b/src/policy.c
@@ -191,26 +191,23 @@ qd_error_t qd_register_policy_manager(qd_policy_t *policy, void *policy_manager)
 }
 
 
-long qd_policy_c_counts_alloc()
+void *qd_policy_c_counts_alloc()
 {
-    qd_policy_denial_counts_t * dc = NEW(qd_policy_denial_counts_t);
+    qd_policy_denial_counts_t *dc = NEW(qd_policy_denial_counts_t);
     assert(dc);
     ZERO(dc);
-    return (long)dc;
+    return dc;
 }
 
-
-void qd_policy_c_counts_free(long ccounts)
+void qd_policy_c_counts_free(void *dc)
 {
-    void *dc = (void *)ccounts;
     assert(dc);
     free(dc);
 }
 
 
-qd_error_t qd_policy_c_counts_refresh(long ccounts, qd_entity_t *entity)
+qd_error_t qd_policy_c_counts_refresh(qd_policy_denial_counts_t* dc, qd_entity_t *entity)
 {
-    qd_policy_denial_counts_t *dc = (qd_policy_denial_counts_t*)ccounts;
     if (!qd_entity_set_long(entity, "sessionDenied", dc->sessionDenied) &&
         !qd_entity_set_long(entity, "senderDenied", dc->senderDenied) &&
         !qd_entity_set_long(entity, "receiverDenied", dc->receiverDenied) &&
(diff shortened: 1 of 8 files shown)


jiridanek:
Trying Stuff Until it works

src/policy.c
```diff
@@ -191,26 +191,23 @@ qd_error_t qd_register_policy_manager(qd_policy_t *policy, void *policy_manager)
 }
 
 
-long qd_policy_c_counts_alloc()
+void *qd_policy_c_counts_alloc()
 {
-    qd_policy_denial_counts_t * dc = NEW(qd_policy_denial_counts_t);
+    qd_policy_denial_counts_t *dc = NEW(qd_policy_denial_counts_t);
```
jiridanek, line 196: What's the convention regarding that space, actually?

kgiusti: Good question! C type syntax is a mess.  I don't think we've formally established a convention here, but personally I put the "star" right next to the var name. I do this because "star" binds right-to-left, which means it binds with the var, not the type.  So if you wrote this "c" code:

int* x, y; 

it seems like the code is defining two pointers to int, but in fact the * binds with the x, so you actually get a pointer to int (x) and an int (y).

Doing it this way:

int *x, y;

makes it clearer that x is the pointer.  Not that you'd actually write that code, but it is legal C.

jiridanek:
> So was qd_dispatch_policy_c_counts_free() never called from python?

Yes, I was not called.

src/policy.c
jiridanek: `clang-format` all the things!

jiridanek:
```
% gh workflow run --repo jiridanek/skupper-router --ref jd_2022_04_30_py_capsule build.yaml \
      -f protonBranch=0a770196301376069bcddf5c624cb3726b3f4a40
```

I'm hoping to see all blue, to feel better about this https://github.com/jiridanek/skupper-router/actions/runs/2389565428

Issue #417, opened 2022-05-02
test_head_request failed in system_tests_http2_tls in flaky manner `AssertionError: False is not true`


jiridanek:
https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/568809767#L6092

This is Travis CI ppc64le, but I saw it on GitHub Actions as well. (On my PRs where it was not 100% certain that the PR did not broke things.) When I see it again, i'll add link to GHA because that has downloadable logs.

```
60: ======================================================================
60: FAIL: test_head_request (system_tests_http2_tls.Http2TlsAuthenticatePeerOneRouter)
60: ----------------------------------------------------------------------
60: Traceback (most recent call last):
60:   File "/home/travis/build/jiridanek/skupper-router/tests/system_tests_http2_tls.py", line 568, in test_head_request
60:     self.assertTrue(conn_closed)
60: AssertionError: False is not true
60: 
60: ----------------------------------------------------------------------
60: Ran 65 tests in 112.799s
60: 
60: FAILED (failures=1, skipped=5)
60/67 Test #60: system_tests_http2_tls ..................***Failed  131.98 sec
```

jiridanek:
Ask, and Travis shall provideth, https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/568978643#L10065

```
60: ======================================================================
60: FAIL: test_head_request (system_tests_http2_tls.Http2TlsAuthenticatePeerOneRouter)
60: ----------------------------------------------------------------------
60: Traceback (most recent call last):
60:   File "/home/travis/build/jiridanek/skupper-router/tests/system_tests_http2_tls.py", line 567, in test_head_request
60:     self.assertTrue(conn_closed, msg="err is %s and out is %s" % (err, out))
60: AssertionError: False is not true : err is curl: (52) Empty reply from server
60: 
60:  and out is 
60: 
60: ----------------------------------------------------------------------
60: Ran 65 tests in 161.666s
60: 
60: FAILED (failures=1, skipped=5)
60/67 Test #60: system_tests_http2_tls ..................***Failed  189.59 sec
```

ganeshmurthy:
Yikes. When run on the ppc64le CPU architecture, the err is `curl: (52) Empty reply from server`
That is not what I expected. Does this test fail *every time* it is run under that particular architecture on Travis ?

jiridanek:
No idea, let me push empty commit (or two), we will see:

[...]

PR #422, opened 2022-05-03
#417: Print the err and the out. This is not a fix for the issue

ganeshmurthy:


jiridanek:
Let's see if it fails on Travis CI, https://app.travis-ci.com/github/jiridanek/skupper-router/builds/250170589

ganeshmurthy:
This is a good to have

Issue #420, opened 2022-05-03
RouterTestSslInterRouter::test_connected_tls_sasl_routers, part of system_tests_ssl, takes a long time (>300s)


jiridanek:
https://github.com/skupperproject/skupper-router/runs/6253209682?check_suite_focus=true#step:26:3238

Note the time difference between previous passing test at 06:48:43 and test_connected_tls_sasl_routers passing at 06:53:52.

```
2022-05-02T06:48:43.2487811Z 40: ::RouterTestSslClient::test_tls_all [32mPASSED[0m
2022-05-02T06:48:43.2492787Z 40: Router QDR.A output file:
2022-05-02T06:48:43.2496696Z 40: >>>>
2022-05-02T06:48:43.2500996Z 40: -----------------------------------------------------
2022-05-02T06:48:43.2504875Z 40: Suppressions used:
2022-05-02T06:48:43.2508657Z 40:   count      bytes template
2022-05-02T06:48:43.2512522Z 40:       4        224 ^IoAdapter_init$
2022-05-02T06:48:43.2516288Z 40:     720     708025 /libpython3.*.so
2022-05-02T06:48:43.2520386Z 40: -----------------------------------------------------
2022-05-02T06:48:43.2524265Z 40: 
2022-05-02T06:48:43.2565966Z 40: 
2022-05-02T06:48:43.2566201Z 40: <<<<
2022-05-02T06:48:43.2566398Z 40: 

[...]

2022-05-02T06:53:52.3057972Z 40: ::RouterTestSslInterRouter::test_connected_tls_sasl_routers [32mPASSED[0m
```

[...]

```
32/32 Test #40: system_tests_ssl ......................   Passed  325.81 sec
```

It seems an invariable rule that the ssl tests always finish last.

After setting a 300s timeout per test (here meaning python test method), which is IMO still overly generous, the test gets reported as exceeding that timeout

https://github.com/skupperproject/skupper-router/runs/6263670044?check_suite_focus=true#step:26:3841

```
        # Wait till listener is running and all expected connectors are connected
        cls.routers[0].wait_ports()
        for router in cls.connected_tls_sasl_routers[1:]:
>           cls.routers[0].wait_router_connected(router)

/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_ssl.py:744: 
[...]

PR #423, opened 2022-05-03
#420: Added some prints to report time it takes for router QDR.A to c…

ganeshmurthy:
…onnect to the other routers. This is not a fix for the issue


ganeshmurthy:
This is useful to have in case of future failures in this test

PR #431, opened 2022-05-04
Fixes #430 - Perform dlopen entirely on the Python side

### LINKED ISSUE #430: Reconsider DISPATCH-87 design choices about dlopen in skrouterd
The [DISPATCH-87 Double load of libqpid-dispatch.so](https://issues.apache.org/jira/browse/DISPATCH-87) fix protects against the situation when the qdrouterd binary loads a different libqpid-dispatch.so by means of the dynamic linker than what Python then loads through ctypes.

The fix was implemented by @alanconway in de480acf8b2a2cc9a4b6cf502bc4ab7668170718 commit, see commit message for the fix idea.

Since there is no shared library at this point, it should be possible to tell ctypes to either load `NULL`, like in the C dlopen API, or maybe `/proc/self/exe`, or whatever the Python way of referring to self is. That means obtaining the handle on the C side and the handoff of it to Python can be avoided.

jiridanek:
Fixes #430

### DIFF
--- a/python/skupper_router_internal/dispatch.py
+++ b/python/skupper_router_internal/dispatch.py
@@ -49,8 +49,9 @@ class QdDll(ctypes.PyDLL):
     internally makes python calls.
     """
 
-    def __init__(self, handle: int) -> None:
-        super(QdDll, self).__init__("skupper-router", handle=handle)
+    def __init__(self) -> None:
+        # `dlopen(NULL, ...)` opens the current executable; the router used to dlopen libqpid-dispatch.so before
+        super().__init__(name=None)
 
         # Types
         self.qd_dispatch_p = ctypes.c_void_p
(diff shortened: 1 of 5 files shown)


kgiusti (commented):
LGTM but let's hear from @astitcher as well

jiridanek:
> @astitcher One general comment is that I think we are now python 3 only so we could (should?) remove the .pyi files and shift the type annotations directly into the .py files.

The `dispatch.pyi` are type annotations for the C module. There is a test-only `dispatch.py` that is used to unittest Python code, but the real implementation of `dispatch` module is in C, not Python 3.

jiridanek:
And yeah, I am not sure why the `.pyi` also has type annotations for `QdDll`, that indeed looks out of place.

Issue #260, opened 2022-03-30
[perf] Do not destroy lists by repeatedly biting off the head


jiridanek:
Biting off the head is convenient way of iterating through a list while deleting its elements. The disadvantage is that it performs more operations than is necessary. It takes care of repeatedly adjusting the list pointers, even though ultimately the entire list is going away.

When @ssorj showed the per-instruction profiling costs, a lot of it was concentrated on `mov`s. So, let's not mov uselessly.

Instead of

```c
void qd_buffer_list_free_buffers(qd_buffer_list_t *list)
{
    qd_buffer_t *buf = DEQ_HEAD(*list);
    while (buf) {
        DEQ_REMOVE_HEAD(*list);
        qd_buffer_free(buf);
        buf = DEQ_HEAD(*list);
    }
}
```

Do this

```c
void qd_buffer_list_free_buffers(qd_buffer_list_t *list)
{
    qd_buffer_t *buf = list->head;
    while (buf != NULL) {
        qd_buffer_t *next = buf->next;
        qd_buffer_free(buf);
        buf = next;
    }
    ZERO(list);
}
```

There can be helper macro to wrap the deletion, that would take the for-each deallocating function.

Looking at the generated code, the difference between the two looks like this

```diff
--- <unnamed>
+++ <unnamed>
@@ -3,60 +3,35 @@
   43a650:       41 56                   push   %r14
   43a652:       53                      push   %rbx
   43a653:       50                      push   %rax
-  43a654:       48 8b 17                mov    (%rdi),%rdx
-    qd_buffer_t *buf = ((*list).head);
-    while (buf) {
-  43a657:       48 85 d2                test   %rdx,%rdx
-  43a65a:       74 5a                   je     43a6b6 <qd_buffer_list_free_buffers+0x66>
-  43a65c:       48 89 fb                mov    %rdi,%rbx
-  43a65f:       4c 8d 77 08             lea    0x8(%rdi),%r14
-  43a663:       66 2e 0f 1f 84 00 00    cs nopw 0x0(%rax,%rax,1)
-  43a66a:       00 00 00 
-  43a66d:       0f 1f 00                nopl   (%rax)
-        do {
-            {
-                ((void) (0));
-            };
-            if ((*list).head) {
-                (*list).scratch = (*list).head;
-  43a670:       48 89 53 10             mov    %rdx,0x10(%rbx)
-                (*list).head    = (*list).head->next;
[...]

PR #432, opened 2022-05-04
Issue #260 - Remove from DEQ without updating shared `scratch` variable

jiridanek:
This uses the `__typeof__` operator to declare the variable in the macro and compute correct type for it.


jiridanek:
/me realizing this would work just fine with `void*` instead of the `__typeof__`

jiridanek:
/me realizing this would work just fine with `void*` instead of the `__typeof__`

kgiusti (approved):

PR #433, opened 2022-05-04
Issue #260 - Avoid `DEQ_REMOVE_HEAD` to make list deletion in `qd_buffer_list_free_buffers` more efficient

jiridanek:
See the issue for the asm diff. 

There are other places that use the inefficient deletion, but they did not show on the cpu profile. It is possible to create a macro for these deletions, with the function to call to free item as a macro parameter. Some list deletions perform a more complicated freeing, however, and there such macro is not sufficient.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -86,11 +86,12 @@ unsigned int qd_buffer_list_clone(qd_buffer_list_t *dst, const qd_buffer_list_t 
 void qd_buffer_list_free_buffers(qd_buffer_list_t *list)
 {
     qd_buffer_t *buf = DEQ_HEAD(*list);
-    while (buf) {
-        DEQ_REMOVE_HEAD(*list);
+    while (buf != NULL) {
+        qd_buffer_t *next = DEQ_NEXT(buf);
         qd_buffer_free(buf);
-        buf = DEQ_HEAD(*list);
+        buf = next;
     }
+    DEQ_INIT(*list);
 }
 
 
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -86,7 +86,7 @@ unsigned int qd_buffer_list_clone(qd_buffer_list_t *dst, const qd_buffer_list_t 
 void qd_buffer_list_free_buffers(qd_buffer_list_t *list)
 {
     qd_buffer_t *buf = DEQ_HEAD(*list);
-    while (buf != NULL) {
+    while (buf) {
         qd_buffer_t *next = DEQ_NEXT(buf);
         qd_buffer_free(buf);
         buf = next;


src/buffer.c
```diff
@@ -86,11 +86,12 @@ unsigned int qd_buffer_list_clone(qd_buffer_list_t *dst, const qd_buffer_list_t
 void qd_buffer_list_free_buffers(qd_buffer_list_t *list)
 {
     qd_buffer_t *buf = DEQ_HEAD(*list);
-    while (buf) {
-        DEQ_REMOVE_HEAD(*list);
+    while (buf != NULL) {
```
ganeshmurthy, line 89: is there a difference in performance between `while (buf)` and `while (buf != NULL)` ?
Just curious

jiridanek: It has never occured to me to investigate that. I would strongly expect not, both should compile to the same machine instructions.

The change is my style preference. I am used to p !=NULL because I want to pass a result of a bool expression to the if. I saw that Teď likes to do !!p In these situations. That I remember from JavaScript, I never saw that done in C.

If I recall correctly, clang-tidy prefers !=NULL in these situations.

I'd of course do what's standard for router, if you tell me there is such convention.

jiridanek: Looks to be the same in all cases, both according to objdump and godbolt, https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:24,positionColumn:1,positionLineNumber:24,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'%23include+%3Cstdlib.h%3E%0A%23include+%3Cstdio.h%3E%0A%0Avoid+if_p()+%7B%0A++++void+*+p+%3D+malloc(42)%3B%0A++++if+(p)+%7B%0A++++++++printf(%22got+a+p%5Cn%22)%3B%0A++++%7D%0A%7D%0A%0Avoid+if_bang_bang_p()+%7B%0A++++void+*+p+%3D+malloc(42)%3B%0A++++if+(!!!!p)+%7B%0A++++++++printf(%22got+a+p%5Cn%22)%3B%0A++++%7D%0A%7D%0A%0Avoid+if_p_neq_null()+%7B%0A++++void+*+p+%3D+malloc(42)%3B%0A++++if+(p+!!%3D+NULL)+%7B%0A++++++++printf(%22got+a+p%5Cn%22)%3B%0A++++%7D%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:49.99999999999999,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cg113,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2+-g',selection:(endColumn:32,endLineNumber:9,positionColumn:32,positionLineNumber:9,selectionStartColumn:32,selectionStartLineNumber:9,startColumn:32,startLineNumber:9),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.3+(C,+Editor+%231,+Compiler+%231)',t:'0')),k:49.99999999999999,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

ganeshmurthy: If there is not much difference between `while (buf)` and `while (buf != NULL)` then I would recommend keeping the former. I don't have a preference either way and there is no set convention regarding this but when I look at this commit which is trying to optimize this function, I don't want to be left wondering if one convention is faster than the other. In other words, when I look at this commit, I want to see the exact change that made the code faster and nothing else. Thanks.

jiridanek: Done. The performance fix is avoiding `DEQ_REMOVE_HEAD`.

ganeshmurthy (approved):

src/buffer.c
ganeshmurthy: > Done. The performance fix is avoiding `DEQ_REMOVE_HEAD`.

Approved ! I learned something. Thanks.

kgiusti (approved):
+1 nice!

PR #435, opened 2022-05-05
Fixes #120: control the lifecycle of the tcpListener socket on the availability of the tcpConnector

### LINKED ISSUE #120: TCP adaptor listener connections hang if no connectors are defined
If a connector is not defined but a listener is defined:

    A client connects to the listener. Listener starts with SYN, SYN/ACK, ACK.
    The client sends the first payload bytes
    No further traffic is passed on the client connection
    Even if the server is started while the client is waiting no further traffic is passed on to the client.
    The client seems to hang

If there is no address in the address table, close the client connection. Also free all the connection related resources in the skupper-router.

kgiusti:


kgiusti:
DO NOT MERGE:

Disables some tests that fail - needs fixing before it can land.

src/adaptors/tcp_adaptor.c
```diff
@@ -1316,22 +1375,96 @@ static qd_tcp_listener_t *qd_tcp_listener(qd_server_t *server)
...
+            sys_atomic_inc(&li->ref_count);
+
+        sys_mutex_unlock(tcp_adaptor->listener_lock);
+
+        if (li) {
+
+            qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG,
+                   "Listener address %s consumer count updates: local=%"PRIu32" in-process=%"PRIu32" remote=%"PRIu32,
+                   li->config->address, local_consumers, in_proc_consumers, remote_consumers);
+
```
ted-ross, line 1418: I think you could safely ignore in-process consumers here.

ted-ross (approved):

PR #438, opened 2022-05-05
Fixes #201: Give some additional time for the httpRequestInfo to upda…

ganeshmurthy:
…te since it is handled by the core thread

PR #439, opened 2022-05-06
Fixes #446 - Refactor `retry_exception` into less-forgiving general retry (of same name) and `retry_assertion` test helper function

### LINKED ISSUE #446: Refactor retry_exception into less-forgiving general retry (of same name) and retry_assertion test helper function

jiridanek:
The old `retry_exception` was too forgiving regarding `SyntaxError` and other such similar exceptions.
These should not trigger retry, as they are likely an unintentional programming error.

The case of retrying on `AssertionError` is now to be handled by `retry_assertion`.

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -186,7 +186,7 @@ def retry_exception(
     while True:
         try:
             return function()
-        except exception as e:
+        except exception as e:  # pylint: disable=misc  # error: Exception type must be derived from BaseException
             if exception_test:
                 exception_test(e)
             delay = retry_delay(deadline, delay, max_delay)
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -50,7 +50,7 @@ from datetime import datetime
 from subprocess import PIPE, STDOUT, TimeoutExpired
 from threading import Event
 from threading import Thread
-from typing import Callable, TextIO, List, Optional, Tuple
+from typing import Callable, TextIO, List, Optional, Tuple, Union, Type, Set, Any, TypeVar
 
 import __main__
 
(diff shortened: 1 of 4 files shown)


tests/system_test.py
```diff
@@ -154,6 +154,14 @@ def retry_exception(function, timeout=TIMEOUT, delay=.001, max_delay=1, exceptio
     Returns what function returns if it succeeds before timeout.
     Raises last exception raised by function on timeout.
     """
+
+    def default_exception_test(exception: Exception):
+        """Do not permit retry on exceptions usually caused by programmer error. The list might be incomplete."""
+        if isinstance(exception, (SyntaxError, UnboundLocalError, RecursionError, ImportError, TypeError)):
+            raise exception
+
+    exception_test = exception_test or default_exception_test
```
ganeshmurthy, line 163: It seems to me that if the user passes in an exception_test, we will use it, otherwise we will do the default_exception_test. It seems like a good idea to always call the default_exception_test() irrespective of whether an exception_test is passed in or not. By that I mean that the retry_exception function should fail immediately if the exception thrown is any instance of the exceptions mentioned in default_exception_test which indicates user error.

jiridanek: The `exception_test` function has been so far implemented as an allow-list (whitelist) as far as I examined usages. When this is done, then you don't need any default set like this.

Sometimes you might want to retry on `TypeError`, I am thinking, and with your modification this would not be possible because function would throw on TypeError always.

ganeshmurthy: oh ok, I always thought of the exception_test as an exception handler function. So, if an exception is thrown inside the retry_exception function it would call exception_test with the exception so you can further analyze/process the thrown exception

ganeshmurthy:
I was wondering what if retry_exception took an exception parameter. Like so - 
```
def retry_exception(function, timeout=TIMEOUT, delay=.001, max_delay=1, exception_test=None, exception=None, **kwargs):
```

If an exception is passed in, only that exception would be caught (currently the `Exception` is caught). That way we could completely eliminate the `retry_assertion` function and just call the `retry_exception` with the appropriate exception. Or we could keep the `retry_assertion` function and it would just pass the exception parameter as `AssertionError`

jiridanek:
> I was wondering what if retry_exception took an exception parameter. Like so -
> 
> ```
> def retry_exception(function, timeout=TIMEOUT, delay=.001, max_delay=1, exception_test=None, exception=None, **kwargs):
> ```
> 
> If an exception is passed in, only that exception would be caught (currently the `Exception` is caught). That way we could completely eliminate the `retry_assertion` function and just call the `retry_exception` with the appropriate exception. Or we could keep the `retry_assertion` function and it would just pass the exception parameter as `AssertionError`

Makes sense, I'll do it this way, thanks

ganeshmurthy:
LGTM. Can you please create a new issue and commit this change against the new issue ?

ganeshmurthy (approved):
Once you commit this to main, I will make the change in this PR - https://github.com/skupperproject/skupper-router/pull/438

PR #447, opened 2022-05-07
Fixes #440: Replace distutils with setuptools

### LINKED ISSUE #440: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```
56:   /usr/local/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:168: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

This usage is of course in a transitive dependency so we are not responsible for fixing this, but there are other usages of `distutils` directly in skupper router.

The other deprecation warning that is around is filed as issue https://github.com/skupperproject/skupper-router/issues/414.

jiridanek:
This is straightforward replacement of one package with another. It does not give any thought to getting rid of the CMake preprocessing (`.py.in` files`) or to the ideas behind installing Python code into site-packages in the first place.

I am certainly open to looking into those and maybe doing something about it, instead of just replacing one package with another in response to a deprecation warning.

The top level config.sh and build.sh should be replaced too, with some kind of "project orchestration tool", maybe just a Makefile?

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -17,8 +17,6 @@
 # under the License.
 #
 
-set -Eeuo pipefail
-
 if [[ ! -f config.sh ]]; then
     echo "You must source config.sh from within its own directory"
     return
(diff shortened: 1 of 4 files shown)


astitcher (commented):
LGTM

Issue #450, opened 2022-05-09
system_tests_topology failing again


ganeshmurthy:
This time it is failing with a different error message - 
```
=========================================================================== FAILURES ===========================================================================
___________________________________________________________ TopologyTests.test_01_topology_failover ____________________________________________________________

self = <tests.system_tests_topology.TopologyTests testMethod=test_01_topology_failover>

    def test_01_topology_failover(self):
        name = 'test_01'
        if self.skip[name]:
            self.skipTest("Test skipped during development.")
        test = TopologyFailover(name,
                                self.client_addrs,
                                "closest/01"
                                )
        test.run()
>       self.assertIsNone(test.error)
E       AssertionError: 'Timeout Expired: n_sent=0 n_received=0 n_accepted=6' is not None

/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_topology.py:310: AssertionError
```

https://github.com/skupperproject/skupper-router/runs/6334169830?check_suite_focus=true#step:26:3589

jiridanek:
[test_01_topology_failover](https://github.com/skupperproject/skupper-router/runs/6833387685?check_suite_focus=true#step:27:1678)

kgiusti:
Seen recently on main: TopologyDispositionTests::test_01_delete_spurious_connector - AssertionError: 'No confirmed kill on connector.' is not None

kgiusti:
Possibly related to #612

ganeshmurthy:
```
 35: self = <tests.system_tests_topology.TopologyTests testMethod=test_01_topology_failover>
35: 
35:     def test_01_topology_failover(self):
35:         name = 'test_01'
35:         if self.skip[name]:
35:             self.skipTest("Test skipped during development.")
35:         test = TopologyFailover(self.client_addrs,
35:                                 "closest/01"
35:                                 )
35:         test.run()
35: >       self.assertIsNone(test.error)
35: E       AssertionError: 'Timeout Expired: n_sent=0 n_received=0 n_accepted=6' is not None
35: 
35: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_topology.py:309: AssertionError
```
https://github.com/skupperproject/skupper-router/actions/runs/7700489518/job/20984455875?pr=1389#step:34:5854

ganeshmurthy:
Not seen this in 6 months, closing

PR #451, opened 2022-05-09
#450: Turned on debug logging, changed send method name, removed unus…

ganeshmurthy:
…ed test name


ganeshmurthy:
Merging this to main, so it will run on each commit and we can see the debug log output if the test fails again.

PR #456, opened 2022-05-09
Issue # 420 - Decrease per-test pytest timeout to 100

jiridanek:


ganeshmurthy:
@jiridanek do you still want to do this ? Going from 400 to 100 ? I would rather not do this and find out that we need to move it back up again ?

jiridanek:
> @jiridanek do you still want to do this ? Going from 400 to 100 ? I would rather not do this and find out that we need to move it back up again ?

@ganeshmurthy This timeout is the default per-test timeout in pytest, so if you write a single python test method that takes more than 100 seconds, you can

1. split it into multiple test methods
2. add some pytest annotation to extend the timeout for this specific test method

IIRC, when this PR was newly created, there was one or two tests exceeding the 100 seconds, and I suspected in both cases it is because of some test bug, I don't exactly recall what it was. Looks like there is a few tests going over the 100 seconds.

I'll try to look why they take that much time and if they truly need it, or if there is something wrong going on.

PR #458, opened 2022-05-10
Fixes #452 - Better document dispatch.py to explain why .pyi is needed

### LINKED ISSUE #452: Remove out-of-place annotations from `dispatch.pyi`
This lgtm.

jiridanek:


python/skupper_router_internal/dispatch.pyi
```diff
@@ -17,7 +17,13 @@
 # under the License
 #
 
-"""Type stubs for objects implemented in the C extension module"""
+"""Type stubs for objects implemented in the C extension module.
```
ganeshmurthy, line 20: Perhaps mention here that the definition for IoAdaptor and LogAdaptor can be found in python_embedded.c ?

ganeshmurthy (approved):

python/skupper_router_internal/dispatch.pyi
jiridanek: I wrote that in `dispatch.py` already; too bad there does not seem to be one obvious place for the information, due to the nature of .pyi.

Analogy with C would suggest to put info into header file, that is pyi, but in my experience mostly documentation comment in python is in py files.

PR #463, opened 2022-05-10
Fixes #452: Get rid of dispatch.pyi. Moved stubs into dispatch.py

### LINKED ISSUE #452: Remove out-of-place annotations from `dispatch.pyi`

ganeshmurthy:


jiridanek:
I'm honestly surprised this can be made to work, but yeah, it seems to work!

Instead of the `# type: ignore`, I'd actually correct the type annotations. Mypy does not like when you want to reuse a variable with different type. The following patch passed tests for me

```diff
Index: python/skupper_router_internal/management/config.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/python/skupper_router_internal/management/config.py b/python/skupper_router_internal/management/config.py
--- a/python/skupper_router_internal/management/config.py	(revision b1725f7d327bfa991dfa23df2f5afafe11897767)
+++ b/python/skupper_router_internal/management/config.py	(revision 5040318d4d8d6fafdd52f366a62f8e6f5d3ecbb4)
@@ -32,6 +32,7 @@
 from .qdrouter import QdSchema
 
 if TYPE_CHECKING:
+    from ctypes import c_void_p
     from .schema import EntityType
 
 try:
@@ -277,10 +278,10 @@
         return [s for s in self.config_types if 'policy' in s.name]
 
 
-def configure_dispatch(dispatch: int, filename: str) -> None:
+def configure_dispatch(dispatch_int: int, filename: str) -> None:
     """Called by C router code to load configuration file and do configuration"""
     qd = QdDll()
-    dispatch = qd.qd_dispatch_p(dispatch)  # type: ignore[assignment]
+    dispatch: 'c_void_p' = qd.qd_dispatch_p(dispatch_int)
     config = Config(filename)
 
     # NOTE: Can't import agent until dispatch C extension module is initialized.
```

The module comment in `dispatch.py` is outdated and should probably mention `embedded_python.c`, which implements `IoAdaptor` and `LogAdaptor`, and it also implements the constants.

One reservation I still have is that if somehow something broke `embedded_python.c` and it did not replace the Adapter class stubs with actual implementations, then the stubs would serve as no-op implementations. If instead the classes were defined in a separate pyi, then we'd get an `ImportError` when trying to instantiate the classes.

Consider this program, which tries to use a type stub and which runs "just fine"

```python
class A:
    def __init__(self) -> None:
        ...
    def baf(self) -> int:
        ...

a = A()
a.baf()
```

ganeshmurthy:
> I'd actually correct the type annotations

Thanks, I fixed this.

ganeshmurthy:
> The module comment in `dispatch.py` is outdated and should probably mention `embedded_python.c`

I stole your comments from the other PR and added a bit more

ganeshmurthy:
> somehow something broke `embedded_python.c`

If this happened, the log module would not function properly and we would not see the logs we usually expect to see ?

jiridanek:
> > somehow something broke `embedded_python.c`
> 
> If this happened, the log module would not function properly and we would not see the logs we usually expect to see ?

When I took this PR and broke it intentionally, the tests started to timeout, without any obvious error message pointing out the cause, https://github.com/jiridanek/skupper-router/runs/6383742307?check_suite_focus=true

Not that anything like this is likely to happen. I don't think getting rid of one `pyi` file is worth this, but that is a matter of opinion, of course. It is not too bad.

jiridanek:
LGTM, on the whole!

kgiusti (approved):

jiridanek:
This obsoleted https://github.com/skupperproject/skupper-router/pull/458

PR #466, opened 2022-05-12
Combine adaptors

### LINKED ISSUE #487: Create an adaptor_common.c and adaptor_common.h and move any common code used by the adaptors

ganeshmurthy:


ganeshmurthy:
This PR tries to get the config load of the adaptors into a common area. A couple of things to consider
* Is this change complex ? Is it tough to understand ? 
* Does it make sense to introduce this change ?
* This common code is the first of more common changes to come. The TLS common code will live in this common area.

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,92 @@
...
+    QD_AGGREGATION_MULTIPART
+} qd_http_aggregation_t;
+
+// Currently we only have TCP and HTTP (HTTP/1 and HTTP/2) adaptors
+typedef enum {
+    TCP,
+    HTTP
+} qd_adaptor_type_t;
+
+typedef struct qd_bridge_config_t qd_bridge_config_t;
```
ted-ross, line 48: Is "bridge config" the right name for this entity?  "Bridge" is not a concept/term that we use generally.

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,92 @@
...
+    QD_AGGREGATION_NONE,
+    QD_AGGREGATION_JSON,
+    QD_AGGREGATION_MULTIPART
+} qd_http_aggregation_t;
+
+// Currently we only have TCP and HTTP (HTTP/1 and HTTP/2) adaptors
+typedef enum {
+    TCP,
+    HTTP
+} qd_adaptor_type_t;
```
ted-ross, line 46: Why do we separate the HTTP v1 and v2 as sub-categories of HTTP?  Aren't HTTP1 and HTTP2 completely separate adaptors?  Might it be simpler so just have HTTP1 and HTTP2 enums and remove the version?

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,92 @@
...
+typedef struct qd_bridge_config_t qd_bridge_config_t;
+
+struct qd_bridge_config_t
+{
+    char              *name;
+    char              *host;
+    char              *port;
+    char              *address;
+    char              *site_id;
+    char              *host_port;
```
ted-ross, line 57: host_port appears to the redundant with host and port.  Does the client really need to provide both forms of this data?

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,92 @@
...
+           qd_http_version_t      version;
+           bool                   event_channel;
+           qd_http_aggregation_t  aggregation;
+           char                  *host_override;
+       } http;
+
+       struct {
+        // run time statistics updated by connections
+        sys_mutex_t  *stats_lock;
+        uint64_t      connections_opened;
```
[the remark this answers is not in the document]
ganeshmurthy: That is actually a good question. Not sure why the stats need to be in the config structure. I suggest that we move to a stats model like the one used in the HTTP adaptors where a *record* is inserted with each request. This common function is called from both HTTP1 and HTTP2 adaptors - https://github.com/skupperproject/skupper-router/blob/main/src/adaptors/http_common.c#L692

I don't want to make that change as part of this commit but want to do it in a totally new commit.

src/adaptors/adaptor_common.h
ganeshmurthy: We could perhaps just call it qd_config_t instead of qd_bridge_config_t ?

src/adaptors/adaptor_common.h
ganeshmurthy: When it comes to handling of config and creation of listeners and connectors, the HTTP1 and the HTTP2 adaptors behave exactly the same, so at this point in time, having two adaptor types seemed to be enough, but if the need arises, we should remove HTTP and add HTTP1 and HTTP2

src/adaptors/adaptor_common.h
ganeshmurthy: host_port is set using host and port as seen here - https://github.com/skupperproject/skupper-router/blob/main/src/adaptors/tcp_adaptor.c#L1234

src/adaptors/adaptor_common.h
ted-ross: I would say that TCP behaves exactly the same as well.  The event-channel stuff is optional configuration that could just as well apply to TCP.

src/adaptors/adaptor_common.h
ted-ross: Too general.  Would `qd_adaptor_config_t` be incorrect?

ganeshmurthy: Done. Renamed to qd_adaptor_config_t

src/adaptors/adaptor_common.h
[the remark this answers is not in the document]
kgiusti: Yeah I agree with ted here - in general we should keep stats separate from config data.  This allows config data to be passed as read-only (const reference) as often as possible.

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,114 @@
...
+        uint64_t      connections_closed;
+        uint64_t      bytes_in;
+        uint64_t      bytes_out;
+        sys_atomic_t  ref_count;
```
kgiusti, line 83: What is "ref_count" counting?  If ref_count is counting references to the entire qd_adaptor_config_t then this ref_count should be moved out of the union and made a top level field.

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,114 @@
...
+ALLOC_DECLARE(qd_adaptor_config_t);
+
+qd_error_t qd_load_adaptor_config(qd_dispatch_t *qd, qd_adaptor_config_t *config, qd_entity_t* entity, qd_adaptor_type_t adaptor_type, qd_log_source_t  *log_source   );
+void qd_free_adaptor_config(qd_adaptor_config_t *config, qd_adaptor_type_t adaptor_type, qd_log_source_t  *log_source);
+
+/**
+ * Configure a connection's pn_tls objects
+ *     Info log describes objects being configured
+ * On success:
+ *     conn->tls_config and conn->tls_session are set up
```
kgiusti, line 98: Where does "conn" come from?  There is no "conn" argument.

src/adaptors/adaptor_common.h
```diff
@@ -0,0 +1,114 @@
...
+ * On failure:
+ *     Error log is written
+ *     All in-progress pn_tls objects are destroyed
+ *     conn->tls_config and conn->tls_session are freed and set to zero
```
kgiusti, line 102: So is *tls_config and *tls_session non-null when passed in?  The comment implies that what they are pointing to is released by this call.  Or are they zero'ed?

src/adaptors/adaptor_common.h
kgiusti: Originally the version enum was only part of the HTTP/* common stuff - it holds the value of the protocolVersion field in the httpConnector/httpListener schema entity (since these management entities are used for both HTTP1 and 2 - the version field is the only distinguishing element).

That said if we're introducing a new "qd_adaptor_type_t" enum then we could fold the version into it like tross suggests.

src/adaptors/adaptor_common.c
```diff
@@ -0,0 +1,270 @@
...
+            qd_log(log_source, QD_LOG_ERROR, "[C%"PRIu64"] Unable to create tls session for sslProfile %s with hostname: '%s'", conn_id, config->ssl_profile_name, config->host);
+            break;
+        }
+        pn_tls_set_peer_hostname(*tls_session, config->host);
+
+        tls_setup_success = true;
+
+        qd_log(log_source, QD_LOG_INFO, "[C%"PRIu64"] Successfully configured ssl profile %s", conn_id, config->ssl_profile_name);
+
+
```
kgiusti, line 153: You _could_ simply return(true) here, and do unconditional error cleanup after the while(0) statement. That avoids needing the tls_seup_success flag at all.  A matter of taste - your call.

src/adaptors/adaptor_common.c
```diff
@@ -0,0 +1,270 @@
...
+    return tls_setup_success;
+}
+
+void qd_free_adaptor_config(qd_adaptor_config_t *config, qd_adaptor_type_t adaptor_type, qd_log_source_t  *log_source)
+{
+    if (!config)
+        return;
+
+    if (adaptor_type == TCP) {
```
kgiusti, line 178: Why not put the adaptor_type into the qd_adaptor_config_t itself?  It would be useful to have it there, even if we only used it to assert() the correct type when accessing the proper union subfield.

src/adaptors/adaptor_common.c
```diff
@@ -0,0 +1,270 @@
...
+    free(config->host);
+    free(config->port);
+    free(config->site_id);
+    free(config->host_port);
+    free(config->ssl_profile_name);
+
+    if (adaptor_type == TCP) {
+        sys_atomic_destroy(&config->adaptor.tcp.ref_count);
+        sys_mutex_free(config->adaptor.tcp.stats_lock);
+        free_qd_adaptor_config_t(config);
```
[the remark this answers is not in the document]
kgiusti: Oh, I see now: for http the config is stack allocated, for tcp it is a sub-structure within the tcp connection structure which is allocated on the heap.

That's a no-go from me - that's just a recipe for disaster.  

Let's fix that: make the config a heap allocated object and use it consistently across all adaptors.  The load-adaptor function should be a constructor for the qd_adaptor_config_t object - it allocates it on the heap, sets is up, and returns a pointer to it.  Similarly the free function should decref the atomic counter and if zero free the config.

kgiusti (commented):
Overall I think this refactor is a good improvement to the code.

Other than a few nits (see inline) the only thing I really don't like about it is that it tries to allow the adaptors to either use a stack-based or heap based config. I think this patch needs to go further and remove the stack support and simply make the config a heap (pooled) object that is reference counted.

kgiusti (approved):

PR #468, opened 2022-05-13
Prefix for connectors generated by automesh script

nluaces:
When executing `skupper init` with the option `--routers <number>` the automesh script generates connectors based on the listener of one router. The config-sync sidecar needs to ignore this connectors when checking for connectors to delete, and for that reason a prefix is concatenated to the host name.


scripts/auto_mesh.py
```diff
@@ -301,15 +301,15 @@ def query():
         # Keep only pods created after current pod
         si = ip_list.index(ip)
         ip_list = ip_list[:si]
-        return [{"role": "inter-router", "host": host} for host in ip_list]
+        return [{"role": "inter-router", "host": "config-sync-ignore-%s" % host} for host in ip_list]
```
grs, line 304: I think we want to add a *name* with this pattern, but leave the host as it is (otherwise it won't resolve correctly).

jiridanek:
From the procedural side, please read about the skupper-router convention for commit messages, https://github.com/skupperproject/skupper-router/blob/main/CONTRIBUTING.adoc

Regarding substantial comments, I'll try to learn what the script actually does and how it fits into the bigger picture, and then I'll hopefully be qualified to say something about it.

scripts/auto_mesh.py
nluaces: That makes a lot of sense, I will fix that, thank you!

nluaces:
> From the procedural side, please read about the skupper-router convention for commit messages, https://github.com/skupperproject/skupper-router/blob/main/CONTRIBUTING.adoc

Apologies, I was going to do that when merging the PR, but indeed, I need to follow this convention with any commit. Thank you.

> Regarding substantial comments, I'll try to learn what the script actually does and how it fits into the bigger picture, and then I'll hopefully be qualified to say something about it.

I added you because I thought that you made some modifications at the script; just in case you could detect that something was wrong. You don't need to review this if you are not familiar with it; thanks for pointing out about the commit messages!

jiridanek:
> > From the procedural side, please read about the skupper-router convention for commit messages, https://github.com/skupperproject/skupper-router/blob/main/CONTRIBUTING.adoc
> 
> @jiridanek Apologies, I was going to do that when merging the PR, but indeed, I need to follow this convention with any commit. Thank you.

Doing this on merge is fine, the only problem with that is to not forget ;P Usually (at least when @ganeshmurthy is the judge) it is enough to put the proper commit message format into the PR title lane. That way, when you rebase/squash-rebase in GitHub UI, and there is more than one commit in PR, you get this prefilled in the form...

Not that the commit format matters much, at least for now. There is no automated (or manual) processing that relies on it. It helps to find additional info when trying to understand code history, sometimes. Like, what were the symptoms of the issue that this commit is solving.
 
> > Regarding substantial comments, I'll try to learn what the script actually does and how it fits into the bigger picture, and then I'll hopefully be qualified to say something about it.
> 
> I added you because I thought that you made some modifications at the script; just in case you could detect that something was wrong. You don't need to review this if you are not familiar with it; thanks for pointing out about the commit messages!

I was fixing python-checker ;P adding extra blank lines requires no understanding of the function of the script ;p

Btw, why doesn't GitHub show link to https://github.com/skupperproject/skupper/pull/764 here, when it was added from the other side?

jiridanek (commented):
So this queries Kubernetes, (or figures out from its own pod hostname), what other skupper-router pods were created before this one and and adds connectors to connect to them, creating a full mesh?

Looks ok to me, but I'd prefer if @fgiorgetti or @ganeshmurthy had a look at this too.

scripts/auto_mesh.py
```diff
@@ -301,15 +303,23 @@ def query():
...
     (prefix, index) = os.environ["HOSTNAME"].rsplit("-", 1)
-    return [{"role": "inter-router", "host": "%s-%s-%s.%s.%s.svc.cluster.local" % ("config-sync-ignore-", prefix, i, service_name, namespace)} for i in range(int(index))]
+    return [{"role": "inter-router", "host": "%s-%s.%s.%s.svc.cluster.local" % (prefix, i, service_name, namespace)} for
+            i in range(int(index))]
+
+
+def generate_connector_name(connector):
+    host = connector.get("host")
+    port = connector.get("port")
+    name = "config-sync-ignore-connector/{}:{}".format(host, port)
```
grs, line 321: I'd suggest auto-mesh-connector or similar for the name as it is more directly tied to the purpose of the connector.

nluaces: I agree, done!

kgiusti (approved):

PR #470, opened 2022-05-14
Increase coverage of `connection_manager.c`

jiridanek:


kgiusti (changes requested):
Seems like something is broken:

https://github.com/skupperproject/skupper-router/runs/6433878540?check_suite_focus=true#step:9:240

kgiusti (approved):

PR #481, opened 2022-05-18
Fixes #452: fix python linter errors

### LINKED ISSUE #452: Remove out-of-place annotations from `dispatch.pyi`

kgiusti:
The previous commit to fix #452 introduced linter errors on Fedora 34.  This patch fixes those errors.

### DIFF
--- a/python/skupper_router_internal/dispatch.py
+++ b/python/skupper_router_internal/dispatch.py
@@ -40,7 +40,7 @@ import sys
 from ctypes import c_char_p, c_long, py_object, c_void_p, c_bool, PyDLL
 from types import ModuleType
 
-from typing import List, Callable, TYPE_CHECKING, Optional
+from typing import List, Callable, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from .router.message import Message
@@ -76,8 +76,7 @@ class IoAdapter:
     def __init__(self, handler: Callable, address: str, aclass: str, treatment: int) -> None:
         ...
 
-    def send(self, message: 'Message', no_echo: Optional[bool] = True, control:
-             Optional[bool] = False) -> None:
+    def send(self, message: 'Message', no_echo: bool = True, control: bool = False) -> None:
         ...
 
 
--- a/python/skupper_router_internal/dispatch.py
+++ b/python/skupper_router_internal/dispatch.py
@@ -35,14 +35,12 @@ For Python unit-testing, this module is replaced by tests/mock/dispatch.py to br
 the dependency on the C.
 """
 import builtins
-import ctypes
 import os
 import sys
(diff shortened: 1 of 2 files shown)


jiridanek (commented):
The tsan failure (timeout) seems new, opening

* https://github.com/skupperproject/skupper-router/issues/483

python/skupper_router_internal/dispatch.py
```diff
@@ -78,7 +76,7 @@ class IoAdapter:
     def __init__(self, handler: Callable, address: str, aclass: str, treatment: int) -> None:
         ...
 
-    def send(self, message: 'Message', no_echo: int, control: int) -> None:
+    def send(self, message: 'Message', no_echo: bool = True, control: bool = False) -> None:
```
ganeshmurthy, line 79: What is the difference between `no_echo: bool = True` and just `no_echo=True` ? Is the latter old school or does the latter not fix the linter error ?

kgiusti: I didn't want to change the original meaning of the arguments - doing so would involve finding all the callers and changing them to adopt the new meaning. Beyond the scope of this fix.

kgiusti: The bool designation is used by the linters as a hint to check the callers to ensure they are invoking the method with the proper types.

'typing' (type hints) is a new(ish) feature in python and helps static analysis - we really should use it going forward.  See https://docs.python.org/3/library/typing.html?highlight=typing#module-typing

jiridanek: @kgiusti It would help if you showed what linter warnings you are actually fixing here

ganeshmurthy (approved):

python/skupper_router_internal/dispatch.py
kgiusti: @jiridanek ah, yes - sorry about that:



    67: ************* Module skupper_router_internal.dispatch
    67: /home/kgiusti/work/skupper/skupper-router/python/skupper_router_internal/dispatch.py:44:0: W0404: Reimport 'ctypes' (imported line 38) (reimported)
    67: /home/kgiusti/work/skupper/skupper-router/python/skupper_router_internal/dispatch.py:44:0: C0412: Imports from package ctypes are not grouped (ungrouped-imports)
    67: ************* Module skupper_router_internal.management.agent
    67: /home/kgiusti/work/skupper/skupper-router/python/skupper_router_internal/management/agent.py:863:12: E1120: No value for argument 'no_echo' in method call (no-value-for-parameter)
    67: /home/kgiusti/work/skupper/skupper-router/python/skupper_router_internal/management/agent.py:863:12: E1120: No value for argument 'control' in method call (no-value-for-parameter)

PR #490, opened 2022-05-19
Fixes #489: fix bug where CI tests ignored process exit error status

### LINKED ISSUE #489: CI is ignoring error codes returned from qdrouterd, allowing failed tests to pass

kgiusti:


tests/system_test.py
```diff
@@ -360,51 +359,35 @@ def error(msg):
...
-                # Wait for TIMEOUT seconds and then catch the
-                # TimeoutExpired and kill() the process.
-                self.wait(timeout=TIMEOUT)
+                self.wait(TIMEOUT)
             except TimeoutExpired:
-                # The terminate call() has failed to terminate the process even after we waited for TIMEOUT seconds.
-                # Send a SIGKILL to the child by calling kill()
-                self._logger.log("Wait timeout expired trying to terminate "
-                                 "process %s (name=%s). Trying to kill process now" % (self.pid, self.name))
                 self.kill()
```
jiridanek, line 369: Not within scope of PR, but what about sending it SIGABORT instead of SIGKILL? That way, core is dumped and backtrace can be obtained to investigate where process got stuck.

ganeshmurthy:
The dichotomy here is the status before and after calling terminate. My earlier commit failed to take into account the return code from terminate(). The best fix like you suggested would be https://github.com/skupperproject/skupper-router/pull/377#issuecomment-1111067495 but short of that this is the correct approach,

ganeshmurthy (approved):

PR #496, opened 2022-05-25
[Rust] Implement and document capability to run `skrouterd` in system-tests under `rr` debugger

jiridanek:
This requires a sigforwarder utility so that the tests are able to SIGTERM the skrouterd cleanly.

### DIFF
--- a/scripts/sigforwarder.rs
+++ b/scripts/sigforwarder.rs
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


jiridanek:
> Is this related to #552 ? Do we need both?

@kgiusti One of these is sufficient. @astitcher seemed to be unwilling to approve the rust version, so I added a PR with Python version of the sigforwarder, hoping that he might like it better

kgiusti:
> > Is this related to #552 ? Do we need both?
> 
> @kgiusti One of these is sufficient. @astitcher seemed to be unwilling to approve the rust version, so I added a PR with Python version of the sigforwarder, hoping that he might like it better

@jiridanek thanks for the explanation.

All things being equal, I think we should go with the Python impl simply to avoid adding a whole new language dependency into this project.  Keeps things consistent with the rest of the project and developer expectations.

jiridanek:
> All things being equal, I think we should go with the Python impl simply to avoid adding a whole new language dependency into this project. Keeps things consistent with the rest of the project and developer expectations.

Closing the Rust one, then

Issue #533, opened 2022-06-01
[2.0.x] Crash in system_tests_address_watch


jiridanek:
https://github.com/skupperproject/skupper-router/runs/6684792971?#step:27:48

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007ff5b400ac95 in __strlen_avx2 () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7ff5aa307700 (LWP 2736))]

Thread 6 (Thread 0x7ff5ad34a700 (LWP 2732)):
#0  0x00007ff5b510e81d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007ff5b5107ac9 in pthread_mutex_lock () from /lib64/libpthread.so.0
#2  0x000000000045d639 in sys_mutex_lock (mutex=<optimized out>) at /home/runner/work/skupper-router/skupper-router/skupper-router/src/posix/threading.c:57
#3  0x000000000044f60e in qd_vlog_impl (source=0xd1bb20, level=QD_LOG_DEBUG, check_level=check_level@entry=true, file=file@entry=0x4ae578 "/home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/delivery.c", line=line@entry=641, fmt=fmt@entry=0x4ae810 "[C%lu][L%lu][D%u] Delivery decref_CT: rc:%u %s", ap=0x7ff5ad349168) at /home/runner/work/skupper-router/skupper-router/skupper-router/src/log.c:437
#4  0x000000000044f8bb in qd_log_impl (source=<optimized out>, level=level@entry=QD_LOG_DEBUG, file=file@entry=0x4ae578 "/home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/delivery.c", line=line@entry=641, fmt=fmt@entry=0x4ae810 "[C%lu][L%lu][D%u] Delivery decref_CT: rc:%u %s") at /home/runner/work/skupper-router/skupper-router/skupper-router/src/log.c:463
#5  0x000000000047031b in qdr_delivery_decref_CT (core=0xe79180, dlv=dlv@entry=0x13[48](https://github.com/skupperproject/skupper-router/runs/6684792971?check_suite_focus=true#step:27:49)f50, label=label@entry=0x4adcf0 "qdrc_endpoint_settle_CT - no longer held by endpoint") at /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/delivery.c:640
[...]

Issue #520, opened 2022-05-27
python_embedded.c:665: heap-use-after-free in `qd_io_rx_handler`


jiridanek:
https://github.com/jiridanek/skupper-router/runs/6627949055?check_suite_focus=true#step:26:2070

```
E           RuntimeError: Process 2786 (name=INT.A) error: returned error code 1
E           skrouterd -c INT.A.conf -I /home/runner/work/skupper-router/skupper-router/skupper-router/python
E           /__w/skupper-router/skupper-router/skupper-router/build/tests/system_test.dir/tests/system_tests_topology/RouterFluxTest/test_02_shutdown_INT_A/INT.A-9.cmd
E           >>>>
E           =================================================================
E           ==2786==ERROR: AddressSanitizer: heap-use-after-free on address 0x60800059e750 at pc 0x00000056a0ed bp 0x7ffcd8193d60 sp 0x7ffcd8193d58
E           READ of size 8 at 0x60800059e750 thread T0
E               #0 0x56a0ec in qd_io_rx_handler /home/runner/work/skupper-router/skupper-router/skupper-router/src/python_embedded.c:665
E               #1 0x56a0ec in qd_io_rx_handler /home/runner/work/skupper-router/skupper-router/skupper-router/src/python_embedded.c:636
E               #2 0x5e71ec in qdr_forward_on_message /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/forwarder.c:324
E               #3 0x5fde13 in qdr_general_handler /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/router_core.c:824
E               #4 0x6114fb in qdr_core_free /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/router_core.c:252
E               #5 0x66e9a5 in qd_router_free /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_node.c:2096
E               #6 0x4fb3c2 in qd_dispatch_free /home/runner/work/skupper-router/skupper-router/skupper-router/src/dispatch.c:361
E               #7 0x4346b4 in main_process /home/runner/work/skupper-router/skupper-router/skupper-router/router/src/main.c:113
E               #8 0x433552 in main /home/runner/work/skupper-router/skupper-router/skupper-router/router/src/main.c:363
E               #9 0x7f324039343f in __libc_start_call_main (/lib64/libc.so.6+0x4043f)
[...]

PR #544, opened 2022-06-06
Fixes #542 - Added a callback to confirm the cancellation of an address-watch.

### LINKED ISSUE #542: Add a cleanup callback to the address-watch API
Currently, when canceling an address-watch, there is a window of time after the cancel when an address update callback can be invoked.  This requires the user of the API to go to lengths to validate the context of callbacks.  By adding a "cleanup" callback that confirms the completion of the cancellation, this effort can be avoided.  Once the cleanup callback is invoked, it is guaranteed that no further address update events shall be generated.

ted-ross:
This is an improvement (I think) over the previous fix for #533.

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -199,6 +199,16 @@ typedef void (*qdr_address_watch_update_t)(void     *context,
                                            uint32_t  remote_consumers,
                                            uint32_t  local_producers);
 
+
+/**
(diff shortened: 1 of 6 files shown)


kgiusti (approved):

kgiusti:
The test failure is a known issue (separate from this PR) 
https://github.com/skupperproject/skupper-router/issues/520

PR #550, opened 2022-06-08
Fixes #120: update tcp listener to use new watched address callbacks

### LINKED ISSUE #120: TCP adaptor listener connections hang if no connectors are defined

kgiusti:


ganeshmurthy (approved):
This is much easier to understand

PR #564, opened 2022-06-10
Fixes #799: Split the pooled type declarations into weak-reffed and the non-weak-reffed

### LINKED ISSUE #799: Split the pooled type declarations into weak-reffed and the non-weak-reffed
There are three reasons why use memory pool for a type. First, because it makes allocation/deallocation faster, second because it prevents memory fragmentation, and third it allows holding weak references to the type.

The pool is constrained by having to support these use cases. Namely, a pool that supports weak references may never free any memory to the operating system, because it must be always able to dereference any previously obtained weak pointer.

This PR creates a separate API to declare weak-reffed types, so it is now obvious what types need weak references. There are no changes to functionality.

Note the functions mentioned in the second commit, which bypass the usual weak pointer API and mess with weak pointers under the table. This approach makes it look like the types don't get used with weak pointers, but they actually do get weak-pointed. (This only becomes visible when you (for testing purposes) replace pooled allocation for the non-weak types with malloc/free; the way the PR does things, the underhanded approach can still be used to create weak references even though type was not declared to support it.)

jiridanek:
There are three reasons why use memory pool for a type. First, because it makes allocation/deallocation faster, second because it prevents memory fragmentation, and third it allows holding weak references to the type.

The pool is constrained by having to support these use cases. Namely, a pool that supports weak references may never free any memory to the operating system, because it must be always able to dereference any previously obtained weak pointer.

This PR creates a separate API to declare weak-reffed types, so it is now obvious what types need weak references. There are no changes to functionality.

Note the functions mentioned in the second commit, which bypass the usual weak pointer API and mess with weak pointers under the table. This approach makes it look like the types don't get used with weak pointers, but they actually do get weak-pointed. (This only becomes visible when you (for testing purposes) replace pooled allocation for the non-weak types with malloc/free; the way the PR does things, the underhanded approach can still be used to create weak references even though type was not declared to support it.)

### DIFF
--- a/include/qpid/dispatch/alloc_pool.h
+++ b/include/qpid/dispatch/alloc_pool.h
@@ -108,14 +108,14 @@ static inline void *qd_alloc_deref_safe_ptr(const qd_alloc_safe_ptr_t *sp)
 #define ALLOC_DECLARE(T) \
     extern __thread qd_alloc_pool_t *__local_pool_##T; \
     T *new_##T(void);    \
-    void free_##T(T *p)
+    void free_##T(T *p); \
+    qd_alloc_stats_t *alloc_stats_##T(void)
 
 #define ALLOC_DECLARE_SAFE(T) \
     ALLOC_DECLARE(T); \
     typedef qd_alloc_safe_ptr_t T##_sp; \
     void set_safe_ptr_##T(T *p, T##_sp *sp); \
-    T *safe_deref_##T(T##_sp sp); \
-    qd_alloc_stats_t *alloc_stats_##T(void)
+    T *safe_deref_##T(T##_sp sp)
 
 /**
  * Define allocator configuration.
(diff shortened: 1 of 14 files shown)


include/qpid/dispatch/alloc_pool.h
```diff
@@ -106,6 +106,11 @@ static inline void *qd_alloc_deref_safe_ptr(const qd_alloc_safe_ptr_t *sp)
  * Declare functions new_T and alloc_T
  */
 #define ALLOC_DECLARE(T) \
+    extern __thread qd_alloc_pool_t *__local_pool_##T; \
+    T *new_##T(void);    \
+    void free_##T(T *p)
+
+#define ALLOC_DECLARE_WEAK(T) \
```
jiridanek, line 113: @ganeshmurthy wants to call this _SAFE

jiridanek:
... one more command is needed to process the generated fixup! commits

```
git rebase -i --autosquash origin/main
```

jiridanek:
AFAIK we do want to do this,

image

ganeshmurthy:
@jiridanek Looks like your new rebase fixed the issue and all tests have passed. Please feel free to merge.

ganeshmurthy (approved):

PR #566, opened 2022-06-14
Fixes #565: properly free mutex in tcp connector on shutdown

### LINKED ISSUE #565: Leak of mutex associated with TCP Connector instances (shutdown leak)

kgiusti:


jiridanek (approved):
The `pn_` leaks should go away after Proton 0.38 is the new minimum. So the mutex leaks are now the main entry in `lsan.supp`. Too bad that the other mutex leaks are related to the leaks of the pooled objects, so the fixing gets tougher.

ganeshmurthy (approved):

PR #575, opened 2022-06-24
ISSUE-574: fix infinte loop on tcpListener delete

kgiusti:


fgiorgetti (approved):
LGTM. Both the skupper integration test and my reproducer passed just fine using your branch.

PR #595, opened 2022-07-08
Fixes #594: use local environment settings for HTTP/2 test server

### LINKED ISSUE #594: HTTP/2 test server environment settings leak between tests
The HTTP/2 test servers are configured via environment variables. Currently this is done by modifying the environment of the test process, then spawning the test server. This approach leaves the environment modified when the next test is run. Stale test configuration can contaminate following tests.

kgiusti:
Rather than modifying the testing process environment, pass the
environment configuration into the HTTP/2 test server subprocess.

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -801,6 +801,27 @@ class Qdrouterd(Process):
             assert retry(lambda c=c: self.is_connected(port=c['port'], host=c.get('host') if c.get('host') else self.get_host(c.get('socketAddressFamily'))),
                          **retry_kwargs), "Port not connected %s" % c['port']
 
+    def wait_log_message(self, pattern, logfile_path=None, **retry_kwargs):
+        """Wait for a log message matching the pattern to appear in the routers
+        log file. The log file for the DEFAULT log module is used unless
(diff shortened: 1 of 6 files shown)


tests/system_test.py
```diff
@@ -430,7 +431,11 @@ def __init__(self, name=None, listen_port=None, wait=True,
         self.args = [sys.executable, os.path.join(os.path.dirname(os.path.abspath(__file__)), self.server_file)]
         if self.cl_args:
             self.args += self.cl_args
-        super(Http2Server, self).__init__(self.args, name=name, expect=expect, **kwargs)
+        server_env = dict(os.environ)
```
ganeshmurthy, line 434: Was wondering if this line is necessary for the tests to pass 
```
server_env = dict(os.environ)
```

kgiusti: When passing an environment to a Process, the sub-process's entire environment is replaced.  Since we still need things like PATH we need to pass the parent's environment down. This patch creates a new environment based on the parent's env + the extra config vars. By doing a copy of the env we leave the parent's env unchanged.

Does that answer your question?

ganeshmurthy (approved):

PR #611, opened 2022-07-15
Fixes #610 - Remove all static state that is mutable and not explicit…

### LINKED ISSUE #610: VFlow - Remove mutable static state that is not explicitly initializied

ted-ross:
…ly initialized

### DIFF
--- a/src/vanflow.c
+++ b/src/vanflow.c
@@ -120,42 +120,46 @@ DEQ_DECLARE(vflow_work_t, vflow_work_list_t);
 
 static const char *event_address_all           = "mc/sfe.all";
 static const char *event_address_my_prefix     = "mc/sfe.";
-static       char *event_address_my            = 0;
 static const char *command_address_prefix      = "sfe.";
-static       char *command_address             = 0;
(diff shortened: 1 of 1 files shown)


src/vanflow.c
```diff
@@ -1532,23 +1539,28 @@ static void _vflow_final(void *adaptor_context)
...
-    sys_mutex_free(lock);
-    sys_mutex_free(id_lock);
+    sys_cond_free(state->condition);
+    sys_mutex_free(state->lock);
+    sys_mutex_free(state->id_lock);
+
+    //
+    // Free the module state
+    //
+    free(state);
```
kgiusti, line 1562: for extra re-entrancy Karma, set state = 0; after the free.

kgiusti (approved):

src/vanflow.c
jiridanek: can't think of a situation where that would be all that helpful; regarding karmic effects I hesitate to speculate

anyways, whatever this is, it's not re-entrancy, even with this you cannot run two routers from the same process _at the same time_ ;P

PR #613, opened 2022-07-15
Issue #609 - Reset the .data and .bss sections for every unittest

jiridanek:
note to myself: commit [fixup nobody was initializing event_lock before](https://github.com/skupperproject/skupper-router/commit/f0a835d37a05796a5fefe8a04fa7da290d348d1c) contains additional fix

PR #614, opened 2022-07-18
Flush collected test coverage data when resetting static variables

jiridanek:
This fixes mistake in https://github.com/skupperproject/skupper-router/pull/613 which caused that test coverage was not being correctly counted.


kgiusti (approved):

jiridanek:
There is _no reason_ why codecov should not comment on this PR. But comment is not forthcoming. I am concerned.

jiridanek:
> There is _no reason_ why codecov should not comment on this PR. But comment is not forthcoming. I am concerned.

I got that comment in a repost under https://github.com/skupperproject/skupper-router/pull/617. Maybe the comment bot was broken yesterday. Or maybe the slash in branch name stymied it the last time (the reposted PR is from a branch without `/`).

Issue #126, opened 2022-03-04
AttributeError: 'NoneType' object has no attribute 'properties' in OversizeMulticastTransferTest


jiridanek:
```
23: DEBUG    root:system_test.py:1390 2022-03-04 18:41:21.956829 OversizeMulticastTransferTest - multicast/e81: Container run exception: ['Traceback (most recent call last):
  File "/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_policy_oversize_compound.py", line 631, in run
    Container(self).run()
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_reactor.py", line 194, in run
    while self.process():
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_reactor.py", line 257, in process
    event.dispatch(handler)
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_events.py", line 164, in dispatch
    self.dispatch(h, type)
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_events.py", line 161, in dispatch
    _dispatch(handler, type.method, self)
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_events.py", line 128, in _dispatch
    m(*args)
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/proton/_handlers.py", line 753, in on_reactor_init
    self.on_start(event)
  File "/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_policy_oversize_compound.py", line 423, in on_start
    self.test_class.wait_router_network_connected()
  File "/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_tests_policy_oversize_compound.py", line 740, in wait_router_network_connected
    cls.INT_A.wait_router_connected(\'INT.B\')
  File "/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py", line 745, in wait_router_connected
    retry(lambda: self.is_router_connected(router_id), **retry_kwargs)
  File "/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py", line 140, in retry
    ret = function()
[...]

PR #619, opened 2022-07-19
ISSUE-618: recreate the client each time is_router_connected fails

kgiusti:


tests/system_test.py
```diff
@@ -846,19 +846,23 @@ def wait_ready(self, **retry_kwargs):
...
+
+        # TODO aconway 2015-01-29: The above check should be enough, we
+        # should not advertise a remote router in management till it is fully
+        # connected. However we still get a race where the router is not
+        # actually ready for traffic. Investigate.
+        # Meantime the following actually tests send-thru to the router.
+        try:
+            with Node.connect(self.addresses[0], router_id) as node:
+                return node.query('io.skupper.router.router')
+        except Exception as exc:
```
jiridanek, line 863: I don't like this broad except. It catches syntax errors and so on. That makes it hard to revisit such test code in the future.

jiridanek: I'd put back the line

```
except (proton.ConnectionException, NotFoundStatus, proton.utils.LinkDetached) as exc:
```

and probably the associated comment?

jiridanek (approved):
Looks ok, restoring the aconway comment is good, but I am not fan of that `except Exception` block.

jiridanek:
This might maybe help with https://github.com/skupperproject/skupper-router/issues/126, I don't remember the details there, but recreating the connection for every check should avoid the possibility of (one of the) failures in that ticket.

tests/system_test.py
kgiusti: Ok, I've restored it.

jiridanek:
Thanks a lot for adding the specific exceptions. I hope it will be good for something in the future. More reliable tests!?

Issue #622, opened 2022-07-20
Direct leak of 5376 byte(s) in 70 object(s) allocated from: 0x7f36108ba68f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)


jiridanek:
```
67: =================================================================
67: ==109869==ERROR: LeakSanitizer: detected memory leaks
67:
67: Direct leak of 5376 byte(s) in 70 object(s) allocated from:
67:     #0 0x7f36108ba68f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)
67:     #1 0x7f3610381b6f in _PyMem_DebugRawAlloc (/lib64/libpython3.10.so.1.0+0x181b6f)
67:
67: -----------------------------------------------------
67: Suppressions used:
67:   count      bytes template
67:      14        896 sys_mutex
67: -----------------------------------------------------
67:
67: SUMMARY: AddressSanitizer: 5376 byte(s) leaked in 70 allocation(s).
1/1 Test #67: c_unittests ......................***Failed    1.77 sec
```

ganeshmurthy:
Did this happen in CI ? I am guessing this failure did not happen in CI since you would have posted a link to the failing CI. 

Was this from @ted-ross ? If yes, he is on Fedora 36. Thanks

jiridanek:
The best I can do is to preserve some Python global variables when reset is being done (the breaking commit above is for the globals reset)

```
        _typeobject  my_PyBool_Type        = PyBool_Type;
        _object    *my_PyExc_Exception    = PyExc_Exception;
        _object    *my_PyExc_RuntimeError = PyExc_RuntimeError;
        _object    *my_PyExc_TypeError    = PyExc_TypeError;
        _longobject  my__Py_FalseStruct    = _Py_FalseStruct;
        _typeobject my_PyFloat_Type       = PyFloat_Type;
        _object     my__Py_NoneStruct     = _Py_NoneStruct;
        _longobject  my__Py_TrueStruct     = _Py_TrueStruct;

[...]

            PyBool_Type = my_PyBool_Type;
            PyExc_Exception = my_PyExc_Exception;
            PyExc_RuntimeError = my_PyExc_RuntimeError;
            PyExc_TypeError = my_PyExc_TypeError;
            _Py_FalseStruct = my__Py_FalseStruct;
            PyFloat_Type = my_PyFloat_Type;
            _Py_NoneStruct = my__Py_NoneStruct;
[...]

PR #624, opened 2022-07-20
Issue #622 - Set `-DSANITIZE_PYTHON` to either `OFF` or `ON`, never leave it empty

jiridanek:


jiridanek:
This failed in the correct way in https://github.com/skupperproject/skupper-router/runs/7431173143?check_suite_focus=true#step:28:1369

when I rebase now, it should all succeed

ganeshmurthy (approved):

PR #625, opened 2022-07-20
Give threads nice names so that they can have nice names

jiridanek:
I'm not sure what are all the places this shows itself, but I am hoping this is useful for _something_.

There is a limit in pthreads saying that thread names may be at most 16 bytes (including the final '\0' byte). That poses severe limits on any expansive and poetic facility with language that might be present among those coming up with the names.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1486,9 +1486,9 @@ void qd_server_run(qd_dispatch_t *qd)
     const int n = qd_server->thread_count;
     sys_thread_t **threads = (sys_thread_t **)qd_calloc(n, sizeof(sys_thread_t*));
     for (i = 0; i < n; i++) {
-        char buffer[16];
-        snprintf(buffer, sizeof(buffer), "wrkr_%d", i);
-        threads[i] = sys_thread(buffer, thread_run, qd_server);
+        char thread_name[16];
+        snprintf(thread_name, sizeof(thread_name), "wrkr_%d", i);
+        threads[i] = sys_thread(thread_name, thread_run, qd_server);
     }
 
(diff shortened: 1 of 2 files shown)


src/posix/threading.c
```diff
@@ -173,12 +176,15 @@ static void *_thread_init(void *arg)
...
-sys_thread_t *sys_thread(void *(*run_function) (void *), void *arg)
+sys_thread_t *sys_thread(const char * thread_name, void *(*run_function) (void *), void *arg)
 {
+    // thread name may be at most 16 bytes, counting the final \0 byte
+    assert(strlen(thread_name) < 16);
     sys_thread_t *thread = NEW(sys_thread_t);
     thread->f = run_function;
     thread->arg = arg;
     pthread_create(&(thread->thread), 0, _thread_init, (void*) thread);
+    pthread_setname_np(thread->thread, thread_name);
```
jiridanek, line 187: return code from these should be checked

kgiusti: Not a bad idea to check the return codes, but I wouldn't hold up this patch for it.

Usually for those functions that should always work unless we pass garbage to them then doing a check only in the debug build case would catch any misuse during development.  You'll see code that does this alot:

int result = foo(x);
(void)result; assert(result == 0);

kgiusti (approved):

ganeshmurthy:
This is very useful especially when looking at threads thru gdb.

jiridanek:
It shows in tsan traces as well, https://github.com/skupperproject/skupper-router/runs/7457440719?check_suite_focus=true#step:28:2340

PR #629, opened 2022-07-21
Tross edge in process

### LINKED ISSUE #626: In-process destination is unreachable on an edge router

ted-ross:
Fixes #626 - Allow in-process destination address to be reachable from across the network on both edges and interiors.

### DIFF
--- a/src/router_core/modules/edge_router/addr_proxy.c
+++ b/src/router_core/modules/edge_router/addr_proxy.c
@@ -254,7 +254,7 @@ static void on_conn_event(void *context, qdrc_event_t event, qdr_connection_t *c
                 // If the address has more than zero attached destinations, create an
                 // incoming link from the interior to signal the presence of local consumers.
                 //
-                if (DEQ_SIZE(addr->rlinks) > 0) {
+                if (DEQ_SIZE(addr->rlinks) > 0 || (DEQ_SIZE(addr->subscriptions) > 0 && addr->propagate_local)) {
                     if (DEQ_SIZE(addr->rlinks) == 1) {
                         //
                         // If there's only one link and it's on the edge connection, ignore the address.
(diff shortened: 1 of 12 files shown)


src/router_core/modules/edge_router/addr_proxy.c
```diff
@@ -254,7 +254,7 @@ static void on_conn_event(void *context, qdrc_event_t event, qdr_connection_t *c
                 // If the address has more than zero attached destinations, create an
                 // incoming link from the interior to signal the presence of local consumers.
                 //
-                if (DEQ_SIZE(addr->rlinks) > 0) {
+                if (DEQ_SIZE(addr->rlinks) > 0 || (DEQ_SIZE(addr->subscriptions) > 0 && addr->propagate_local)) {
                     if (DEQ_SIZE(addr->rlinks) == 1) {
                         //
                         // If there's only one link and it's on the edge connection, ignore the address.
```
kgiusti, line 260: Q: should an inlink be added unconditionally if there's a propagate subscription? IOW, does the addr->rlinks == 1... precondition apply in that case?

src/router_core/modules/edge_router/addr_proxy.c
```diff
@@ -329,7 +329,7 @@ static void on_addr_event(void *context, qdrc_event_t event, qdr_address_t *addr
         // not the link to the interior.
         //
         link_ref = DEQ_HEAD(addr->rlinks);
-        if (link_ref->link->conn != ap->edge_conn)
+        if (link_ref->link->conn != ap->edge_conn || addr->propagate_local)
```
kgiusti, line 332: See previous comment, 'cause IIUC there's a conflict here.

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -336,7 +336,7 @@ static qd_message_t *qcm_mobile_sync_compose_absolute_mau(qdrm_mobile_sync_t *ms
         //
         if (qcm_mobile_sync_addr_is_mobile(addr)
             && (DEQ_SIZE(addr->rlinks) > 0
-                || DEQ_SIZE(addr->subscriptions) > 0
+                || (DEQ_SIZE(addr->subscriptions) > 0 && addr->propagate_local)
```
kgiusti, line 339: Does this change the visibility of $management?

ted-ross: It does not.  $management is a subscription on every router and the forwarding treatment is "CLOSEST".  All traffic to $management is always directed to the agent on the ingress router.

Issue #252, opened 2022-03-28
[intel vtune] Consider pooling or sharding mutexes for some types; take proactor guarantees into consideration


jiridanek:
Intel V-Tune dislikes the mutexes in the router code

![sys_mutex](https://user-images.githubusercontent.com/442720/160381053-5a4d00b2-0162-449a-a8d0-d295687091ab.png)

I am not sure what the profiler is actually showing here, and how can I get same result (these functions shown at the top) from linux `perf`.

Regarding solutions. I think proactor guarantees do not help here. Proactor ensures the program never processes the same connection (link?) in more than one thread. Messages are by the nature of the router's function associated with more than one connection.

@franz1981 once considered pooling the mutexes. That is, if each instance of `qd_message_t`, etc. needs a mutex, then create a pool of initialized mutexes, and upon message creation, take a mutex out for it; on destruction of message, leave mutex initialized and put it back to pool. https://github.com/apache/qpid-dispatch/pull/508#issuecomment-505461704

There is alternative design, where you create an array of mutexes, choosing a suitable size N for it, probably based on number of worker threads, and some additional empiric parameters or constants. Then, you compute a hash of qd_message_t mod N. This is the index to the mutex in the array that will be protecting this message. Multiple unrelated qd_message_t's may now share a mutex, meaning that in some cases, two threads which were formerly free to work concurrently now cannot. Some sort of probing, as in hash tables, can be introduced on top, to minimize such collisions. Overall, this has a chance to produce speedup, however.

Finally, one possible speedup could be if mutex field in the structs was not a pointer but mutex was directly contained in the structs. When I tried this, I did not see the miraculous speedups I hoped for. Probably not worth pursuing it.

kgiusti:
From my flamegraph analysis the real overhead with sys_mutex() is the memory allocation overhead for the sys_mutex_t wrapper object (look a qd_message() in the flamegraph below):

![qdr_perf_worker_117881_117881](https://user-images.githubusercontent.com/164694/162004176-26c7cfd9-1525-4fb3-9dee-7f75790ca115.svg)

https://github.com/skupperproject/skupper-router/blob/9a20e6e04c7f4c8a1fa6012e84075f2f732af88b/src/posix/threading.c#L41

The actual setup of the pthread mutex is very cheap.  So a a very first step I'd recommend you're suggestion of simply in-lining the mutex directly and avoiding the memory alloc completely.

Once you have a pull request for that re-run your vtune analysis and see what happens.

[...]

PR #631, opened 2022-07-22
Fixes Issue #630: inline the mutex/condition implementation

kgiusti:


jiridanek (commented):
I guess there is little for me to say about this. The change is on one hand rather straightforward, on the other hand quite tedious to make across all the files... It seems to be working as intended.

This is related to https://github.com/skupperproject/skupper-router/issues/252. The futex experiment there probably shows that gains from any mutex sharding or pooling would be marginal, and this PR seems to already be giving ~1/2 of the speedup that the futex did. (I am assuming that futexes are so lightweight that using them essentially simulates using a very efficient lock pool).

lgtm

PR #642, opened 2022-07-28
Fixes #638 - Allow multiple addresses watches on an address.

### LINKED ISSUE #638: One of two tcpListeners are not activated on a router that has a tcpConnector
I have a router config which has two tcpListeners and one tcpConnector, all on the same address like below

```
tcpListener {
    host: localhost
    port: 20829
    address: ES_INTA
    siteId: mySite
}
tcpListener {
    host: localhost
    port: 20730
    address: ES_INTA
    siteId: mySite
}

tcpConnector {
    host: localhost
    port: 42621
    address: ES_INTA
    sslProfile: tcp-connector-ssl-profile
    siteId: mySite
}
```

When the router starts up, it listens only on one of the tcpListeners (in this case on port 20829). The other tcpListener is not activated. In other words, the router does not listen on port 20730 and anyone trying to connect on port 20730 gets their connection refused by the router.

ted-ross:


ganeshmurthy (approved):
Looks good to me. Please merge this to main branch and I will test it using my test case.

PR #649, opened 2022-08-02
Fixes #602: user configurable listener backlog

### LINKED ISSUE #602: Feature: user configurable listener backlog
The TCP and HTTP adaptors hardcode the server socket's backlog to 50.

mgoulish:
Make a new attribute 'backlog' for entity tcpListener.
User can enter integer >= 1, or "max".
Default is "max" -- which means it uses SOMAXCONN.

Invalid value logs error, the uses old value of 50.

I tested setting it to 'max' ,  small integers, and 'foo'.
And then used my connection rate test on it.
Difference in connection rate is very substantial using 'max' vs 5 or 10.

### DIFF
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -1184,6 +1184,12 @@
                     "default": "0.0.0.0",
                     "create": true
                 },
+                "backlog": {
+                    "description": "Sets the value of the backlog parameter in call to listen(). This controls the maximum queue depth the OS will use for some stages of TCP connection creation.  Value should be integer >= 1, or 'max'. If 'max', the value will be set to the upper limit allowed by the OS.'",
+                    "type": "string",
+                    "default": "max",
+                    "create": true
+                },
                 "port": {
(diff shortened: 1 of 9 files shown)


python/skupper_router/management/skrouter.json
```diff
@@ -1184,6 +1184,12 @@
                     "default": "0.0.0.0",
                     "create": true
                 },
+                "backlog": {
+                    "description": "Sets the value of the backlog parameter in call to listen(). This controls the maximum queue depth the OS will use for some stages of TCP connection creation.  Value should be integer >= 1, or 'max'. If 'max', the value will be set to the upper limit allowed by the OS.'",
+                    "type": "string",
```
kgiusti, line 1189: It might be easier to make the type "integer" and use qd_entity_opt_long(e, "backlog", 0) and treat the zero value as MAXCONN.

In today's brave new world of skupper humans no longer get to access to the config file directly - it's all through the skupper cli.  Having user friendly names like "max" aren't really necessary.

The other advantage is using and integer will fix the leak errors you're seeing in CI.

kgiusti (changes requested):
See inline, thanks.

PR #652, opened 2022-08-04
Fixes #651: Frees the qdr_connection_t and qdr_http2_connection_t obj…

### LINKED ISSUE #651: http2 adaptor leaks qdr_connection_t and qdr_http2_connection_t objects when connector and server deleted at the same time
A http2 server has a http2Connector to a server. When a management delete is run on the connector at the same time that the server is killed, the http2 adaptor leaks qdr_connection_t and qdr_http2_connection_t objects 

The following reproducer is courtesy of @kgiusti . You will see the leaked connections number increasing slowly

```
class Http2AdaptorIssue627Test(TestCase):
    """
    Test connector delete via managment during server connection loss
    """
    CONNECTOR_TYPE = 'io.skupper.router.httpConnector'
    PROTOCOL_VERSION = "HTTP2"

    @classmethod
    def setUpClass(cls):
        super(Http2AdaptorIssue627Test, cls).setUpClass()

        cls.test_name = 'HTTPAdaptorIssue627'

        config = Qdrouterd.Config([('router', {'mode': 'interior',
                                               'id': 'INTA'}),
                                   ('listener', {'role': 'normal',
                                                 'port': cls.tester.get_port()}),
                                   ('address', {'prefix': 'closest',   'distribution': 'closest'}),
                                   ('address', {'prefix': 'multicast', 'distribution': 'multicast'}),
                                   ])
        cls.INTA = cls.tester.qdrouterd('INTA', config, wait=True)

    def start_server(self, connector_port):
        server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        server.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        server.settimeout(TIMEOUT)
        server.bind(("", connector_port))
        server.listen(1)
        return server

    def test_01_issue_627(self):
        """Reproduce Issue #627"""
        mgmt = self.INTA.management

        for i in range(3000):
            connector_name = f"Connector_{self.test_name}_{i}"
            van_address = f"closest/{self.test_name}/{i}"
            connector_port = self.tester.get_port()


            cmd = ["test-receiver",
                   "-a", self.INTA.addresses[0],
                   "-c", "0"]
            rx = self.popen(cmd, expect=-15)

            cmd = ["test-sender",
                   "-a", self.INTA.addresses[0],
                   "-c", "0"]
            tx = self.popen(cmd, expect=-15)

            with self.start_server(connector_port) as server:
                attributes = {'address': van_address,
                              'host': '127.0.0.1',
                              'port': connector_port,
                              'protocolVersion': self.PROTOCOL_VERSION}
                mgmt.create(type=self.CONNECTOR_TYPE,
                            name=connector_name,
                            attributes=attributes)
                # server_conn, _ = server.accept()

                self.INTA.wait_address(van_address, subscribers=1)

                # server_conn.shutdown(socket.SHUT_RDWR)
                # server_conn.close()

            mgmt.delete(type=self.CONNECTOR_TYPE, name=connector_name)

            self.INTA.wait_address_unsubscribed(van_address, timeout=15.0)

            tx.terminate()
            tx.wait()
            tx.teardown()
            rx.terminate()
            rx.wait()
            rx.teardown()

            results = mgmt.query(type='io.skupper.router.connection').results
            print(f"Test cycle {i} complete {len(results)}!", flush=True)
```

ganeshmurthy:
…ect leaks by freeing them when a closed connection is being activated

### DIFF
--- a/src/adaptors/http2/http2_adaptor.c
+++ b/src/adaptors/http2/http2_adaptor.c
@@ -1625,6 +1625,7 @@ qdr_http2_connection_t *qdr_http_connection_ingress(qd_http_listener_t* listener
     ingress_http_conn->context.context = ingress_http_conn;
     ingress_http_conn->context.handler = &handle_connection_event;
     ingress_http_conn->listener = listener;
+    ingress_http_conn->admin_status = QD_CONN_ADMIN_ENABLED;
 
     // Incref the ref count on the listener since the qdr_http2_connection_t object is holding a ref to the listener
     sys_atomic_inc(&listener->ref_count);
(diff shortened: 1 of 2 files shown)


src/adaptors/http2/http2_adaptor.c
```diff
@@ -3237,8 +3253,15 @@ void qd_http2_delete_connector(qd_dispatch_t *qd, qd_http_connector_t *connector
         //
         // Deleting a connector must delete the corresponding qdr_connection_t and qdr_http2_connection_t objects also.
         //
-        if (connector->ctx)
-            qdr_core_close_connection((qdr_connection_t  *)connector->ctx);
+        if (connector->ctx) {
+            qdr_connection_t  *conn = (qdr_connection_t  *)connector->ctx;
```
jiridanek, line 3257: this line is not according to clang-format, it is not aligned with the lane below

ganeshmurthy: We should make the clang-format as part of the build ASAP. We should just bite the bullet and do it. I will fix this. Thx.

jiridanek: Maybe we can make `git clang-format` be part of precommit checks. Therefore, all new (added or modified) code will have to be formatted, while old code can remain unchanged.

Advantage: it should be reatively painless.

ganeshmurthy: I think we should do what you did earlier which is agree on a clang format, format *all* the code in the repo by running the tool and check all that modified code back into the repo. Then enable the tool in cmake so that it fails the build (locally and on CI) if the code violates the format. This is what I meant by bite the bullet and do it. :)

jiridanek: I know. I proposed to do this in two steps, instead, and soften the bullet for the first step.

kgiusti: @jiridanek 2 steps most definitely

jiridanek: * https://github.com/skupperproject/skupper-router/issues/660
* https://github.com/skupperproject/skupper-router/issues/661

Issue #543, opened 2022-06-06
TSAN failure in system_tests_http


ganeshmurthy:
```
Process 71179 (name=https-test-router) error: returned error code 66
skrouterd -c https-test-router.conf -I /home/tross/git/ted-ross/skupper-router/python
/home/tross/git/ted-ross/skupper-router/build/tests/system_test.dir/system_tests_http/RouterTestHttp/test_https_get/https-test-router-5.cmd
>>>>
==================
WARNING: ThreadSanitizer: data race (pid=71179)
  Read of size 8 at 0x0000005747f8 by main thread:
    #0 logger /home/tross/git/ted-ross/skupper-router/src/http-libwebsockets.c:58 (skrouterd+0x48e82c)
    #1 __lws_logv <null> (libwebsockets.so.19+0x293d0)
    #2 qd_http_server_listen /home/tross/git/ted-ross/skupper-router/src/http-libwebsockets.c:1029 (skrouterd+0x492fcf)
    #3 qd_listener_listen_http /home/tross/git/ted-ross/skupper-router/src/server.c:1644 (skrouterd+0x528de2)
    #4 qd_listener_listen /home/tross/git/ted-ross/skupper-router/src/server.c:1657 (skrouterd+0x528f4e)
    #5 qd_connection_manager_start /home/tross/git/ted-ross/skupper-router/src/connection_manager.c:860 (skrouterd+0x4816d3)
    #6 ffi_call_unix64 <null> (libffi.so.8+0x7745)
    #7 main_process /home/tross/git/ted-ross/skupper-router/router/src/main.c:99 (skrouterd+0x4267a3)
    #8 main /home/tross/git/ted-ross/skupper-router/router/src/main.c:363 (skrouterd+0x427631)

  Previous write of size 8 at 0x0000005747f8 by thread T3:
    #0 log_init /home/tross/git/ted-ross/skupper-router/src/http-libwebsockets.c:62 (skrouterd+0x48e8cf)
    #1 listener_start /home/tross/git/ted-ross/skupper-router/src/http-libwebsockets.c:320 (skrouterd+0x48f4dc)
    #2 http_thread_run /home/tross/git/ted-ross/skupper-router/src/http-libwebsockets.c:943 (skrouterd+0x492929)
    #3 _thread_init /home/tross/git/ted-ross/skupper-router/src/posix/threading.c:172 (skrouterd+0x4b4879)

[...]

PR #654, opened 2022-08-04
Enable s390x Travis CI build; Add coredump analyzer for Travis

jiridanek:
https://app.travis-ci.com/github/jiridanek/skupper-router/builds/254021857

```
$ sudo sysctl -w kernel.core_pattern="coredump.%e.%p"
sysctl: permission denied on key "kernel.core_pattern"
```

Looks like we have to live with the default pattern, which seems to be `core`.

The working of the coredump analysis can be tested by replacing the CI script with

```
python3 -c 'import os; import signal; os.kill(os.getpid(), signal.SIGABRT)'
```


jiridanek:
```
60: test_get_image_jpg (system_tests_http2_tls.Http2TestTlsStandaloneRouter) ... skipped 'Python 3.7 or greater, Quart 0.13.0 or greater and curl needed to run http2 tests'
```

https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/578785860#L4657

That line is worrisome. All requirements for the test should be fulfulled!

ganeshmurthy:
> ```
> 60: test_get_image_jpg (system_tests_http2_tls.Http2TestTlsStandaloneRouter) ... skipped 'Python 3.7 or greater, Quart 0.13.0 or greater and curl needed to run http2 tests'
> ```
> 
> https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/578785860#L4657
> 
> That line is worrisome. All requirements for the test should be fulfulled!
Somehow this function is returning True, which is why the tests are being skipped. Surely something is wrong with one of those three functions. This might be unrelated but this function - https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_http2.py#L39 - never returns False

jiridanek:
I'll try debug it tomorrow. Find out what exactly is returning (not true).

ganeshmurthy:
> I'll try debug it tomorrow. Find out what exactly is returning (not true).

Thanks much @jiridanek

jiridanek:
> > I'll try debug it tomorrow. Find out what exactly is returning (not true).
> 
> Thanks much @jiridanek

The reason was the following, https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/578900073#L298

The `pip install` was not updating jinja2 and markupsafe packages, because they were already installed systemwide. New version of quart did not work with the old versions of jinja2 or markupsafe. The return of `None` in the function to check Python version did not matter; None is falsey, so that works fine.

jiridanek:
So, currently, there is only a single real issue as far as I can tell

https://app.travis-ci.com/github/jiridanek/skupper-router/jobs/578903357#L4619

(and https://github.com/skupperproject/skupper-router/issues/543)

PR #659, opened 2022-08-04
Feature - Support multiple data connections between interior routers

### LINKED ISSUE #703: Allow creation of multiple data connections alongside inter-router connections to share streaming deliveries
Creating multiple data connections inter-router to carry streaming delivery data allows the inter-router link to utilize multiple threads and CPUs.  This has been demonstrated to dramatically increase cross-network throughput for service data.

ted-ross:


src/connection_manager.c
```diff
@@ -356,6 +363,11 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
     config->policy_vhost         = qd_entity_opt_string(entity, "policyVhost", 0);    CHECK();
     config->conn_props           = qd_entity_opt_map(entity, "openProperties");       CHECK();
 
+    if (strcmp(config->role, "inter-router") == 0) {
+        config->data_connection_count = qd_entity_opt_long(entity, "dataConnectionCount", 1); CHECK();
```
kgiusti, line 367: nit: validate data connection count value?  Probably should fail if zero - definitely fail if < 0.

src/connection_manager.c
```diff
@@ -745,12 +757,39 @@ QD_EXPORT qd_connector_t *qd_dispatch_configure_connector(qd_dispatch_t *qd, qd_
...
         ct->policy_vhost = qd_entity_opt_string(entity, "policyVhost", 0); CHECK();
         DEQ_ITEM_INIT(ct);
         DEQ_INSERT_TAIL(cm->connectors, ct);
         log_config(cm->log_source, &ct->config, "Connector");
 
+        //
+        // If this connection has a data-connection-group, set up the group members now
+        //
+        if (ct->config.has_data_connectors) {
+            qd_generate_discriminator(ct->group_correlator);
```
kgiusti, line 770: How would you feel about adding the router node id as part of the group_correlator value?  For debug trace of Open frames and probably prevent very unlikely naming collisions?

src/connection_manager.c
```diff
@@ -837,6 +877,23 @@ void qd_connection_manager_free(qd_connection_manager_t *cm)
         connector = DEQ_HEAD(cm->connectors);
     }
 
+    connector = DEQ_HEAD(cm->data_connectors);
```
kgiusti, line 880: nit: if you use a local list and append cm->connectors and cm->data_connectors can this code duplication be removed?

src/connection_manager.c
kgiusti: N.M.  Looks like a fixed size is necessary for allocation the correct sized tracking arrays, etc.

src/server.c
```diff
@@ -482,8 +485,20 @@ static void decorate_connection(qd_server_t *qd_server, pn_connection_t *conn, c
...
+    if (ctx->connector && ctx->connector->is_data_connector) {
+        pn_data_put_symbol(pn_connection_properties(conn),
+                           pn_bytes(strlen(QD_CONNECTION_PROPERTY_ROLE_KEY), QD_CONNECTION_PROPERTY_ROLE_KEY));
+        pn_data_put_int(pn_connection_properties(conn), QDR_ROLE_INTER_ROUTER_DATA);
+    }
+
+    if (ctx->connector && (ctx->connector->is_data_connector || ctx->connector->config.has_data_connectors)) {
+        pn_data_put_symbol(pn_connection_properties(conn),
+                           pn_bytes(strlen(QD_CONNECTION_PROPERTY_GROUP_CORRELATOR_KEY), QD_CONNECTION_PROPERTY_GROUP_CORRELATOR_KEY));
+        pn_data_put_string(pn_connection_properties(conn),
```
kgiusti, line 497: Looking at the protocol trace this appears to include the null termination byte when written to the wire:]

[C2]:FRAME: 0 -> @open(16) [container-id="TestRouterB", ... :"qd.conn-id"=2, :"qd.inter-router-role"=4, :"qd.group-correlator"="PJ2n9qvOMWiyG2y\x00"}]

QD_DISCRIMINATOR_SIZE is for sizing buffers to hold the discriminator and the trailing null octet.  Maybe just use strlen?

kgiusti (commented):
Aside from a few nits this looks like a nice improvement. Eventually we should merge this to main when it's fully baked.

kgiusti:
Thanks for the updates.

There's a use-after-free error in one of the tests that appears to be connector-related:

https://github.com/skupperproject/skupper-router/runs/8121523690?check_suite_focus=true#step:29:2851

The topology tests are... delicate.  We've discussed refactoring them for awhile now since they tend towards flakiness, but haven't gotten around to it.  This patch may force our hand.

src/server.c
```diff
@@ -585,7 +600,11 @@ qd_connection_t *qd_server_connection_impl(qd_server_t *server, qd_server_config
     ctx->connection_id = server->next_connection_id++;
     DEQ_INSERT_TAIL(server->conn_list, ctx);
     sys_mutex_unlock(&server->lock);
-    decorate_connection(ctx->server, ctx->pn_conn, config);
+    if (!!connector) {
+        ctx->connector = connector;
```
kgiusti, line 604: Should the sys_atomic_inc(&connector->ref_count) done on line 1141 be moved here where the reference is set?

kgiusti (approved):
I have one inline comment which may/may not be applicable, but otherwise good to go.

PR #663, opened 2022-08-05
Fixes #660 - Add pre-commit hooks

### LINKED ISSUE #660: Add git precommit with clang-format
This follows up to

* https://github.com/skupperproject/skupper-router/pull/504

jiridanek:
Here's how this fails in CI, https://github.com/skupperproject/skupper-router/runs/7690658958?check_suite_focus=true#step:4:11

You can skip hooks by either not installing them (see below) or by doing `git commit --no-verify`.

Here's how this fails locally, when hooks are installed

```
% git commit                              
git clang-format.........................................................Failed
- hook id: git_clang-format
- exit code: 1
- files were modified by this hook

ERROR: you need to run git clang-format on your commit
```

For local usage, it is necessary to install the pre-commit hook first, by running

```
python3 -m pip install -r requirements-dev.txt
pre-commit install
```

### DIFF
--- a/scripts/git-clang-format
+++ b/scripts/git-clang-format
@@ -1,0 +1,641 @@
+#!/usr/bin/env python3
+#
+#===- git-clang-format - ClangFormat Git Integration ---------*- python -*--===#
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+#===------------------------------------------------------------------------===#
+
+r"""
+clang-format git integration
+============================
+
+This file provides a clang-format integration for git. Put it somewhere in your
+path and ensure that it is executable. Then, "git clang-format" will invoke
+clang-format on the changes in current files or a specific commit.
+
+For further details, run:
+git clang-format -h
+
+Requires Python 2.7 or Python 3
+"""
+
+from __future__ import absolute_import, division, print_function
+import argparse
+import collections
+import contextlib
+import errno
+import os
+import re
+import subprocess
+import sys
+
+usage = ('git clang-format [OPTIONS] [<commit>] [<commit>|--staged] '
+         '[--] [<file>...]')
+
+desc = '''
+If zero or one commits are given, run clang-format on all lines that differ
+between the working directory and <commit>, which defaults to HEAD.  Changes are
+only applied to the working directory, or in the stage/index.
+
+Examples:
+  To format staged changes, i.e everything that's been `git add`ed:
+    git clang-format
+
+  To also format everything touched in the most recent commit:
+    git clang-format HEAD~1
+
+  If you're on a branch off main, to format everything touched on your branch:
+    git clang-format main
+
+If two commits are given (requires --diff), run clang-format on all lines in the
+second <commit> that differ from the first <commit>.
+
+The following git-config settings set the default of the corresponding option:
+  clangFormat.binary
+  clangFormat.commit
+  clangFormat.extensions
+  clangFormat.style
+'''
+
+# Name of the temporary index file in which save the output of clang-format.
+# This file is created within the .git directory.
+temp_index_basename = 'clang-format-index'
+
+
+Range = collections.namedtuple('Range', 'start, count')
+
+
+def main():
+  config = load_git_config()
+
+  # In order to keep '--' yet allow options after positionals, we need to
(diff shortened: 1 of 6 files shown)


scripts/git-clang-format
```diff
@@ -0,0 +1,660 @@
+#!/usr/bin/env python3
```
kgiusti, line 1: Is this a copy of the git-clang-format file that can be installed via 'dnf install git-clang-format'?

I would avoid creating our own copy that we'd have to maintain if possible and require folks to install this via their distro packages.

If their distro does not have a git-clang-format package then they can download it from llvm project as described in [code-conventions.md](https://github.com/skupperproject/skupper-router/blob/ed76beacbefac284df34400760df00bbc2b25dee/docs/notes/code-conventions.md?plain=1#L26)

scripts/git_clang-format_hook.sh
```diff
@@ -0,0 +1,48 @@
...
+# extract by the Jenkins build process located at the following URL:
+# https://builds.apache.org/view/M-R/view/Qpid/job/Qpid-proton-c/
+#
+
+# https://gist.github.com/alexeagle/c8ed91b14a407342d9a8e112b5ac7dab
+
+readonly help=$(cat <<- EOF
+  This repository requires you to install the git clang-format command.
+  One-time setup step (on Fedora Linux):
+     $ dnf install git-clang-format
```
kgiusti, line 33: Maybe include the url to git-clang-format as well just in case there isn't a git-clang-format package in their distro.

kgiusti (commented):
This doesn't work the way I'd expect - maybe I don't understand how to use it.

I've installed pre-commit and git-clang-format via dnf.

I've modified message.c to introduce format errors.  I then commit message.c.... and it commits without error.

I had expected the commit to fail due to the format errors but it did not.

Ideally we want to flag the error in a way that it's impossible to ignore... and commit upstream.

Also: can cmake check that git-clang-format and per-commit are installed and fail if not?

jiridanek:
> I've installed pre-commit and git-clang-format via dnf.

Did you do `pre-commit install` in the skupper-router directory?

scripts/git-clang-format
jiridanek: > Is this a copy of the git-clang-format file that can be installed via 'dnf install git-clang-format'?

Yes

> I would avoid creating our own copy that we'd have to maintain if possible and require folks to install this via their distro packages.

It's going to be low maintenance, and it is going to go away when all .c/.h files are correctly formatted; then we would no longer need to use it to limit the check to the changed lines.

I'm going to leave it in the PR for now, not touch anything, until you make the precommit work on your machine. Then we may make changes.

scripts/git-clang-format
```diff
@@ -0,0 +1,660 @@
...
+    git clang-format main
+
+If two commits are given (requires --diff), run clang-format on all lines in the
+second <commit> that differ from the first <commit>.
+
+The following git-config settings set the default of the corresponding option:
+  clangFormat.binary
+  clangFormat.commit
+  clangFormat.extensions
+  clangFormat.style
```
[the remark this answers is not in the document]
jiridanek: I am forcing this to --style=file below; another option is to force the user to have style set to file (that's what I saw used elsewhere). I think my way is best.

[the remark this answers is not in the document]
jiridanek: Anyways, if the user messes this up, the CI will wail on them afterwards, so, their loss and not a big deal for us

kgiusti:
> > I've installed pre-commit and git-clang-format via dnf.
> 
> Did you do `pre-commit install` in the skupper-router directory?

That's it: I missed that part.

Now when I attempt to commit a bad format I get:

]$ git commit -am "debug: do not merge"
[INFO] Initializing environment for local:clang-format.
[INFO] Installing environment for local.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
git clang-format.........................................................Failed
- hook id: git_clang-format
- exit code: 1
- files were modified by this hook

ERROR: you need to run git clang-format on your commit

The nice thing is that it already fixed the file locally.  All I needed was to re-run the commit and it succeeded.

scripts/git-clang-format
kgiusti: Sounds good, thanks

jiridanek: I'm going to write something to README.adoc, to explain things

kgiusti (approved):
Works well, thanks.

scripts/git-clang-format
kgiusti: +1 README.adoc.

PR #683, opened 2022-08-09
Fixes #677 - [coverity] double_free: Calling `qd_free_http_adaptor_config` frees pointer `config` which has already been freed

### LINKED ISSUE #677: [coverity] double_free: Calling `qd_free_http_adaptor_config` frees pointer `config` which has already been freed
image

image

The first report is correct, the second is just Coverity being dumb and unable to figure out that when `config` is freed, the function always returns early and the access it complains about does not happen. The fix for the first reported issue should cause the second false-positive go away.

jiridanek:


src/adaptors/http_common.c
```diff
@@ -84,9 +84,8 @@ qd_error_t qd_load_http_adaptor_config(qd_dispatch_t *qd, qd_http_adaptor_config
     //
     qd_error_t qd_error = qd_load_adaptor_config(qd, config->adaptor_config, entity, log_source);
     if (qd_error != QD_ERROR_NONE) {
-        qd_log(qd_log_source(QD_HTTP_LOG_SOURCE), QD_LOG_ERROR,
-               "Unable to load config information: %s", qd_error_message());
-        qd_free_http_adaptor_config(config);
+        qd_log(qd_log_source(QD_HTTP_LOG_SOURCE), QD_LOG_ERROR, "Unable to load config information: %s",
```
kgiusti, line 87: This does fix the problem, but it leaves a big potential foot-gun in qd_dispatch_configure_http_listener() where some poor soul looks at that code and says "Hey I found a leak - then need to free config on error."

I propose to change the qd_load_http_adaptor_config() to do the call to qd_http_adaptor_config() internally, then return the config pointer on success.  We can eliminate the qd_error_t in qd_dispatch_configure_http_listener() and simply have qd_load_http_adaptor_config() log the error and return NULL.

jiridanek: done

kgiusti (approved):

PR #706, opened 2022-09-01
Upgrade doctest.h: 2.4.8 -> 2.4.9

jiridanek:
* https://github.com/skupperproject/skupper-router/pull/792

### DIFF
--- a/tests/c_unittests/doctest.h
+++ b/tests/c_unittests/doctest.h
@@ -48,7 +48,7 @@
 
 #define DOCTEST_VERSION_MAJOR 2
 #define DOCTEST_VERSION_MINOR 4
-#define DOCTEST_VERSION_PATCH 8
+#define DOCTEST_VERSION_PATCH 9
 
 // util we need here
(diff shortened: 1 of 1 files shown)


jiridanek:
@kgiusti @ganeshmurthy Looks like the 2.4.9 version of doctest is somewhat broken, see

* https://github.com/doctest/doctest/issues/686
* https://github.com/doctest/doctest/issues/683
* https://github.com/doctest/doctest/issues/675
* https://github.com/doctest/doctest/issues/665

I am not sure which (if any) of these issues captures our problems, but in any case, we cannot adopt this version.

Looking at Catch (which is a very similar framework to doctest) as possible solution to use in the future (also, Proton uses it). It seems that Catch 3.0 decided not to be a single-header library, so its going to be a little harder to use it.

See this for intro to the Future Of Catch

* https://github.com/catchorg/Catch2/releases/tag/v3.0.1
* https://codingnest.com/the-future-of-catch2/

PR #709, opened 2022-09-07
Fixes #701: Use pooled adaptor buffers instead of ping pong buffers i…

### LINKED ISSUE #701: Remove the use of ping pong buffers in tcp adaptor and use pooled buffers
The TCP adaptor maintains one single read buffer and one single write buffer per connection. These buffers are given and gotten back from raw proton. The TCP adaptor maintains a flag to know who currently owns the buffer, either raw proton or the adaptor. Sometimes, when these buffers are under the possession of the raw connection, the raw connection asks for *more* buffers but the TCP adaptor waits for that single buffer to be given back to the adaptor upon which it gives back that same buffer to proton.

We could increase the performance of the TCP adaptor by using pooled buffers and giving proton as many read and write buffers that the raw connection API is asking for. I am doing this change as a proof of concept.

ganeshmurthy:
…n tcp adaptor. Introduced some common raw buffer read and write function in adaptor_common.c

### DIFF
--- a/src/adaptors/adaptor_buffer.c
+++ b/src/adaptors/adaptor_buffer.c
@@ -22,8 +22,8 @@
 #include "qpid/dispatch/ctools.h"
 
 // Buffer size will be set to 32k across all adaptors.
-// const size_t QD_ADAPTOR_MAX_BUFFER_SIZE = 16384;
-const size_t QD_ADAPTOR_MAX_BUFFER_SIZE = 32768;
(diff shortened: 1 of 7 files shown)


kgiusti (approved):
very minor nits - nothing to hold up this patch.

ganeshmurthy:
Changed the buffer size to 16 and removed all occurrences of sys.stdout.flush()

PR #720, opened 2022-09-13
Fixes #717: Add as many adaptor buffers that are needed to fill with …

### LINKED ISSUE #717: handle_outgoing() only writes one adaptor buffer per call even though there is a lot more BODY DATA available to be written

ganeshmurthy:
…the message BODY DATA


ganeshmurthy:
This is a draft PR, please do not merge.

PR #731, opened 2022-09-23
Issue 726

mgoulish:


mgoulish:
I don't understand what's happening here.
I canceled my previous PR because I wanted to get the two commits into a single one -- and now it has merge issues that look like this (small) change is already in the code.   !?!

PR #737, opened 2022-09-26
Fixes #567: Implemented TLS in the TCP adaptor. Created a new TLS API…

### LINKED ISSUE #567: Implement TLS in the TCP adaptor
Generalize the already existing TLS code in the http2 adaptor so it can be used across all adaptors.

ganeshmurthy:
… that can be shared across adaptors and modified the HTTP2 adaptor to use it.


python/skupper_router/management/skrouter.json
```diff
@@ -1229,6 +1229,28 @@
                     "graph": true,
                     "description": "The number of bytes sent from servers to clients on all connections to this listener."
                 },
+                "encryptedBytesIn": {
+                    "type": "integer",
+                    "graph": true,
+                    "description": "The number of encrypted bytes sent from clients to servers on all connections to this listener. This will be zero if the tcpListener does not have an associated sslProfile."
```
kgiusti, line 1235: Should be "total encrypted bytes"?  Is it "to servers" or is it "from all clients that attached via this listener"?  

The "to servers on all connections" doesn't make sense to me, but I usually defer to jross on all matters schema related...

python/skupper_router/management/skrouter.json
```diff
@@ -1229,6 +1229,28 @@
...
+                },
+                "encryptedBytesOut": {
+                    "type": "integer",
+                    "graph": true,
+                    "description": "The number of encrypted bytes sent from servers to clients on all connections to this listener. This will be zero if the tcpListener does not have an associated sslProfile."
```
kgiusti, line 1240: ditto

python/skupper_router/management/skrouter.json
```diff
@@ -1282,6 +1304,29 @@
...
+                    "type": "string",
+                    "required": false,
+                    "description": "Name of the sslProfile.",
+                    "create": true
+                },
+                "verifyHostname": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "yes: Ensures that when initiating a connection (as a client) the host name in the URL to which this connector connects to matches the host name in the digital certificate that the peer sends back as part of the TLS connection; no: Does not perform host name verification",
+                    "deprecationName": "verifyHostName",
```
kgiusti, line 1328: remove: this is a new attribute that never existed before so there was no name to deprecate, right?

src/adaptors/adaptor_common.c
```diff
@@ -291,6 +146,29 @@ int qd_raw_connection_write_buffers(pn_raw_connection_t *pn_raw_conn, qd_adaptor
...
+int qd_raw_connection_drain_read_write_buffers(pn_raw_connection_t *pn_raw_conn)
+{
+    pn_raw_buffer_t buffs[RAW_BUFFER_BATCH];
+    size_t          n;
+    int             buffers_drained = 0;
+    while ((n = pn_raw_connection_take_written_buffers(pn_raw_conn, buffs, RAW_BUFFER_BATCH))) {
+        for (size_t i = 0; i < n; ++i) {
+            buffers_drained++;
+            qd_adaptor_buffer_t *qd_adaptor_buffer = (qd_adaptor_buffer_t *) buffs[i].context;
+            free_qd_adaptor_buffer_t(qd_adaptor_buffer);
```
kgiusti, line 158: be consistent when using the api: call qd_adaptor_buffer_free()

src/adaptors/adaptor_common.c
```diff
@@ -291,6 +146,29 @@ int qd_raw_connection_write_buffers(pn_raw_connection_t *pn_raw_conn, qd_adaptor
...
+        }
+    }
+
+    while ((n = pn_raw_connection_take_read_buffers(pn_raw_conn, buffs, RAW_BUFFER_BATCH))) {
+        for (size_t i = 0; i < n; ++i) {
+            buffers_drained++;
+            qd_adaptor_buffer_t *qd_adaptor_buffer = (qd_adaptor_buffer_t *) buffs[i].context;
+            free_qd_adaptor_buffer_t(qd_adaptor_buffer);
```
kgiusti, line 166: ditto

tests/TCP_echo_client.py
```diff
@@ -231,7 +285,7 @@ def run(self):
 
         self.is_running = False
 
-    def wait(self, timeout=TIMEOUT):
+    def wait(self, timeout=TIMEOUT * 2):
```
kgiusti, line 288: ctest will fail the test if it takes > TIMEOUT anyway, why is TIMEOUT * 2 used?  Does it actually take 2 * 60 seconds to shut down the client?  If so that seems like a bug to me.

tests/system_tests_tcp_adaptor.py
```diff
@@ -112,8 +123,9 @@ class EchoClientRunner:
     def __init__(self, test_name, client_n, logger, client, server, size,
                  count,
                  print_client_logs=True,
-                 timeout=TIMEOUT,
-                 port_override=None):
+                 timeout=TIMEOUT * 2,
```
kgiusti, line 126: Why does this now require twice the time?

src/adaptors/adaptor_tls.h
```diff
@@ -0,0 +1,137 @@
...
+ * list is initialized and empty.
+ *
+ * @return QD_TLS_ERROR if there is an error when trying to decrypt an encrypted buffer, the number of undecrypted bytes
+ * totally processed otherwise
+ *
+ */
+int qd_tls_decrypt(qd_tls_t                 *tls,
+                   pn_raw_connection_t      *pn_raw_conn,
+                   qd_adaptor_buffer_list_t *decrypted_buffs,
+                   qd_adaptor_buffer_list_t *granted_read_buffs);
```
kgiusti, line 76: Why does this function take a list of buffers that were already granted to the pn_raw_connection_t?  Why not simply take available full read buffers from the raw connection?

src/adaptors/adaptor_tls.c
```diff
@@ -0,0 +1,645 @@
...
+    pn_raw_buffer_t raw_buffers[TLS_MAX_INPUT_CAPACITY];
+    size_t          n;
+    while ((n = pn_raw_connection_take_read_buffers(
+                pn_raw_conn,
+                raw_buffers,
+                fetch_size))) {  // Repeatedly get as much buffers (from raw connection) as possible
+        for (size_t i = 0; i < n; ++i) {
+            // Give each decrypt input buffer to proton tls
+            qd_adaptor_buffer_t *adaptor_buff = (qd_adaptor_buffer_t *) raw_buffers[i].context;
+            DEQ_REMOVE(*granted_read_buffs, adaptor_buff);
```
kgiusti, line 406: So the HTTP/1 adaptor now has to start tracking read buffers granted to the tls layer?  Why?

kgiusti (commented):
This patch is way too large.  Can you please break it up into a series of patches - it is very hard to review this.

python/skupper_router/management/skrouter.json
ganeshmurthy: Removed the deprecationName attribute

src/adaptors/adaptor_common.c
ganeshmurthy: Done

src/adaptors/adaptor_common.c
ganeshmurthy: Done

tests/system_tests_tcp_adaptor.py
ganeshmurthy: When the TCP over TLS tests are run, the echo client stays up for a duration on 60 seconds for the entire test run. Sometimes, in the CentOS stream case, the echo client would abruptly stop after 60 seconds but the router is slow and still in the process of sending large messages back to the echo client. I extended the timeout for this reason. This failure became more pronounced after I added a test called test_14_tcp_EA2_EC1_2000000. I will set the timeout back to TIMEOUT  and see what happens on the PR branch

src/adaptors/adaptor_tls.c
ganeshmurthy: I am researching this and will let you know what I find

ganeshmurthy:
> This patch is way too large. Can you please break it up into a series of patches - it is very hard to review this.

This patch consists of the adaptor_tls api, and related changes to http2 and tcp adaptors. How would you like me to split it up?

python/skupper_router/management/skrouter.json
ganeshmurthy: yep, we can change it to whatever @ssorj comes up with.

PR #738, opened 2022-09-26
Fixes #725: Configure clang-format to skip non-C files

### LINKED ISSUE #725: clang-format attempts to reformat skrouterd.json - non-C files should not be reformatted

jiridanek:


ganeshmurthy:
@kgiusti can you please take a quick look at this and approve ?

kgiusti (commented):
This patch didn't work for me - I'm using "git clang-format" command on my local repo.  With this patch "git clang-format" still reformats the python/skupper_router/management/skrouterd.json file if I make any changes to it.

jiridanek:
There is

```
The following git-config settings set the default of the corresponding option:
  clangFormat.binary
  clangFormat.commit
  clangFormat.extension
  clangFormat.style
```

... afaik this is something everyone needs to setup for themselves, it is not stored in the repository when you clone it initially

jiridanek:
Looks like this is possible if I explicitly blacklist jsons in `.clang-format`, analogous to

```
Language: Proto
# Don't format .proto files.
DisableFormat: true
```

jiridanek:
@kgiusti please re-review; now it should work the way you expected

Issue #672, opened 2022-08-07
When Timeout happens, the correct response is to get stacktraces of every subprocess and only then stop Python test


jiridanek:
https://github.com/skupperproject/skupper-router/runs/7679443495?check_suite_focus=true#step:10:5169

* see also https://github.com/skupperproject/skupper-router/issues/637

Issue #637, opened 2022-07-26
`skrouterd` shutdown timeouted in SkstatTest.test_yy_query_many_links


jiridanek:
https://github.com/skupperproject/skupper-router/runs/7522672136?check_suite_focus=true#step:29:1832

```
___________________________________________________ ERROR at teardown of SkstatTest.test_yy_query_many_links ___________________________________________________

cls = <class 'tests.system_tests_skstat.SkstatTest'>

    @classmethod
    def tearDownClass(cls):
        if hasattr(cls, 'tester'):
>           cls.tester.teardown()

/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:1086: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:1031: in teardown
    cleanup()
/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:598: in teardown
    super(Qdrouterd, self).teardown()
/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:368: in teardown
    self.wait(TIMEOUT)
/usr/lib64/python3.10/subprocess.py:1207: in wait
    return self._wait(timeout=timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Qdrouterd: returncode: None args: ['skrouterd', '-c', 'test-router.conf', '...>, timeout = 300.0

    def _wait(self, timeout):
        """Internal implementation of wait() on POSIX."""
        if self.returncode is not None:
            return self.returncode
    
        if timeout is not None:
            endtime = _time() + timeout
            # Enter a busy loop if we have a timeout.  This busy loop was
[...]

Issue #535, opened 2022-06-02
Http1AdaptorQ2Standalone::test_02_backpressure_server: Process 3642 (name=Q2Router2) error: did not terminate properly, required SIGABORT


jiridanek:
https://github.com/skupperproject/skupper-router/runs/6698510074#step:29:84

```
Thread 5 (Thread 0x7f18f55cb640 (LWP 3643)):
#0  0x00007f18fa331605 in __GI___clock_nanosleep (clock_id=clock_id@entry=0, flags=flags@entry=0, req=req@entry=0x7f18f558a5a0, rem=rem@entry=0x0) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:48
#1  0x00007f18fa3361a7 in __GI___nanosleep (req=req@entry=0x7f18f558a5a0, rem=rem@entry=0x0) at ../sysdeps/unix/sysv/linux/nanosleep.c:25
#2  0x00007f18fa363189 in usleep (useconds=<optimized out>) at ../sysdeps/posix/usleep.c:32
#3  0x00007f18fa92f53b in __interceptor_usleep (usec=100000) at ../../../../libsanitizer/tsan/tsan_interceptors_posix.cpp:359
#4  0x00007f18fa95a529 in __tsan::BackgroundThread (arg=<optimized out>) at ../../../../libsanitizer/tsan/tsan_rtl.cpp:180
#5  0x00007f18fa2bdda2 in start_thread (arg=<optimized out>) at pthread_create.c:443
#6  0x00007f18fa25d9e0 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

Thread 4 (Thread 0x7f18ecc03640 (LWP 3647)):
#0  __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7b2800014ff4) at futex-internal.c:57
#1  __futex_abstimed_wait_common (futex_word=futex_word@entry=0x7b2800014ff4, expected=expected@entry=0, clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0, cancel=cancel@entry=true) at futex-internal.c:87
[...]

Issue #57, opened 2022-02-09
Remove link-routing capability


ted-ross:
Remove the link-routing support.  This will greatly simplify testing and future development of new routing features.  Most of the unique and valuable benefits of link-routing can be achieved using message streaming and non-terminal delivery dispositions.

PR #743, opened 2022-09-28
Fixes #742: Removed granted read buffer tracking in http2 and tcp ada…

### LINKED ISSUE #742: Remove granted read buffer tracking in http2 and tcp adaptors
The read buffers are tracked in the http2 and tcp adaptors using a list called granted_read_buffs.

Instead of tracking these buffers, modify the router shutdown to delete all adaptor connectors and listeners before the router shuts down. 
This will make sure that the buffers are cleaned up upon the receipt of an empty PN_RAW_CONNECTION_READ event on connection close.

This assumes that deletion of connectors deletes the actual connection as well.

ganeshmurthy:
…ptors

### DIFF
--- a/src/adaptors/adaptor_common.h
+++ b/src/adaptors/adaptor_common.h
@@ -85,19 +85,16 @@ bool qd_tls_initial_setup(qd_adaptor_config_t *config,
                           const char          *protocols[]);
 
 /**
- * Grants as many read qd_adaptor buffers as returned by pn_raw_connection_read_buffers_capacity() with a maximum of
- * RAW_BUFFER_BATCH(16) read buffers. Stuffs the granted adaptor buffers into the passed in granted_read_buffs list if
- * provided.
+ * Grants as many read qd_adaptor buffers as returned by pn_raw_connection_read_buffers_capacity().
+ * Maximum read capacity is set to 16 in proton raw api.
  *
  * @param raw_conn - The pn_raw_connection_t to which read buffers are granted.
- * @param granted_read_buffs - (optional) the qd_adaptor_buffer_list which will container the granted buffs, else 0
  */
-int qd_raw_connection_grant_read_buffers(pn_raw_connection_t      *pn_raw_conn,
-                                         qd_adaptor_buffer_list_t *granted_read_buffs);
+int qd_raw_connection_grant_read_buffers(pn_raw_connection_t *pn_raw_conn);
 
 /**
- * Writes as many adaptor buffers as allowed by pn_raw_connection_write_buffers_capacity() with a maximum of 16 write
- * buffers.
+ * Writes as many adaptor buffers as allowed by pn_raw_connection_write_buffers_capacity().
+ * Maximum write capacity is set to 16 in proton raw api.
  *
  * @param raw_conn - The pn_raw_connection_t to which read buffers are granted.
  * @param blist - qd_adaptor_buffer_list_t which contains that buffers that need to be written.
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -272,14 +272,26 @@ void qdr_record_link_credit(qdr_core_t *core, qdr_link_t *link)
 
 void qdr_close_connection_CT(qdr_core_t *core, qdr_connection_t  *conn)
 {
+    bool conn_already_closed = false;
     sys_mutex_lock(&conn->work_lock);
-    conn->closed = true;
-    conn->error  = qdr_error(QD_AMQP_COND_CONNECTION_FORCED, "Connection forced-closed by management request");
+    //
+    // In some cases, this function can be called on an already closed connection in which case
+    // we don't want to leak a qdr_error_t object
+    //
+    if (conn->closed) {
+        conn_already_closed = true;
+    } else {
+        conn->closed = true;
+        conn->error  = qdr_error(QD_AMQP_COND_CONNECTION_FORCED, "Connection forced-closed by management request");
+    }
+
     conn->admin_status = QD_CONN_ADMIN_DELETED;
     sys_mutex_unlock(&conn->work_lock);
 
-    //Activate the connection, so the I/O threads can finish the job.
-    qdr_connection_activate_CT(core, conn);
+    if (!conn_already_closed) {
+        // Activate the connection, so the I/O threads can finish the job.
+        qdr_connection_activate_CT(core, conn);
+    }
 }
 
 
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -64,10 +64,15 @@ from proton.reactor import AtLeastOnce, Container
 from proton.reactor import AtMostOnce
 
 from skupper_router.management.client import Node
-from skupper_router.management.error import NotFoundStatus
+from skupper_router.management.error import NotFoundStatus, BadRequestStatus
 
 # Optional modules
 MISSING_MODULES = []
+HTTP_LISTENER_TYPE = 'io.skupper.router.httpListener'
+TCP_LISTENER_TYPE = 'io.skupper.router.tcpListener'
+HTTP_CONNECTOR_TYPE = 'io.skupper.router.httpConnector'
+TCP_CONNECTOR_TYPE = 'io.skupper.router.tcpConnector'
+CONNECTION_TYPE = 'io.skupper.router.connection'
 
 try:
     import qpidtoollibs  # pylint: disable=unused-import
@@ -567,6 +572,7 @@ class Qdrouterd(Process):
         args = shlex.split(os.environ.get('QPID_DISPATCH_RUNNER', '')) + args
         super(Qdrouterd, self).__init__(args, name=name, expect=expect)
         self._management = None
+        self._qd_manager = None
         self._wait_ready = False
         if wait:
             self.wait_ready()
@@ -577,6 +583,20 @@ class Qdrouterd(Process):
 
     def __exit__(self, *_):
         self.teardown()
+
+    @property
+    def qd_manager(self):
+        if not self._qd_manager:
+            try:
+                self._qd_manager = QdManager(address=self.addresses[0])
+            except Exception as e:
+                # As it stands now, not all router configs are expected to have a valid amqp listening
+                # port on self.addresses[0]. So for now, it is ok to return None
+                # In the future we might require all routers to have a valid amqp listening port available
+                # on self.addresses[0] at which point, we should remove this except clause and
+                # force a failure if there is no self.addresses[0]
+                return None
+        return self._qd_manager
 
     @property
     def management(self):
@@ -594,6 +614,31 @@ class Qdrouterd(Process):
             self._management = None
 
         teardown_exc = None
+
+        try:
+            if self.qd_manager:
+                # Delete all adaptor connectors and listeners before shutting down the router
+                long_types = [HTTP_LISTENER_TYPE, HTTP_CONNECTOR_TYPE, TCP_LISTENER_TYPE, TCP_CONNECTOR_TYPE]
+                for long_type in long_types:
+                    self.qd_manager.delete_all_entities(long_type)
+                retry_assertion(self.qd_manager.delete_adaptor_connections)
+        except (IndexError, TypeError, BadRequestStatus):
+            # The router might not have any amqp listeners at all
+            # These are known exceptions we can ignore.
+            pass
+        except Exception as e:
+            # QdManager class wraps all exceptions in Exception.
+            # BadRequestStatus happens when you are trying to delete a connection that is already deleted.
+            # The delete of the connector might delete the connection and we might try to delete that connection
+            # again and that is ok.
+            exception_string = str(e)
+            if "ConnectionException" in exception_string or \
+                    "LinkDetached" in exception_string or \
+                    "ConnectionClosed" in exception_string or \
+                    "BadRequestStatus" in exception_string:
+                pass
+            else:
+                raise e
         try:
             super(Qdrouterd, self).teardown()
         except Exception as exc:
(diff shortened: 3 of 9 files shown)


jiridanek (commented):
> Here is the test failure -
https://github.com/skupperproject/skupper-router/actions/runs/3146738179/jobs/5115896447#step:29:2330
The timeout of 400 seconds, is it set here - https://github.com/skupperproject/skupper-router/blob/main/.github/workflows/build.yaml#L167
Is that an individual timeout per test of 400 seconds ?

Yes, SkstatTest::test_yy_query_many_links had 400 seconds to finish, but it did not finish. I logged issue for this test already https://github.com/skupperproject/skupper-router/issues/637

Also, I had some thoughts in https://github.com/skupperproject/skupper-router/issues/672, if we had that, it would be easier to debug this. Currently we cannot tell if this problem is caused by https://github.com/skupperproject/skupper-router/issues/535 or if it is a completely new issue.
 
> Also, the PR is attempting to close all outstanding adaptor (tcp, http) listeners, connectors and non amqp connections before the router shuts down in test. This way, I don't have to track granted read buffers in the adaptor code. But those buffers are still leaking here - https://github.com/skupperproject/skupper-router/pull/743/checks#step:6:505
It looks like we will have to do the same thing (shutdown all connections) in the container image test that is running h2spec.

What the test does now is described in the comment https://github.com/skupperproject/skupper-router/blob/000de22ce0049c2079d462fca669cc85f0be07e9/tests/testcontainers/test_h2spec.rs#L92-L99

What you suggest is that we need to run skmanage, find out what connections are active, and terminate the connections using a management operation? Or is the trick to shutdown the nghttpd first, and then shutdown the router? How do I detect that the connection is already scrapped in the router? With a skmanage call?

Wouldn't it be simpler to teardown all connections in the router itself if it is running in a memory-debug mode?

jiridanek:
@ganeshmurthy When I put https://github.com/skupperproject/skupper-router/pull/750 on top of this, the leak from image tests disappears for me

ganeshmurthy:
> https://github.com/skupperproject/skupper-router/actions/runs/3159057422/jobs/5141859848#step:10:5046
> 
> ```
> 33/33 Test #57: system_tests_http1_adaptor ...........***Timeout 1199.71 sec
> ```
> 
> What just happened?

I don't think that failure has anything to do with this PR in particular but I am not sure. This PR is trying to shut down all connections just before the router shuts down, so I am hoping that the PR did not cause the error. @kgiusti, what do you think ?

src/adaptors/adaptor_common.h
```diff
@@ -86,14 +86,11 @@ bool qd_tls_initial_setup(qd_adaptor_config_t *config,
 
 /**
  * Grants as many read qd_adaptor buffers as returned by pn_raw_connection_read_buffers_capacity() with a maximum of
- * RAW_BUFFER_BATCH(16) read buffers. Stuffs the granted adaptor buffers into the passed in granted_read_buffs list if
- * provided.
+ * RAW_BUFFER_BATCH(16) read buffers.
```
kgiusti, line 89: > with a maximum of RAW_BUFFER_BATCH(16) read buffers.

Really? Reading the code that doesn't seem to be the case - it will always grant read_buffers_capacity().  It is not limited to RAW_BUFFER_BATCH - it simply passes at most RAW_BUFFER_BATCH buffers in one call to pn_raw_connection_give_read_buffers. The outer loop calls pn_raw_connection_give_read_buffers() multiple times to fill the entire capacity.

Which is correct - it should fill the capacity. My point is that it appears that the comment is wrong.

src/router_core/connections.c
```diff
@@ -273,8 +273,14 @@ void qdr_record_link_credit(qdr_core_t *core, qdr_link_t *link)
 void qdr_close_connection_CT(qdr_core_t *core, qdr_connection_t  *conn)
 {
     sys_mutex_lock(&conn->work_lock);
-    conn->closed = true;
-    conn->error  = qdr_error(QD_AMQP_COND_CONNECTION_FORCED, "Connection forced-closed by management request");
+    //
+    // In some cases, this function can be called on an already closed connection in which case
```
kgiusti, line 277: Calling this twice seems like a bug - is this new behavior caused by this patch?  The adaptors are not calling the qdr_core_close_connection() API twice on the same qdr_connection_t, are they?  Or is the management agent attempting to re-delete a deleted connection?

kgiusti (commented):
I like what this does, but I'm concerned that there's something not right if the router is calling close connection twice now.

I would also break this up into two separate PRs: one patch to the CI that does the adaptor/connector/connection cleanup and a follow up PR patch that removes the adaptor buffer tracking.  These really are two separate fixes.

tests/system_test.py
```diff
@@ -66,8 +66,14 @@
 from skupper_router.management.client import Node
 from skupper_router.management.error import NotFoundStatus
 
+# Optional modules
```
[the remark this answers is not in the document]
ganeshmurthy: Not sure how that got added, will remove

src/router_core/connections.c
ganeshmurthy: It is the latter and it happens in the http2 adaptor. When a connector is deleted in the http2 adaptor, it immediately (but asynchronously tries) to delete the associated connection (in http2 adaptor, one connector can have only one associated connection). While the core thread is actually trying to delete the connection, the management command comes in to tries to delete the connection as well. This happened 1 out of 10 times in my testing, so I added that bit code to check if the conn is already closed to prevent a qdr_error leak.

ganeshmurthy:
> I like what this does, but I'm concerned that there's something not right if the router is calling close connection twice now.
> 
> I would also break this up into two separate PRs: one patch to the CI that does the adaptor/connector/connection cleanup and a follow up PR patch that removes the adaptor buffer tracking. These really are two separate fixes.

The reason, I had all this in the same patch is because these are related. Deleting the connectors/connections is what is allowing me to take out the tracking of the granted_read_buffers from the adaptors. If I do a PR with the management change separately and another PR which just removes the granted_read_buffs, I am sure someone might wonder looking at the second PR why suddenly we are able to remove the granted_read_buffs.

tests/system_test.py
```diff
@@ -594,6 +613,22 @@ def teardown(self):
             self._management = None
 
         teardown_exc = None
+
+        try:
+            qd_manager = QdManager(address=self.addresses[0])
```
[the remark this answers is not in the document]
kgiusti: Related #753

tests/system_test.py
```diff
@@ -594,6 +613,22 @@ def teardown(self):
...
+            # Delete all adaptor connectors and listeners before shutting down the router
+            long_types = [HTTP_LISTENER_TYPE, HTTP_CONNECTOR_TYPE, TCP_LISTENER_TYPE, TCP_CONNECTOR_TYPE]
+            for long_type in long_types:
+                qd_manager.delete_all_entities(long_type)
+            retry_assertion(self.delete_adaptor_connections(qd_manager))
+        except Exception as e:
+            # We tried to delete some stuff but ran into issues. There could
```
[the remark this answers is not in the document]
jiridanek: +1, also, catching `Exception` will catch SyntaxError, etc... Best to have explicit list, esp. because it is hard to improve this later on

src/router_core/connections.c
kgiusti: Ah! Ok, so if I understand correctly:

1. The CI change adds a management call to delete the _connection_ - which is cool.
2. At the same moment the adaptor is servicing a previous management request to delete the _connector_, which then correctly calls qdr_core_close_connection()

So at the core we have two connection delete paths - one via the management agent and one via the action queued by the adaptor thread - both are valid and can potentially happen simultaneously "in the field".

Given that I agree the check for conn->closed is the correct move.

But - and it seems I always have a big but, right? - I wouldn't activate the connection if conn->closed was already true. That will cause the connection process to call the close handler twice which it probably shouldn't do.

See next comment.

src/router_core/connections.c
```diff
@@ -273,8 +273,14 @@ void qdr_record_link_credit(qdr_core_t *core, qdr_link_t *link)
...
+    // we don't want to leak a qdr_error_t object
+    //
+    if (!conn->closed) {
+        conn->closed = true;
```
kgiusti, line 281: In here I'd try setting a local flag to call the activate once the lock is dropped.  That would ensure that the close handler is not invoked twice.

Try that.  Does it break anything?

src/adaptors/adaptor_common.h
ganeshmurthy: I fixed the comment.

src/router_core/connections.c
ganeshmurthy: I used a local flag to call activate outside the lock and it does the job, does not break anything. I am running the http2 and http2_tls tests in a loop with -j4 and will run it for some more time to see if it catches anything

tests/system_test.py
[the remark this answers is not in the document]
ganeshmurthy: Sometimes a proton ConnectionException is thrown which proton does not seem to expose ?

tests/system_test.py
```diff
@@ -1453,6 +1488,14 @@ def get_log(self, limit=None):
...
         return json.loads(self(cmd))
 
+    def delete_all_entities(self, long_entity_type):
+        # First, query the router to get all entities of entity type
+        results = self.query(long_entity_type)
+        for result in results:
+            entity_name = result.get('name')
+            if entity_name:
+                self.delete(long_entity_type, name=entity_name)
+
```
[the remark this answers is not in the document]
ganeshmurthy: Since the delete command is synchronous, the delete is successful (meaning the entity is deleted) by the time the response to the delete command reaches qdmanage. So, does adding an additional retry loop buy us anything since we already know that the entity has already been deleted ?
If the delete command is going to start a core thread action, we never know when that is going to complete ?

tests/system_test.py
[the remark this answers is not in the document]
jiridanek: It should be exposed

https://github.com/apache/qpid-proton/blob/a4375a8351c3435bd5025bb2481307efe97dc99e/python/proton/__init__.py#L57

so

```
from proton import ConnectionException
```

?

kgiusti:
> > https://github.com/skupperproject/skupper-router/actions/runs/3159057422/jobs/5141859848#step:10:5046
> > ```
> > 33/33 Test #57: system_tests_http1_adaptor ...........***Timeout 1199.71 sec
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What just happened?
> 
> I don't think that failure has anything to do with this PR in particular but I am not sure. This PR is trying to shut down all connections just before the router shuts down, so I am hoping that the PR did not cause the error. @kgiusti, what do you think ?

Oh great... yet another buffer corruption issue!

From the server log:

2022-09-30 13:26:40.230373 +0000 HTTP_ADAPTOR (trace) [C1] HTTP request/response codec done. **Octets read: 393570 written: 210** (../src/adaptors/http1/http1_server.c:1176)

Yep, those numbers look good.  Now what did the client see?  From the client log:

2022-09-30 13:31:40.304957 +0000 HTTP_ADAPTOR (trace) [C66] HTTP request msg-id=18 cancelled. **Octets read: 210 written: 195584** (../src/adaptors/http1/http1_client.c:965)

TL;DR test client is hanging since it hasn't received the full response although the server sent it.

I'll go re-open the old Issue... Oh look! It's about Beer O'clock on a Friday!  So long, folks!!

src/router_core/connections.c
ganeshmurthy: Used a local flag (outside of lock) to avoid activation and it worked !

tests/system_test.py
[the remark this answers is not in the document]
ganeshmurthy: Also, there  is a loop that deletes all connections as well.

tests/system_test.py
```diff
@@ -578,6 +584,15 @@ def __enter__(self):
     def __exit__(self, *_):
         self.teardown()
 
+    @property
+    def qd_manager(self):
+        if not self._qd_manager:
+            try:
+                self._qd_manager = QdManager(address=self.addresses[0])
+            except Exception as e:
+                return None
```
kgiusti, line 593: Probably don't want to silently discard the exception - it will be very hard to figure out what happened if the QdManager fails in GH actions.

tests/system_test.py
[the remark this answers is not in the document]
kgiusti: If I understand correctly the problem is that qdmanager is used to do the operation. Qdmanager isn't using proton directly - it spawns a process that runs skmanage as a shell process then simply raises Exception should the process return an error code.

So we lose any access to the proton exception details.

tests/system_test.py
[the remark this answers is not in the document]
kgiusti: There are multiple ways of de-furring this particular feline. I'll leave it up to the judgment of whomever is brave enough to take on #753.

tests/system_test.py
ganeshmurthy: > Probably don't want to silently discard the exception - it will be very hard to figure out what happened if the QdManager fails in GH actions.

You mean, it will be hard to figure out why the QdManager constructor fails in GHA ? The only one reason it could fail is if there is no self.addresses, right ? I will add a print(e) before returning None

kgiusti: Is not having a self.addresses[0] a bug, or is it OK as long as the caller checks for a None?  I think that's my point of confusion: I don't understand why the exception is being ignored. If the API allows a None return code (not a bug), then I would simply comment the code to explain that is why the exception is discarded.

tests/system_test.py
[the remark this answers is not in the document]
ganeshmurthy: QdManager does indeed wrap all exceptions into an Exception object but the exception details are not lost. QdManager calls ```raise Exception("%s %s" % rc)``` upon failure which the following error message (for example a ConnectionException) prints - 
```ConnectionException: Connection amqp://0.0.0.0:23860 disconnected: Condition('proton.pythonio', 'Connection refused to all addresses')``` - prints on a proton ConnectionException

kgiusti (approved):

tests/system_test.py
ganeshmurthy: As it stands now, not all routers have a valid self.addresses[0], so it is ok to return a None. I will add a comment clarifying that. But the person who fixes https://github.com/skupperproject/skupper-router/issues/753 will make not having a self.addresses[0] illegal and require every router to have a management port on self.addresses[0]

PR #750, opened 2022-09-30
Issue #742 - Shut down router last in container tests, to prevent shutdown memory leaks due to unfinalzed connections

jiridanek:


ganeshmurthy (approved):
LGTM, please merge ASAP. Thanks.

Issue #760, opened 2022-10-03
system_tests_tcp_adaptor times out on test_12_tcp_EA2_EC1_5000000


ganeshmurthy:
Sometimes, the system_tests_tcp adaptor times out when executing the test - test_12_tcp_EA2_EC1_5000000

```
 59: DEBUG    root:system_test.py:1696 2022-10-03 19:08:19.760933 TcpAdaptor: ECHO_SERVER TcpAdaptor ES_EC1 write to : 127.0.0.1:46594 len:1024: b'aaaaaaaaaaaa[localhost:24895:0:4285450]aaaaaaaaaaa' ... b'4895:0:4286400]aaaaaaaaaaaaaaaaaaaaaaa[localhost:2'
59: DEBUG    root:system_test.py:1696 2022-10-03 19:08:19.761082 TcpAdaptor: ECHO_SERVER TcpAdaptor ES_EC1 read from: 127.0.0.1:46594 len:1024: b'4895:0:4286450]aaaaaaaaaaaaaaaaaaaaaaa[localhost:2' ... b'aaaaaaaaaaaaaa[localhost:24895:0:4287450]aaaaaaaaa'
59: DEBUG    root:system_test.py:1696 2022-10-03 19:08:19.761223 TcpAdaptor: ECHO_SERVER TcpAdaptor ES_EC1 write to : 127.0.0.1:46594 len:1024: b'4895:0:4286450]aaaaaaaaaaaaaaaaaaaaaaa[localhost:2' ... b'aaaaaaaaaaaaaa[localhost:24895:0:4287450]aaaaaaaaa'
59: DEBUG    root:system_test.py:1696 2022-10-03 19:08:19.803188 TcpAdaptor-testClass: TCP_TEST test_12_tcp_EA2_EC1_5000000 failed: TCP_TEST TIMEOUT - local wait time exceeded
59: DEBUG    root:system_test.py:1696 2022-10-03 19:08:19.761378 TcpAdaptor: ECHO_SERVER TcpAdaptor ES_EC1 read from: 127.0.0.1:46594 len:1024: b'aaaaaaaaaaaaaa[localhost:24895:0:4287500]aaaaaaaaa' ... b':24895:0:4288450]aaaaaaaaaaaaaaaaaaaaaaa[localhost'
59: ------------ generated xml file: /home/runner/work/skupper-router/skupper-router/skupper-router/build/tests/junitxmls/system_tests_tcp_adaptor.xml -------------
[...]

PR #762, opened 2022-10-03
Fixes #761: Introduced a common function - qd_raw_connection_drain_re…

### LINKED ISSUE #761: Introduce a common function in adaptor_common.h/c to drain read and write buffers
This function will be called when we process the ```PN_RAW_CONNECTION_DRAIN_BUFFERS``` event in the adaptors

ganeshmurthy:
…ad_write_buffers in adaptor_common.c/h. Modified the adaptors to call this function on the PN_RAW_CONNECTION_DRAIN_BUFFERS event

### DIFF
--- a/src/adaptors/adaptor_common.c
+++ b/src/adaptors/adaptor_common.c
@@ -300,3 +300,27 @@ char *qd_raw_conn_get_address(pn_raw_connection_t *pn_raw_conn)
         return strndup(buffer, 1024);
     }
 }
+
+int qd_raw_connection_drain_read_write_buffers(pn_raw_connection_t *pn_raw_conn)
+{
+    pn_raw_buffer_t buffs[RAW_BUFFER_BATCH];
+    size_t          n;
+    int             buffers_drained = 0;
+    while ((n = pn_raw_connection_take_written_buffers(pn_raw_conn, buffs, RAW_BUFFER_BATCH))) {
+        for (size_t i = 0; i < n; ++i) {
+            buffers_drained++;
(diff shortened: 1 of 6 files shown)


src/adaptors/adaptor_common.c
```diff
@@ -300,3 +300,26 @@ char *qd_raw_conn_get_address(pn_raw_connection_t *pn_raw_conn)
...
+            buffers_drained++;
+            qd_adaptor_buffer_t *qd_adaptor_buffer = (qd_adaptor_buffer_t *) buffs[i].context;
+            qd_adaptor_buffer_free(qd_adaptor_buffer);
+        }
+    }
+
+    while ((n = pn_raw_connection_take_read_buffers(pn_raw_conn, buffs, RAW_BUFFER_BATCH))) {
+        for (size_t i = 0; i < n; ++i) {
+            buffers_drained++;
+            qd_adaptor_buffer_t *qd_adaptor_buffer = (qd_adaptor_buffer_t *) buffs[i].context;
```
kgiusti, line 320: Please add an assert here that fails if any of these read buffers contain valid data. In other words: assert(bufs[i].size == 0)

Zero-cost defensive programming.  Trust no one.  Even me...

kgiusti (commented):
See comment.

The tcp tests are failing on ubuntu with this patch.

ganeshmurthy:
> See comment.
> 
> The tcp tests are failing on ubuntu with this patch.

I have raised an issue for the TCP test failure - https://github.com/skupperproject/skupper-router/issues/760
I have raised a PR to fix the TCP test failure as well. Thx

PR #766, opened 2022-10-04
Change commit message format to use colon instead of a dash

### LINKED ISSUE #90: Q2 tests in system_tests_http2 must check for equal number of blocked and unblocked q2 messages
Currently  test_q2_block_unblock does the following 

```
self.assertGreater(num_blocked, 0)
self.assertGreater(num_unblocked, 0)
```
It should also do the following - 
```
self.assertGreaterEqual(num_unblocked, num_blocked)
```

jiridanek:
It seems that people have converged to using a colon in the commit messages, so let's make it official

    Fixes #90: Fix code in http2 adaptor that was doing improper q2 related logging

(Anything else in there that should change?)

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -21,13 +21,13 @@ Whenever you want to commit a change to the skupper-router repo, please first cr
 When creating a PR, please use the following commit message pattern
 
 ```
-Fixes #<github-issue-number> - <Commit message in present tense>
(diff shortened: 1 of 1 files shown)


ganeshmurthy (approved):

mgoulish:
It seems reasonable to me; I find colons more clear.

kgiusti (approved):

Issue #758, opened 2022-10-01
`system_tests_http1_adaptor` in `Http1AdaptorEdge2EdgeTest.test_05_large_streaming_msg` failed with `AssertionError: 4 != 3`


jiridanek:
https://github.com/jiridanek/skupper-router/actions/runs/3164478321/jobs/5152790724#step:29:2969

```
57:         with TestServer.new_server(self.http_server11_port, self.http_listener11_port, TESTS_11) as server11, \
57:              TestServer.new_server(self.http_server10_port, self.http_listener10_port, TESTS_10,
57:                                    handler_cls=RequestHandler10) as server10:
57:             self.EA2.wait_connectors()
57:     
57:             client11 = ThreadedTestClient(TESTS_11,
57:                                           self.http_listener11_port,
57:                                           repeat=2)
57:             client11.wait()
57:             self.assertIsNone(client11.error)
57: >           self.assertEqual(4, client11.count)
57: E           AssertionError: 4 != 3
57: 
57: /home/runner/work/skupper-router/skupper-router/skupper-router/tests/http1_tests.py:887: AssertionError
```

This appears similar to other instances of not waiting long enough, such as was https://github.com/skupperproject/skupper-router/issues/733

jiridanek:
https://github.com/skupperproject/skupper-router/actions/runs/3164478412/jobs/5152790603#step:29:2990

ganeshmurthy:
Have not seen this in two years, closing

Issue #754, opened 2022-09-30
system_tests_http1_adaptor.Http1AdaptorEdge2EdgeTest::test_01_concurrent_requests timeout


kgiusti:
Looks like some sort of buffer corruption issue:

https://github.com/skupperproject/skupper-router/pull/743#issuecomment-1264009827

ganeshmurthy:
This issue is 18 months old. Closing for now

PR #779, opened 2022-10-11
ISSUE #754: work around possible Q2 egress stall

kgiusti:
See ISSUE #754 (and #758). These failures appear to be cause by a stall in the egress side. The test logs show that the egress HTTP/1 message is unable to completely receive a full stream body data section (incomplete). Debug logging shows that the egress message has Q2 in blocking state, which means the remaining octets of the stream body data cannot be received until Q2 is unblocked. Q2 cannot be unblocked since that can only occur when the entire stream body data has arrived and sent out the raw connection.

This work-around disables Q2 on egress which should prevent the stall. Unfortuantely I cannot reliably reproduce this stall locally so time will tell if this workaround actually alleviates the stall.


kgiusti:
Candidate work around for HTTP/1 issues #754 and #758

PR #796, opened 2022-10-21
Fixes #794: Added unit tests for functions in adaptor_buffer.c

### LINKED ISSUE #794: Add unit tests to tests adaptor_buffer.c functions

ganeshmurthy:


tests/CMakeLists.txt
```diff
@@ -66,6 +66,9 @@ target_link_libraries(clogger ${Proton_LIBRARIES})
 add_executable(threaded_timer_test threaded_timer_test.c)
 target_link_libraries(threaded_timer_test skupper-router)
 
+add_executable(adaptor_buffer_tests adaptor_buffer_test.c)
```
jiridanek, line 69: `adaptor_buffer_tests` does not get picked up by the unittests ci job

https://github.com/skupperproject/skupper-router/blob/45acf8967d6644fd4f029a2df881fc7dfe92fea7/.github/workflows/build.yaml#L355

the list there is used as a regex, so it captures anything that has `unittests` or `unit_test` in its name, or one of the other names later in that line.

Can you please add name of this test on that line, or maybe adjust the line to run everything that are not system tests? Maybe instead of ctest `-R` we can use `-E` and with this, exclude everything named`system_test`?

(Or maybe I can do this when the PR is merged, so that you don't have to fiddle with this now.)

ganeshmurthy: How about this change ?

```
[gmurthy@localhost skupper-router]$ git diff
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index 9e3ed81f..d318dbac 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -67,7 +67,7 @@ add_executable(threaded_timer_test threaded_timer_test.c)
 target_link_libraries(threaded_timer_test skupper-router)
 
 add_executable(adaptor_buffer_tests adaptor_buffer_test.c)
-target_link_libraries(adaptor_buffer_tests skupper-router)
+target_link_libraries(adaptor_buffer_unit_tests skupper-router)
 
 # Bubblewrap is an unprivileged sandboxing tool for Linux. Setting --unshare-net allows
 # running tests in parallel (the ctest -j option) without port clashes
[gmurthy@localhost skupper-router]$ 

```

jiridanek: That will do it, but then it is inconsistently named regarding `threaded_timer_test` and `adaptor_buffer_tests`, which are sort of the same thing as adaptor_buffer tests.

ganeshmurthy: > That will do it, but then it is inconsistently named regarding `threaded_timer_test` and `adaptor_buffer_tests`, which are sort of the same thing as adaptor_buffer tests.

I added adaptor_buffer_test to routerCTestExtraArgs

jiridanek: Additional line coverage was recorded, https://app.codecov.io/github/skupperproject/skupper-router/commit/4108274c17a2427d67b2e567ba8524d01c255bdb

image

Issue #773, opened 2022-10-10
Enable sharing of pn_tls_config_t across TLS sessions


ganeshmurthy:
This is regards to a comment by @cliffjansen - https://github.com/skupperproject/skupper-router/pull/765#pullrequestreview-1134515021

This needs to be addressed so there can be an efficient sharing of  pn_tls_config_t object across TLS sessions.

kgiusti:
fixed on merge of #810

PR #806, opened 2022-10-24
Issue #773: Refactor TCP/HTTP configuration management

kgiusti:
This does not solve ISSUE #773 (TLS domain sharing), but it cleans up a number of bugs in the way the TCP and HTTP adaptors load and validate configuration.

Also modifies the HTTP configuration implementation to be consistent with the current TCP implementation.

### DIFF
--- a/python/skupper_router_internal/management/agent.py
+++ b/python/skupper_router_internal/management/agent.py
@@ -506,7 +506,10 @@ class AllocatorEntity(EntityAdapter):
 
 class HttpListenerEntity(EntityAdapter):
     def create(self):
-        return self._qd.qd_dispatch_configure_http_listener(self._dispatch, self)
+        http_listener = self._qd.qd_dispatch_configure_http_listener(self._dispatch, self)
+        if http_listener is None:
+            raise ValidationError("Invalid httpListener configuration: see logs for details.")
+        return http_listener
 
     def _identifier(self):
         return _host_port_name_identifier(self)
@@ -521,6 +524,8 @@ class HttpListenerEntity(EntityAdapter):
 class TcpListenerEntity(EntityAdapter):
     def create(self):
         config_listener = self._qd.qd_dispatch_configure_tcp_listener(self._dispatch, self)
+        if config_listener is None:
+            raise ValidationError("Invalid tcpListener configuration: see logs for details.")
         return config_listener
 
     def _identifier(self):
@@ -536,6 +541,8 @@ class TcpListenerEntity(EntityAdapter):
 class TcpConnectorEntity(EntityAdapter):
     def create(self):
         config_connector = self._qd.qd_dispatch_configure_tcp_connector(self._dispatch, self)
+        if config_connector is None:
+            raise ValidationError("Invalid tcpConnector configuration: see logs for details.")
         return config_connector
 
     def _identifier(self):
@@ -550,7 +557,10 @@ class TcpConnectorEntity(EntityAdapter):
 
 class HttpConnectorEntity(EntityAdapter):
     def create(self):
-        return self._qd.qd_dispatch_configure_http_connector(self._dispatch, self)
+        http_connector = self._qd.qd_dispatch_configure_http_connector(self._dispatch, self)
+        if http_connector is None:
+            raise ValidationError("Invalid httpConnector configuration: see logs for details.")
+        return http_connector
 
     def _identifier(self):
         return _host_port_name_identifier(self)
--- a/src/adaptors/http1/http1_server.c
+++ b/src/adaptors/http1/http1_server.c
@@ -143,11 +143,10 @@ static void _send_request_message(_server_request_t *hreq);
 // An HttpConnector has been created.  Create an qdr_http_connection_t and a
 // qdr_connection_t for it.
 //
-static qdr_http1_connection_t *_create_server_connection(qd_http_connector_t *ctor,
-                                                         qd_dispatch_t *qd,
-                                                         const qd_http_adaptor_config_t *config)
+static qdr_http1_connection_t *_create_server_connection(qd_http_connector_t *ctor, qd_dispatch_t *qd)
 {
     qdr_http1_connection_t *hconn = new_qdr_http1_connection_t();
+    assert(hconn);
 
     ZERO(hconn);
     hconn->type = HTTP1_CONN_SERVER;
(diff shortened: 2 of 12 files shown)


python/skupper_router_internal/management/agent.py
```diff
@@ -506,7 +506,10 @@ def __str__(self):
 
 class HttpListenerEntity(EntityAdapter):
     def create(self):
-        return self._qd.qd_dispatch_configure_http_listener(self._dispatch, self)
+        http_listener = self._qd.qd_dispatch_configure_http_listener(self._dispatch, self)
+        if http_listener is None:
+            raise ValidationError("Invalid httpListener configuration: see logs for details.")
```
ganeshmurthy, line 511: I was wondering if instead of each entity raising a ValidationError, what if the super class raised a common ValidationError if the create() returned None ?

ganeshmurthy: On an unrelated note, I tried the following - 
```
listener {
    host: 0.0.0.0
    port: amqp
    authenticatePeer: no
    saslMechanisms: ANONYMOUS
    sslProfile: test-ssl-profile
}
```
There is no sslProfile called  ```test-ssl-profile``` in my config file.
The router started without issues which does not seem right. Should be bother to fix it ?

ganeshmurthy: On a related note, I tried this with this PR - 
```
tcpListener {
    host: 0.0.0.0
    port: 2345
    authenticatePeer: yes
    address: test
    sslProfile: test-ssl-profile
}
```
There is no sslProfile called ```test-ssl-profile``` in my config file and the router started anyway. But with the above config, the router fails to start on main branch. That looks like a regression?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -2941,15 +2941,9 @@ void qd_http2_delete_listener(qd_dispatch_t *qd, qd_http_listener_t *li)
...
+qd_http_listener_t *qd_http2_configure_listener(qd_http_listener_t *li, qd_dispatch_t *qd, qd_entity_t *entity)
 {
-    qd_http_listener_t *li = qd_http_listener(qd->server, config);
-    if (!li) {
-        qd_log(http2_adaptor->log_source, QD_LOG_ERROR, "Unable to create http listener: no memory");
-        return 0;
-    }
-
-    li->adaptor_listener = qd_adaptor_listener(qd, config->adaptor_config, http2_adaptor->log_source);
+    li->adaptor_listener = qd_adaptor_listener(qd, li->config->adaptor_config, http2_adaptor->log_source);
```
ganeshmurthy, line 2946: On an *unrelated* note, I was looking into why each adaptor has its own listener like say ```qd_tcp_listener``` and then they share a common listener ```qd_adaptor_listener_t```
All adaptors ultimately call ```qd_adaptor_listener_listen()```
Why can't all adaptors just use ```qd_adaptor_listener_t``` ?
Is that because of the ```protocolVersion``` field in the httpConnector/httpListener ? What if we replaced say httpListener with ```http1Listener``` and ```http2Listener``` and remove the ```protocolVersion``` field ? Is that going to be a problem when tomorrow if `http3` comes along ?

src/adaptors/http1/http1_server.c
```diff
@@ -143,11 +143,10 @@ static void _send_request_message(_server_request_t *hreq);
 // An HttpConnector has been created.  Create an qdr_http_connection_t and a
 // qdr_connection_t for it.
 //
-static qdr_http1_connection_t *_create_server_connection(qd_http_connector_t *ctor,
-                                                         qd_dispatch_t *qd,
-                                                         const qd_http_adaptor_config_t *config)
+static qdr_http1_connection_t *_create_server_connection(qd_http_connector_t *ctor, qd_dispatch_t *qd)
```
ganeshmurthy, line 146: This is minor but can we call this `qd_http_connector_t *` parameter not `ctor` but `connector` instead ?

src/adaptors/http1/http1_server.c
```diff
@@ -218,52 +217,38 @@ static qdr_http1_connection_t *_create_server_connection(qd_http_connector_t *ct
 //
 // Note that this runs on the Management Agent thread, which may be running concurrently with the
 // I/O and timer threads.
-qd_http_connector_t *qd_http1_configure_connector(qd_dispatch_t *qd, qd_http_adaptor_config_t *config, qd_entity_t *entity)
+qd_http_connector_t *qd_http1_configure_connector(qd_http_connector_t *c, qd_dispatch_t *qd, qd_entity_t *entity)
```
ganeshmurthy, line 220: Same here. Instead of `qd_http_connector_t *c` can we have `qd_http_connector_t *connector` ?

tests/system_tests_one_router.py
```diff
@@ -279,6 +279,68 @@ def setUpClass(cls):
         ])
         cls.routers.append(cls.tester.qdrouterd(name, config_9, wait=False, expect=Process.EXIT_FAIL))
 
+        # A tcpConnector must supply an 'address' value
```
ganeshmurthy, line 282: Should we add these same tests but have these configs added via qdmanage and see if those configs get rejected by the router (check if the router returns a `BadRequestStatus`) ?

src/adaptors/http1/http1_server.c
kgiusti: Yes that's a good idea

src/adaptors/http1/http1_server.c
kgiusti: Yup

src/adaptors/http2/http2_adaptor.c
kgiusti: I would keep them separate personally because they are concerned with separate issues.  The qd_adaptor_listener_t isn't a complete listener object - it only manages the lifecycle of the associated pn_listener_t socket.  A qd_(tcp|http)_listener_t has-a qd_adaptor_listener_t, but isn't derived from one.

Possibly the qd_adaptor_listener_t name is misleading since it isn't a representation of a tcp or http listener.  I'm lousy at names, but something involving "proton listener socket lifecycle manager" would be a more accurate description.  Or qd_listener_socket_manager_t maybe....

tests/system_tests_one_router.py
kgiusti: Great minds think alike - I've got some skmanage based tests that do exactly that.  These will be part of my next patch, where the separation of TLS domain and session is done.

It's part of that patch because the error descriptions are changed in that patch for clarification and the new tests check for those error messages.

python/skupper_router_internal/management/agent.py
kgiusti: I think my next patch for this issue will fix that.

With that patch a config file that is missing the referenced sslProfile will prevent the router from starting.

In the case of skmanage the CREATE of the tcpListener will fail: the sslProfile has to be created first (either in the config file or via skmanage).

tests/system_tests_one_router.py
ganeshmurthy: > Great minds think alike 

Um

ganeshmurthy: > 
> It's part of that patch because the error descriptions are changed in that patch for clarification and the new tests check for those error messages.

Sounds good.

ganeshmurthy (approved):

PR #810, opened 2022-10-25
ISSUE-773: refactor TCP and HTTP adaptor TLS configuration sequence

kgiusti:
This patch refactors TLS into two separate classes: a qd_tls_domain_t to represent the TLS domain and a qd_tls_t to managed the TLS session.

This patch also modifies the order of initialization of the TLS domain and session have changed to better handle configuration and management errors.

### DIFF
--- a/src/adaptors/http1/http1_adaptor.c
+++ b/src/adaptors/http1/http1_adaptor.c
@@ -53,7 +53,6 @@
      adaptor for meta-data.
  */
 
-const char *http1_alpn_protocols[HTTP1_NUM_ALPN_PROTOCOLS] = {"http/1.1", "http/1.0"};
 
 ALLOC_DEFINE(qdr_http1_out_data_t);
 ALLOC_DEFINE(qdr_http1_connection_t);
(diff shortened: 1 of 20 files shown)


src/adaptors/http1/http1_adaptor.c
```diff
@@ -53,6 +53,7 @@
      adaptor for meta-data.
  */
 
+const char *http1_alpn_protocols[HTTP1_NUM_ALPN_PROTOCOLS] = {"http/1.1", "http/1.0"};
```
ganeshmurthy, line 56: Since this commit does not implement TLS in the http1 adaptor, can we remove all http1 related files from this commit ?

ganeshmurthy (approved):
This PR LGTM, please commit after removing the http1 stuff. Thanks.

src/adaptors/http1/http1_adaptor.c
kgiusti: That makes sense, will do.

PR #815, opened 2022-10-26
Fixes #778: Added some negative TLS tests to the tcp adaptor

### LINKED ISSUE #778: Add some negative tests for TLS over TCP adaptor and TLS over HTTP2 adaptor
After doing a code coverage review on the TCP TLS and HTTP2 TLS code, it was ascertained that there needs to be some negative tests to increase code coverage. Add test scenarios like using a bad certificate (both sides), bad SSL configuration, handling failure to handshake correctly.
See https://github.com/skupperproject/skupper-router/pull/775#pullrequestreview-1137527449

ganeshmurthy:


tests/system_tests_tcp_adaptor.py
```diff
@@ -1045,11 +1090,16 @@ def run_ncat(self,
                  timeout=10,
                  data=b'abcd',
                  use_ssl=False,
-                 use_client_cert=False):
+                 use_client_cert=False,
```
kgiusti, line 1093: Just my opinion - you don't have to do it:

Why not just pass the actual client cert and key file as parameters rather than passing a bunch of flags and having run_ncat figure out what the certs are based on those flags?

IOW:

def run_ncat(... certfile=None, keyfile=None):...

tests/system_tests_tcp_adaptor.py
```diff
@@ -1045,11 +1090,16 @@ def run_ncat(self,
...
+                 use_bad_ca_cert=False,
```
kgiusti, line 1094: Just my opinion - you don't have to do it:

Why not just pass the actual client cert and key file as parameters rather than passing a bunch of flags and having run_ncat figure out what the certs are based on those flags?

IOW:

def run_ncat(... certfile=None, keyfile=None):...

tests/system_tests_tcp_adaptor.py
```diff
@@ -1045,11 +1090,16 @@ def run_ncat(self,
...
+                 error_ok=False):
```
kgiusti, line 1095: Is this necessary?  Why not let run_ncat raise an exception on failure and have the tests use self.assertRaises() in those cases where run_ncat is expected to fail?

tests/system_tests_tcp_adaptor.py
jiridanek: +1, I thought so too, but I did not want to suggest this as this would require rewriting the tests somewhat

also, `openssl s_client -connect` (https://www.openssl.org/docs/man1.0.2/man1/openssl-s_client.html) is maybe a more powerful tool for testing tls; but I see the point of just using ncat, since there are tests using ncat already (Openssl also have a server command, https://www.openssl.org/docs/man1.0.2/man1/s_server.html)

tests/system_tests_tcp_adaptor.py
jiridanek: I don't like the `raise Exception()`. It would be nicer to create a subclass and throw that. Also, you can create fields on the subclass, so the results are easier to pull out

```python
TestProcessException(Exception):
    def __init__(self, return_code, std_out, std_err):
        self.return_code=return_code
        ...
```

Or maybe create parameter `throw_on_falure=True`, and when that's false, return the result of execution as return value and avoid catching expected exceptions? (I am not fan of `assertRaises()`, although there is nothing wrong with using it, of course.)

tests/system_tests_tcp_adaptor_tls.py
```diff
@@ -30,15 +30,101 @@ def test_authenticate_peer(self):
...
+                         use_ssl=True,
+                         use_client_cert=False,
+                         error_ok=True)
+        # Look for a log line that proves the certificate verification failure.
+        self.INTA.wait_log_message("certificate verify failed")
+        self.logger.log("TCP_TEST Stop %s SUCCESS" % name)
+
+    def test_xxx_present_bad_cert_to_good_listener_good_connector(self):
+        if not ncat_available():
+            self.skipTest("Ncat utility is not available")
```
jiridanek, line 84: This could be something like
```suggestion
    @unittests.skipIf(not ncat_available(), "Ncat utility is not available")
    def test_xxx_present_bad_cert_to_good_listener_good_connector(self):
```

tests/system_tests_tcp_adaptor_tls.py
```diff
@@ -30,15 +30,101 @@ def test_authenticate_peer(self):
...
+        self.INTA.wait_log_message("certificate unknown")
+        self.logger.log("TCP_TEST Stop %s SUCCESS" % name)
+
+    def test_yyy_present_bad_cert_to_good_listener_bad_connector(self):
+        if not ncat_available():
+            self.skipTest("Ncat utility is not available")
+        name = "test_yyy_present_bad_cert_to_good_listener_bad_connector"
+        self.logger.log("TCP_TEST TLS Start %s" % name)
+
+        # Run the ncat command using a bad ca certificate to a tcpListener that is good
```
jiridanek, line 110: I haven't yet looked at full log messages from this, but I want to just ask @ganeshmurthy if you think that https://issues.apache.org/jira/browse/PROTON-1870 is implemented, that is, we get good meaningful errors from TLS failures from Proton

jiridanek (changes requested):
```
E           ERROR: Aborted due to unexpected alloc pool leak of type 'qd_adaptor_buffer_t'
```

tests/system_tests_tcp_adaptor.py
jiridanek: There already is https://docs.python.org/3.6/library/subprocess.html#subprocess.CalledProcessError, maybe just catch that?

ganeshmurthy:
I will open a new PR with the changes requested and with some new tests. Closing this PR.

PR #831, opened 2022-11-03
Fixes #830: Closes the raw connection when the on_close callback is c…

### LINKED ISSUE #830: Deleting a tcp connection via skmanage does not delete the connection
Start a router and its corresponding echo server.
Connect a client to the router which in turn will open an outbound connection to the echo server.
Obtain the identity of the this outbound connection using skmanage
update the adminStatus of the connection to delete using skmanage.
This should delete the connection from the connections table and drop the actual outbound connection that the router initiated to the echo server. This does not happen.

ganeshmurthy:
…alled on the tcp adaptor


ganeshmurthy:
> I'm surprised this function is called at all. This is the callback that's invoked when qdr_process

When the update on the connection's adminStatus is called, the `qdr_close_connection_CT()` in `qdra_connection_update_set_status()` (https://github.com/skupperproject/skupper-router/blob/main/src/router_core/agent_connection.c#L515) is called which in turn sets the `conn->closed = true` and activates the connection which then calls the on_close callback via the `qdr_connection_process()`. The on_close used to do nothing which meant the `qdr_connection_t` object never got removed from the list of connections and the proton raw connection itself is still open. This PR calls the `pn_raw_connection_close()` from the on_close callback.

PR #839, opened 2022-11-04
Fixes #838: Added http2 tests that run over TLS enabled tcp adaptor

### LINKED ISSUE #838: Add a two router test that does HTTP2 over TCP with TLS

ganeshmurthy:


tests/system_test.py
```diff
@@ -748,6 +748,19 @@ def http_addresses(self):
                 ret_val.append("http://%s" % self._cfg_2_host_port(listener))
         return ret_val
 
+    @property
+    def tcp_addresses(self):
+        """Return http://host:port addresses for all tcp listeners"""
```
jiridanek, line 753:
```suggestion
        """Return http(s)://host:port addresses for all tcp listeners"""
```

;P

tests/system_tests_http2_tls.py
```diff
@@ -593,3 +593,101 @@ def test_listener_mgmt_missing_ssl_profile(self):
...
     def test_listener_mgmt_missing_ca_file(self):
         self._test_listener_mgmt_missing_ca_file()
+
+
+class Http2TestTlsOverTcpTwoRouter(Http2TestBase, CommonHttp2Tests, RouterTestSslBase):
+    """
+    In this two router test, the tcpListener is on Router QDR.A and the tcpConnector is on router
+    QDR.B. Both the listener and the connector are encrypted. The server that QDR.B connects to
+    is also encrypted.
+    Client authentication is required for curl to talk to the router QDR.A listener http port.
```
jiridanek, line 603: I don't see how the two test methods below in this class perform what's outlined in this comment.

Ahh, I actually do see it now, the actual tests are inherited into this, and the two test methods below are actually test teardown! Right.

tests/system_test.py
```diff
@@ -748,6 +748,19 @@ def http_addresses(self):
...
+        cfg = self.config.sections('tcpListener')
+        ret_val = []
+        for listener in cfg:
+            require_tls = listener.get("sslProfile")
+            if require_tls is not None:
```
jiridanek, line 758: `require_tls` name sounds like it is a bool, so maybe it should be a bool

```suggestion
            require_tls = listener.get("sslProfile") is not None
            if require_tls:
```

but that is just nitpick on my side; I guess it looks ok to me and I won't find anything substantial looking at it more

tests/system_tests_http2_tls.py
jiridanek: /me feels like I already made a comment like this in the past; advice such like https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html sounds great on paper, but I never saw this work in practice, in a test suite, over time... ;(

kgiusti (approved):

PR #858, opened 2022-11-22
Fixes #845: Set http1 and http2 as supported ALPN protocols on the in…

### LINKED ISSUE #845: TCP Adaptor: Cannot do http2 over tcp/tls
http2/grpc require ALPN over TLS. tcpConnector and tcpListener do not do ALPN at the moment. We will have to add a feature to the tcp adaptor so there is a way to to detect ALPN. This could be achieved by introducing an `alpnProtocols` field in the tcpConnector and tcpListener or perhaps through other techniques

ganeshmurthy:
…gress side. ALPN Protocol negotiated

on the ingress side will be presented on the egress side as the desired ALPN protocol.

### DIFF
--- a/src/adaptors/adaptor_tls.c
+++ b/src/adaptors/adaptor_tls.c
@@ -120,6 +120,13 @@ qd_tls_t *qd_tls(qd_tls_domain_t *tls_domain, void *context, uint64_t conn_id, q
     return tls;
 }
 
+int qd_tls_set_alpn_protocols(qd_tls_domain_t *tls_domain, const char *alpn_protocols[], int alpn_protocol_count)
+{
+    assert(alpn_protocol_count > 0);
+    assert(tls_domain);
+    return pn_tls_config_set_alpn_protocols(tls_domain->pn_tls_config, alpn_protocols, alpn_protocol_count);
+}
+
 qd_tls_domain_t *qd_tls_domain(const qd_adaptor_config_t *config,
                                const qd_dispatch_t       *qd,
                                qd_log_source_t           *log_source,
@@ -265,7 +272,7 @@ qd_tls_domain_t *qd_tls_domain(const qd_adaptor_config_t *config,
         // https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.txt
         //
         if (alpn_protocols) {
-            res = pn_tls_config_set_alpn_protocols(tls_domain->pn_tls_config, alpn_protocols, alpn_protocol_count);
+            res = qd_tls_set_alpn_protocols(tls_domain, alpn_protocols, alpn_protocol_count);
             if (res != 0) {
                 qd_log(log_source,
                        QD_LOG_ERROR,
--- a/src/adaptors/adaptor_tls.c
+++ b/src/adaptors/adaptor_tls.c
@@ -62,6 +62,22 @@ static void take_back_encrypt_input_buffs(qd_tls_t *tls);
 static void take_back_decrypt_output_buffs(qd_tls_t *tls);
 static void take_back_encrypt_output_buffs(qd_tls_t *tls);
 
+void qd_tls_get_alpn_protocol(qd_tls_t *tls, char **alpn_protocol)
+{
+    const char *protocol_name;
+    size_t      protocol_name_length;
+    if (pn_tls_get_alpn_protocol(qd_tls_get_pn_tls_session(tls), &protocol_name, &protocol_name_length)) {
+        //
+        // An ALPN protocol was present. Get the ALPN protocol into the passed in alpn_protocol
+        //
+        *alpn_protocol = qd_calloc(protocol_name_length + 1, sizeof(char));
+        memmove(*alpn_protocol, protocol_name, protocol_name_length);
+    }
+    else {
+        *alpn_protocol = 0;
+    }
+}
+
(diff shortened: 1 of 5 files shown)


src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -159,6 +165,21 @@ static void qdr_tcp_connection_ingress_accept(qdr_tcp_connection_t *tc);
 static void handle_outgoing(qdr_tcp_connection_t *conn);
 static void encrypt_outgoing_tls(qdr_tcp_connection_t *conn, qd_adaptor_buffer_t *unencrypted_buff, bool write_buffers);
 
+static void get_alpn_protocol(qdr_tcp_connection_t *conn, char **alpn_protocol)
```
kgiusti, line 168: Can this logic be provided by a new API call to adaptor_tls rather than a tcp-only internal function? Seems like having a general "qd_tls_get_aplpn_protocol()" would be a useful.

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -474,10 +495,24 @@ static int handle_incoming(qdr_tcp_connection_t *conn, const char *msg)
         // Add the application properties
         //
         props = qd_compose(QD_PERFORMATIVE_APPLICATION_PROPERTIES, props);
-        qd_compose_start_map(props);
+        qd_compose_start_map(props);  // start map
         qd_compose_insert_symbol(props, QD_AP_FLOW_ID);
         vflow_serialize_identity(conn->vflow, props);
-        qd_compose_end_map(props);
+        if (conn->require_tls) {
```
kgiusti, line 501: Can't the negotiated ALPN be retrieved once during the "on_secure" callback (storing it in the tcp connection structure)?  Then you can simply check if "conn->negotiated_alpn" (or whatever name) is set and then insert it in the map? This function is pretty long already.

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -1344,6 +1379,35 @@ static void qdr_tcp_create_server_side_connection(qdr_tcp_connection_t* tc)
     qdr_link_set_context(tc->outgoing_link, tc);
 }
 
+static void set_alpn_protocol(qdr_tcp_connection_t *conn, qd_message_t *message, char **alpn_protocol)
```
kgiusti, line 1382: See comment in create_egress_connection()

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -1369,9 +1433,39 @@ static bool qdr_tcp_create_egress_connection(qd_tcp_connector_t      *connector,
     tc->is_egress_dispatcher_conn = false;
 
     tc->require_tls = !!connector->tls_domain;
+    qd_message_t *msg = qdr_delivery_message(initial_delivery);
     if (tc->require_tls) {
         // Since TLS is required on this connection, try to initialize TLS attributes
         // from the associated sslProfile.
+        char *alpn_protocol;
+        set_alpn_protocol(tc, msg, &alpn_protocol);
```
kgiusti, line 1441: Something to consider: walking the message's application properties map is _very_ expensive (remember old router message annotations?).

The problem here is that we're doing it *twice* this this change.

See qdr_associate_vflow_flows() - it is done there as well.

I would recommend creating a utility routine that given the message's QD_FIELD_APPLICATION_PROPERTIES extracts & processes all the interesting stuff *in one pass*.  It looks like (currently) all that is needed is the vflow id and the alpn protocol (if tls is in use) for now.

kgiusti (changes requested):
See comments.  Can you also get Cliff to review?  thanks.

src/adaptors/tcp/tcp_adaptor.c
kgiusti: or better yet "qd_tls_get_alpn_protocol()"

ganeshmurthy: Done ! thanks

src/adaptors/tcp/tcp_adaptor.c
ganeshmurthy: Done !

src/adaptors/tcp/tcp_adaptor.c
ganeshmurthy: Done !

ganeshmurthy:
I have addressed your comments, please take a look one more time, thanks.

src/adaptors/adaptor_tls.c
```diff
@@ -62,6 +62,19 @@ static void take_back_encrypt_input_buffs(qd_tls_t *tls);
...
+{
+    const char *protocol_name;
+    size_t      protocol_name_length;
+    if (pn_tls_get_alpn_protocol(qd_tls_get_pn_tls_session(tls), &protocol_name, &protocol_name_length)) {
+        //
+        // An ALPN protocol was present. Get the ALPN protocol into the passed in alpn_protocol
+        //
+        *alpn_protocol = qd_calloc(protocol_name_length + 1, sizeof(char));
+        memmove(*alpn_protocol, protocol_name, protocol_name_length);
+    }
```
kgiusti, line 75: else set *alpn_protocol = 0;

kgiusti (approved):

PR #873, opened 2022-12-07
Fixes #872: Added system tests that use openssl s_client and s_server

### LINKED ISSUE #872: Add system tests that use openssl s_client and openssl s_server

ganeshmurthy:


tests/system_test.py
```diff
@@ -399,6 +407,19 @@ def error(msg):
             error("expected %s but actual returncode is %s"
                   % (self.expect, self.returncode))
 
+    def wait_out_message(self, pattern, outfile_path=None, **retry_kwargs):
```
kgiusti, line 410: Can you make this a common method and change the wait_log_message() method in the Qdrouterd class to call it?

That would eliminate duplicate implementations of that file search logic - thanks.

ganeshmurthy: Good point. Done ! Please check and approve !

PR #886, opened 2022-12-15
Fixes #885 - Improve vFlow tests; fix issues with cross-network access to router vFlow agents

### LINKED ISSUE #885: vFlow flushes are not received on routers with no mobile destinations.
An interior router that has no mobile destinations (adaptor-connectors, mobile-address consumers) will not propagate its in-process destination for vFlow commands to other interior routers.  This will cause the router to not flush its vFlow records to a newly connected collector.

h/t @ajssmith for discovering this issue.

ted-ross:
There were two issues with propagated indication of the presence of an in-process destination for an address.  One affected interior-interior forwarding and the other affected edge-interior forwarding.  This caused FLUSH commands from a remote vFlow collector to not arrive at the intended event source.

There were a number of defects in the existing vFLow tests that prevented the detection of the above issue.  The tests are also fixed in this PR.

### DIFF
--- a/src/router_core/modules/edge_addr_tracking/edge_addr_tracking.c
+++ b/src/router_core/modules/edge_addr_tracking/edge_addr_tracking.c
@@ -205,6 +205,13 @@ static bool qdrc_can_send_address(qdr_address_t *addr, qdr_connection_t *in_conn
         return true;
     }
 
+    //
+    // If there is a propagated in-process destination for this address, sending is allowed.
+    //
+    if (addr->propagate_local) {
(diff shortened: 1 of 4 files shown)


ted-ross:
@ajssmith (the original finder of this defect) has tested the PR and verified that the vFlow problems are solved.

ganeshmurthy:
@ted-ross this looks ok to me and since @ajssmith has tested already, please feel free to commit

ganeshmurthy (approved):

PR #893, opened 2023-01-06
Fixes #891: Added tests for the expandvars.py script.

### LINKED ISSUE #891: Add a system test for scripts/expandvars.py.
The scripts/expandvars.py will be used later in nginx based testing

ganeshmurthy:


tests/system_tests_expandvars.py
```diff
@@ -0,0 +1,91 @@
...
+        port_expanded = False
+        name_expanded = False
+        with open(output_file) as ofile:
+            for line in enumerate(ofile):
+                test_line = line[1].strip()
+                if "PORT=12345" == test_line:
+                    port_expanded = True
+                if "NAME=YoYoMa" == test_line:
+                    name_expanded = True
+        self.assertTrue(port_expanded and name_expanded)
```
jiridanek, line 57: This is way too circumspect approach, I feel.

First, `for line in enumerate(ofile):` should better be `for line_number, line in enumerate(ofile, start=1):` due to how https://docs.python.org/3/library/functions.html#enumerate works.

What about something like this (I did not run it, just an idea)?

```
with open(output_file) as ofile:
    lines = ofile.readlines()
self.assertIn("PORT=12345\n",  lines)
self.assertIn("NAME=YoYoMa\n",  lines)
```

maybe get rid of the \n's (if they are present, I think they are), for that, `lines = ofile.read().splitlines()` should work.

tests/system_tests_expandvars.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_expandvars_input_file_without_output_file(self):
+        # Provide an input file and NO output file to the expandvars.py script
+        # and make sure the input file is overwritten/populated with the expanded vars.
```
jiridanek, line 61: Method doc comments should be in """ """.

tests/system_tests_expandvars.py
```diff
@@ -0,0 +1,91 @@
...
+from subprocess import STDOUT, CalledProcessError
+import shutil
+from system_test import TestCase
+
+env_config = {
+    'PORT': '12345',
+    'NAME': 'YoYoMa'
+}
+
+current_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)))
```
jiridanek, line 31: It should be possible to use `import pathlib`, it is present since Python 3.4 and it is somewhat nicer to use than `os.path`. But that is matter of preference, of course. https://docs.python.org/3/library/pathlib.html

tests/system_tests_expandvars.py
```diff
@@ -0,0 +1,91 @@
...
+        # Does not pass any arguments to the expandvars.py
+        # and makes sure that the script fails.
+        args = [sys.executable, os.path.join(expandvars_script_folder, script_file)]
+        script_failed = False
+        try:
+            subprocess.run(args, stderr=STDOUT, check=True, env=env)
+        except CalledProcessError as e:
+            if "returned non-zero exit status 2" in str(e):
+                script_failed = True
+        self.assertTrue(script_failed)
```
jiridanek, line 91: This could've been done with https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

```
with self.assertRaises(CalledProcessError) as cm:
    subprocess.run(args, stderr=STDOUT, check=True, env=env)
self.assertIn("returned non-zero exit status 2", cm.exception)
```

or, maybe even better, `self.assertEquals(2, cm.exception.returncode)`, https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError

tests/system_tests_expandvars.py
```diff
@@ -0,0 +1,91 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+import os
+import sys
+import subprocess
```
jiridanek, line 21: Working like this will make it hard to report test coverage, if we ever wanted to do that... https://coverage.readthedocs.io/en/7.0.2/subprocess.html

Alternative would be to write the minimum number of subprocess tests we need (which is two, for wrong args and for correct args) and then a test that imports `expandvars.py` as a module and calls `expand_file(input_filename, output_filename)` directly. But, that is not worth the effort, probably. And only testing `expand_file` would mean we would not test running the script, which is also important to test.

tests/system_tests_expandvars.py
ganeshmurthy: The reason I did not put this in doc comments is because when the test runs, it displays the first line of the doc comment when running the test. I would prefer to see the test name only when the test is being run. If the test fails and I want to look into what the test is doing, I would read the doc comment to help me.

tests/system_tests_expandvars.py
ganeshmurthy: I looked into pathlib and to me the os.path approach looks to be more simpler.

tests/system_tests_expandvars.py
ganeshmurthy: Done !

tests/system_tests_expandvars.py
ganeshmurthy: ok, it does look like the test name is displayed at the end of the line. I will fix this. Thx

jiridanek: That's this behavior, right? https://stackoverflow.com/questions/12962772/how-to-stop-python-unittest-from-printing-test-docstring

Since we are already subclassing `unittest.TestCase` (as `system_tests.TestCase`), we may implement the workaround from the SO there, maybe it will really work

```
    def shortDescription(self):
        return None
```

tests/system_tests_expandvars.py
jiridanek: ok, np

tests/system_tests_expandvars.py
ganeshmurthy: I removed the enumeration but I am still looping for simplicity

jiridanek: The simplest and most direct assertion would be to assert on the file content as a string, I'd say, no looping necessary.

```
with open(output_file) as ofile:
    self.assertEquals("PORT=12345\nNAME=YoYoMa\n", ofile.read())
```

This might have a problem with platform dependent line-ending characters, but I _believe_ it won't be problem here.

ganeshmurthy: I would like to avoid using "\n" in the code. I think what I have now is pretty simple to follow. Thx

ganeshmurthy:
@jiridanek if you don't have any other concerns about this PR, I would like to go ahead and merge it to main. I am working on including nginx in CI and would like to use this script in that project. Thx

jiridanek (approved):

tests/system_tests_expandvars.py
jiridanek: It is not too bad, but I still like my version better ;)

PR approved!

tests/system_tests_expandvars.py
jiridanek: Asserting on the exception text is weird and imo pointless. It would've been better to only assert on exception class and errorcode. Bit it's no big deal either way.

PR #906, opened 2023-01-18
Fixes #857: Prelimnary introduction of nginx into CI testing

### LINKED ISSUE #857: Test enhancement: Integrate nginx into CI for HTTP/x testing
Most of the HTTP/1 tests use the HTTPServer and BaseHTTPRequestHandler from the python library as the test server. This should be replaced with nginx wherever possible:

1. nginx is a very popular http server that is production grade and would make a better verification tool than the python server. In fact the python docs for the server do NOT recommend using the python server in production!
2. currently the python server requires a boatload of python hacking to work around its deficiencies and makes me a Sad Panda because it seems that every time I need to update/fix a test involving that python http server I invite it to vomit up a swirling torrent of pain points that are apparently designed to make my life as miserable as possible [end of rant :P ].

Exactly how nginx can be integrated into CI (and run in GHA) is TBD: containers? Non-root user? etc?

ganeshmurthy:


tests/system_test.py
```diff
@@ -488,6 +493,29 @@ def wait_ports(self, **retry_kwargs):
         wait_ports(self.ports_family, **retry_kwargs)
 
 
+class NginxServer(Process):
+    """
+    Sets up an ngix server
+    """
+    def __init__(self,
```
kgiusti, line 500: Not strictly required, but can you add python type-hints to the arguments?

tests/ssl_certs/chained.pem
```diff
@@ -0,0 +1,43 @@
+-----BEGIN CERTIFICATE-----
```
kgiusti, line 1: Can you updated the gencerts.sh file to show how this certificate was generated in case we need to reproduce a new one in the future.

kgiusti (approved):

.github/workflows/build.yaml
```diff
@@ -138,7 +138,7 @@ jobs:
 
       - name: Install Linux build dependencies
         run: |
-          sudo apt update; sudo apt install -y libdw-dev swig libpython3-dev libsasl2-dev libjsoncpp-dev libwebsockets-dev libnghttp2-dev ccache ninja-build pixz libbenchmark-dev
+          sudo apt update; sudo apt install -y libdw-dev swig libpython3-dev libsasl2-dev libjsoncpp-dev libwebsockets-dev libnghttp2-dev ccache ninja-build pixz libbenchmark-dev nginx
```
jiridanek, line 141: so, it's not going to be a testcontainer after all? all things considered, I probably agree with not using test containers for this; it would add extra complications and there is not all that much benefit, since nginx is easy to install on Linux

tests/system_test.py
```diff
@@ -488,6 +493,29 @@ def wait_ports(self, **retry_kwargs):
...
+                 config_path: str,  # Full path of templated config file (string) /blah-blah/nginx/nginx-template.conf
+                 env: dict[str, str],  # A dict of env variables which will be substituted in the config_path file.
+                 name: str="nginx",
+                 expect: int=Process.RUNNING,
```
jiridanek, line 504: When the type is annotated, there is supposed to be spaces around the equals sign, https://peps.python.org/pep-3107/#syntax

```
def foo(a: expression, b: expression = 5):
    ...
```

I guess the check is disabled in the python checker ci job because there are already places that break these formatting conventions

tests/system_test.py
```diff
@@ -488,6 +493,29 @@ def wait_ports(self, **retry_kwargs):
...
+        config_file = os.path.join(env['setupclass-folder'], 'nginx.conf')
+
+        # Run the expandvars.py script to replace the shell style variables in the config file with
+        # environment variables.
+        args = [sys.executable, os.path.join(expandvars_script_folder, script_file), input_file, config_file]
+        subprocess.run(args, stderr=STDOUT, check=True, env=env)
+
+        # Pass the config file for nginx to use.
+        nginx_cmd = ['nginx', '-c', config_file]
+        super(NginxServer, self).__init__(nginx_cmd, name=name, expect=expect, **kwargs)
```
jiridanek, line 516: Just an observation, in python 3, you don't have to provide name of current class in super()

```
class C(B):
    def method(self, arg):
        super().method(arg)    # This does the same thing as:
                               # super(C, self).method(arg)
```

https://docs.python.org/3/library/functions.html#super

tests/system_tests_http2_tls.py
```diff
@@ -711,3 +711,103 @@ def test_zzz_http_connector_delete(self):
...
+                            'certFile': cls.ssl_file('client-certificate.pem'),
+                            'privateKeyFile': cls.ssl_file('client-private-key.pem'),
+                            'password': 'client-password'}),
+        ])
+        cls.router_qdra = cls.tester.qdrouterd("http2-two-router-tls-A", config_qdra, wait=True)
+        cls.router_qdrb = cls.tester.qdrouterd("http2-two-router-tls-B", config_qdrb)
+        cls.router_qdra.wait_router_connected('QDR.B')
+        cls.router_qdrb.wait_router_connected('QDR.A')
+        cls.router_qdrb.wait_http_server_connected(is_tls=True)
+        wait_http_listeners_up(cls.router_qdra.addresses[0])
```
jiridanek, line 799: this feels like a huge amount of setup code given that the test can be summarized with one line of comment; just a thought, I am not sure what to do about it, and there's probably nothing to do about it right now

tests/system_tests_tcp_adaptor_tls.py
```diff
@@ -610,3 +613,114 @@ def test_connector_bad_server_cert(self):
...
+        address = get_address(self.router_qdra) + "/images" + image_file_name
+        out = run_curl(args=[address, '--verbose'] + self.curl_args +
+                            ['--output', self.router_qdra.outdir + image_file_name] +
+                            ['--http2', '--http2-prior-knowledge'])
+        digest_of_server_file = get_digest(image_file(image_file(image_file_name[1:])))
+        digest_of_response_file = get_digest(self.router_qdra.outdir + image_file_name)
+        self.assertEqual(digest_of_server_file, digest_of_response_file)
+
+    def test_get_image_jpg_http11(self):
+        image_file_name = '/test.jpg'
```
jiridanek, line 720: Is the trex image free to use without citing source? ;P

jiridanek (approved):
Looks to me much like any other systemtest in skupper-router, the only difference being that nginx needs its own textfile with config generated!

tests/system_test.py
```diff
@@ -488,6 +493,29 @@ def wait_ports(self, **retry_kwargs):
         wait_ports(self.ports_family, **retry_kwargs)
 
 
+class NginxServer(Process):
+    """
+    Sets up an ngix server
+    """
+    def __init__(self,
+                 config_path: str,  # Full path of templated config file (string) /blah-blah/nginx/nginx-template.conf
+                 env: dict[str, str],  # A dict of env variables which will be substituted in the config_path file.
```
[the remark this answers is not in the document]
jiridanek: Actually, Python checker in CI is now a noop

```
No tests were found!!!
```

https://github.com/skupperproject/skupper-router/actions/runs/4009925749/jobs/6885845691#step:9:25

.github/workflows/build.yaml
ganeshmurthy: I looked into doing container nginx but this approach seemed like the easiest way to get nginx into our CI system, so I went with it.

tests/system_test.py
ganeshmurthy: Done !

tests/system_test.py
ganeshmurthy: Done !

tests/system_test.py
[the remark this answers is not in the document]
ganeshmurthy: > The type hint here is Python 3.9+, in prior versions you are supposed to do `Dict[str, str]`, https://docs.python.org/3.10/library/typing.html#typing.Dict
> 
> This does not matter for CI, apparently, because we don't run a Python typechecker, and if we wanted to run a type checker, we can always use Python 3.9+ for the type checking

Modified dict to typing.Dict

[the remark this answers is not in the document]
ganeshmurthy: > Actually, Python checker in CI is now a noop
> 
> ```
> No tests were found!!!
> ```
> 
> https://github.com/skupperproject/skupper-router/actions/runs/4009925749/jobs/6885845691#step:9:25

So the standalone Python checker job is not doing anything now ?

tests/system_tests_tcp_adaptor_tls.py
ganeshmurthy: Yes, I got it from a free to use copyright free website

ganeshmurthy:
> Looks to me much like any other systemtest in skupper-router, the only difference being that nginx needs its own textfile with config generated!

My goal is to ultimately use nginx for all tests and get rid of the quart/h2 server

tests/system_test.py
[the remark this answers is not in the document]
jiridanek: > Modified dict to typing.Dict

I thought we are fine with Python 3.9+. And, I forgot, it might be actually problematic in lower python even if not typechecking because annotations are evaluated eagerly, There is `from __future__ import annotations` that is 3.7+  that optionally enables postponed evaluation of annotations https://peps.python.org/pep-0563/

```
$ python3.6

>>> dict[str,str]
TypeError: 'type' object is not subscriptable
```

> So the standalone Python checker job is not doing anything now ?

Well, you saw it for yourself, no? `No tests were found!!!`

ganeshmurthy:
@jiridanek is there any other changes, you want made in this PR? If not, can you please approve it and I will push it to main branch. Thanks.

jiridanek:
> @jiridanek is there any other changes, you want made in this PR? If not, can you please approve it and I will push it to main branch. Thanks.

already approved 5 days ago

ganeshmurthy:
> > @jiridanek is there any other changes, you want made in this PR? If not, can you please approve it and I will push it to main branch. Thanks.
> 
> already approved 5 days ago

My bad. Sorry. Thanks much.

PR #912, opened 2023-01-20
Workaround Containerfile annobin build issue BZ#2162746

jiridanek:


Containerfile
```diff
@@ -23,7 +23,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
     rpm-build \
     gcc gcc-c++ make cmake \
     cyrus-sasl-devel openssl-devel libuuid-devel \
-    python3-devel swig \
+    python39-devel swig \
```
ganeshmurthy, line 26: I am guessing you have to call out python39 because ubi8 comes with python3.6 and we need python39 because proton (main branch) requires it.

jiridanek: We could use python39 for proton python only and let the router continue using python3 (python 3.6), but I thought it is less error-prone to move everything to python39. But that is a matter of making a decision. As far as I can tell, the debate yesterday on this was not conclusive either way.

jiridanek: Btw, I did not find Python 3.11 in ubi8 yet, but it seems to be planned https://bugzilla.redhat.com/show_bug.cgi?id=2137139, https://bugzilla.redhat.com/show_bug.cgi?id=2127923

ganeshmurthy: I hate the whole proton-not-supporting-python3.6 thing. I really do. 

If we can use python3-devel in the Containerfile without any issues that would be great since we don't have to go and keep updating the versions of python in the Containerfile at a later point but if using python3-devel is a problem we should move to python39-devel (sucks)

ganeshmurthy (approved):

Containerfile
jiridanek: python39 in rhel 8 is supported until May 2024, https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle, before that the version need not be raised (unless Proton decides to, I guess ;)

ganeshmurthy: ok. On an related note, I was wondering if we should use CI to start using CentOS stream 9 and also move to a newer version of Ubuntu

jiridanek: I think we should stick with centos 8, simply because the router is still to be supported on rhel 8 (in addition to 9) and for compatibility reasons it makes sense to have the oldest supported version in CI

Regarding ubuntu, I am not sure; probably better to update, although the current ubuntu can have python 3.9 just fine; just change version here

https://github.com/skupperproject/skupper-router/blob/349fce8f388755fa88d3c3335c52a690e25cba94/.github/workflows/build.yaml#L130-L134

jiridanek: it could make sense to have both centos/rhel 8 and 9 in some ci somewhere, but probably not here in github

PR #914, opened 2023-01-23
Issue 676 http1 tls

kgiusti:


src/adaptors/adaptor_tls.c
```diff
@@ -723,7 +729,289 @@ bool qd_tls_has_output(const qd_tls_t *tls)
...
+        // push any unencrypted output data from the protocol adaptor into TLS for encryption.
+
+        if (pn_tls_is_secure(tls->tls_session)) {
+            assert(take_output_cb);
+            capacity = pn_tls_get_encrypt_input_buffer_capacity(tls->tls_session);
+
+            if (capacity > 0) {
+                qd_adaptor_buffer_list_t ubufs = DEQ_EMPTY;
+
+                int64_t out_octets = take_output_cb(take_output_context, &ubufs, capacity);
```
kgiusti, line 809: octets are not necessary.

src/adaptors/adaptor_tls.c
```diff
@@ -723,7 +729,289 @@ bool qd_tls_has_output(const qd_tls_t *tls)
...
+                        qd_log(tls->log_source, QD_LOG_TRACE, "[C%" PRIu64 "] EOS signalled: closing TLS output",
+                               tls->conn_id);
+                    }
+                }
+            }
+        }
+
+        // give empty output buffers to be filled with decrypted data
+
+        capacity = pn_tls_get_decrypt_output_buffer_capacity(tls->tls_session);
```
kgiusti, line 840: move up to other output capacity fill

src/adaptors/adaptor_tls.c
```diff
@@ -723,7 +729,289 @@ bool qd_tls_has_output(const qd_tls_t *tls)
...
+                    assert(DEQ_IS_EMPTY(ubufs));
+                    if (out_octets == QD_TLS_EOS && tls->output_eos == false) {
+                        pn_tls_close_output(tls->tls_session);
+                        tls->output_eos = true;
+                        qd_log(tls->log_source, QD_LOG_TRACE, "[C%" PRIu64 "] EOS signalled: closing TLS output",
+                               tls->conn_id);
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ssorj: The doc for pn_tls_is_secure sugests that it will also return false on some kind of error.  Just checking that that is acounted for below where you inspect tls_error.

tests/system_tests_http1_adaptor.py
```diff
@@ -23,67 +23,63 @@
...
 from http1_tests import Http1ClientCloseTestsMixIn
 from http1_tests import Http1CurlTestsMixIn
-from http1_tests import wait_http_listeners_up
+from http1_tests import wait_http_listeners_up, wait_http_listeners_down
 from http1_tests import HttpAdaptorListenerConnectTestBase
 from http1_tests import HttpTlsBadConfigTestsBase
+from http1_tests import http1_simple_request
+from system_tests_ssl import RouterTestSslBase
 
+CA_CERT = RouterTestSslBase.ssl_file('ca-certificate.pem')
```
ganeshmurthy, line 55: Can we move these constants to system_test.py ? system_tests_tcp_adaptor.py also defines them (https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_tcp_adaptor.py#L71), so they are now defined in two places.

Once these constants are moved to system_test.py, other tests can be modified to use these constants.

src/adaptors/adaptor_tls.c
ganeshmurthy: Move this loop that call `pn_tls_give_decrypt_output_buffers()` before the call to `if (pn_tls_is_secure(tls->tls_session))`
Perhaps even possible add a static function that contains the loop that calls `pn_tls_give_encrypt_output_buffers()` and the loop that contains calls to `pn_tls_give_decrypt_output_buffers()`
and call that static function before the call to `if (pn_tls_is_secure(tls->tls_session))`

src/alloc_pool.c
```diff
@@ -108,6 +110,11 @@ static const char *leaking_types[] = {
     "qd_parsed_field_t",   // DISPATCH-1701
     "qdr_delivery_ref_t",  // DISPATCH-1702
 
+#if PN_VERSION_MAJOR == 0 && PN_VERSION_MINOR == 38
```
ganeshmurthy, line 113: This might mask a leak that @astitcher has been seeing with his proton changes.

src/adaptors/adaptor_tls.c
kgiusti: Out of date: out_octets are used in the trace log output. Very helpful for debug.

src/adaptors/http1/http1_client.c
```diff
@@ -370,19 +420,22 @@ static void _replenish_empty_read_buffers(qdr_http1_connection_t *hconn)
...
+static int64_t _take_output_data(void *context, qd_adaptor_buffer_list_t *abufs, size_t limit)
 {
-    uint64_t                total_octets = 0;
+    int64_t                 total_octets = 0;
     qdr_http1_connection_t *hconn        = (qdr_http1_connection_t *) context;
 
     assert(hconn);
     assert(DEQ_IS_EMPTY(*abufs));
 
+    if (hconn->output_closed)
```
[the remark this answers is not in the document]
kgiusti: Yeah I agree. I'm lousy at naming things - we need something more generic. How about QD_ADAPTOR_EOS?  or QD_IO_EOS?  I'm open to ideas...

src/adaptors/http1/http1_server.c
```diff
@@ -674,170 +948,93 @@ static void _handle_connection_events(pn_event_t *e, qd_server_t *qd_server, voi
...
+    case PN_RAW_CONNECTION_CLOSED_READ: {
+        hconn->q2_blocked = false;
+        // handled in I/O loop
+        break;
+    }
+    case PN_RAW_CONNECTION_NEED_WRITE_BUFFERS: {
+        // handled in I/O loop
         break;
     }
     case PN_RAW_CONNECTION_WRITTEN: {
```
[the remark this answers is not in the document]
kgiusti: I think this is already accounted for in the latest patches. http1_server/client.c print every proton event passed to the handler. adaptor_tls.c now logs every octet read/written to both the TLS layer and the raw connection.  For example: 

https://github.com/kgiusti/skupper-router/blob/ISSUE-676-HTTP1-TLS/src/adaptors/adaptor_tls.c#L913

logs how many encrypted octets were taken from TLS and written to the raw connection

src/alloc_pool.c
kgiusti: This is an unfortunate change, agreed. But the I/O layer uncovers a definite leak in the TLS layer in release 0.38.0.

Removing this conditional will cause all the http1 CI tests to fail when built with proton 0.38.

jiridanek: If there is a plan to build router container image with patched version of qpid proton 0.38, then it might make sense to create a 0.38.1 branch in proton, put the patches there, and run router CI against that, and not the released 0.38... maybe...

src/adaptors/http1/http1_server.c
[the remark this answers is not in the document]
kgiusti: @ganeshmurthy 
Here's actual log output. For unencrypted data flowing from the adaptor thru TLS to the raw connection:

`[C7] 257 unencrypted octets taken by TLS for encryption (1 buffers)`
`[C7] 279 encrypted octets written to the raw connection by TLS (1 buffers)`
`[C7] client proactor event PN_RAW_CONNECTION_WRITTEN`

And here's the reverse: encrypted coming in from the raw connection, going through TLS and up to the adaptor:

`[C7] client proactor event PN_RAW_CONNECTION_READ`
`[C7] 111 encrypted octets read from the raw connection passed to TLS for decryption (1 buffers)`
`[C7] 89 decrypted octets taken from TLS for adaptor input (1 buffers)`

Does that give us enough information on the octet flows?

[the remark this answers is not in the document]
ganeshmurthy: This looks good to me. Just a minor comment. In the `PN_RAW_CONNECTION_WRITTEN` event case, the log messages are written prior to the logging of the actual event and in the case of `PN_RAW_CONNECTION_READ` the log message are written after the logging of the actual event. Can we fix the `PN_RAW_CONNECTION_READ` event log message to appear *after* the actual octet counts so it looks uniform? Also change octets to bytes so it is easy for a pedestrian person looking at the logs to understand ?

src/alloc_pool.c
ganeshmurthy: I see that this is the only way to have the tests not fail with proton 0.38.0. I am ok with this change.

tests/system_tests_http1_adaptor.py
kgiusti: I just tried to do that cleanup, and it's way bigger than it looks. First of all, you cannot put them into system_test.py because that causes python to barf on a cyclic include. We'd have to export them from system_tests_ssl.py instead, but getting that file to consistently use the new path variable is a huge change, nevermind the changes to the other TLS tests.

Sounds like a good idea, but this patch is already really big and I'd rather avoid even more changes if it can be helped. This should be done in a follow up change.

src/adaptors/http1/http1_client.c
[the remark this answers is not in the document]
ganeshmurthy: QD_IO_EOS sounds good. thanks.

tests/system_tests_http1_adaptor.py
ganeshmurthy: ok, let's do that in a subsequent commit and keep this commit only to the relevant files.

src/adaptors/adaptor_tls.c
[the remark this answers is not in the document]
kgiusti: @ssorj see the latest commit on this PR. Added some comments to explain the "not secure" handling.  Lemme know what you think.

ganeshmurthy (approved):
The changes look good to me. It should be merged to main and see how it goes.

tests/system_tests_http1_adaptor.py
kgiusti: https://github.com/skupperproject/skupper-router/issues/958

src/adaptors/http1/http1_server.c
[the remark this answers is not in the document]
kgiusti: Changing the wording to bytes is fine.

I wouldn't recommend changing the order of the logs.

That log sequence mirrors exactly the sequence of events as they are processed by the code: you're going to get a READ event _first_, because that kicks off the I/O loop which (may) result in reading bytes from the raw connection. So the READ event will always come before any READ-related I/O is done. The reverse is true for WRITTEN: proactor won't issue a WRITTEN event until after the adaptor writes bytes to the raw connection and returns control back to proactor for it to do the socket write. The WRITTEN even actually posts later, after the socket write returns.

Having the trace log reorder events from their actual flow is going to misleading.

Issue #925, opened 2023-01-31
Test for the Containerfile does not attempt to use skstat/skmanage, misses issues with Proton Python install


jiridanek:
```
$ docker run --name router -itd --rm quay.io/skupper/skupper-router:main
196e2784d48f4e597cf0057e9daca88d6a88c375d75f62fb98959e935308b6db
$ docker exec router skstat -l
Traceback (most recent call last):
  File "/usr/bin/skstat", line 29, in <module>
    from skupper_router.management.client import Node
  File "/usr/lib/python3.9/site-packages/skupper_router/management/client.py", line 22, in <module>
    import proton
ModuleNotFoundError: No module named 'proton'
```

Reported by @fgiorgetti

ganeshmurthy:
@jiridanek this issue seems to have been fixed by you already. Close this issue ?

jiridanek:
@ganeshmurthy The job still does not check that skstat/skmanage run, so there is more work to do. (They do run now, but it is not tested there.)

jiridanek:
One additional problem; the test does not wait for router to start and open port, so if it starts slowly, test fails. This is also something to be solved here.

See HEALTHCHECK Containerfile instruction, https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/assembly_monitoring-containers_building-running-and-managing-containers#performing-a-health-check-using-a-containerfile_assembly_monitoring-containers

Healthcheck feature is present in the router, https://github.com/skupperproject/skupper-router/commit/1e22255eede6f82828a433839cb25087c2c7b300 hopefully it does what I need it to do (claim Healthy only when all listeners declared in config were opened).

PR #928, opened 2023-01-31
Issue #925 - Fix Proton Python main installation for Containerfile

jiridanek:
I either need to copy the tar.gz from the build container to the run container and install it there, or I need to install the tar.gz in the build container under `--prefix`.

This PR does the latter.

### DIFF
--- a/.github/scripts/compile.sh
+++ b/.github/scripts/compile.sh
@@ -134,7 +134,7 @@ do_build () {
 # Do a regular build without asan or tsan.
 do_build "" OFF
 # and install Proton Python
-python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/dist/" -name 'python-qpid-proton*.tar.gz')"
+python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
 
 # Then perform sanitized builds of Proton and the Router.
 # talking to annobin is not straightforward, https://bugzilla.redhat.com/show_bug.cgi?id=1536569
--- a/.github/scripts/compile.sh
+++ b/.github/scripts/compile.sh
@@ -111,7 +111,7 @@ do_build () {
     -DRUNTIME_CHECK="${runtime_check}" \
     -DENABLE_LINKTIME_OPTIMIZATION=ON \
     -DCMAKE_POLICY_DEFAULT_CMP0069=NEW -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON \
-    -DBUILD_TLS=ON -DSSL_IMPL=openssl -DBUILD_STATIC_LIBS=ON -DBUILD_BINDINGS=python -DSYSINSTALL_PYTHON=ON \
+    -DBUILD_TLS=ON -DSSL_IMPL=openssl -DBUILD_STATIC_LIBS=ON -DBUILD_BINDINGS=python \
     -DBUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF \
     -DCMAKE_INSTALL_PREFIX=/usr
   cmake --build "${PROTON_BUILD_DIR}${suffix}" --verbose
@@ -133,7 +133,10 @@ do_build () {
 
 # Do a regular build without asan or tsan.
 do_build "" OFF
+# and install Proton Python
+python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/dist/" -name 'python-qpid-proton*.tar.gz')"
 
+# Then perform sanitized builds of Proton and the Router.
 # talking to annobin is not straightforward, https://bugzilla.redhat.com/show_bug.cgi?id=1536569
 common_sanitizer_flags="-Wp,-U_FORTIFY_SOURCE -fplugin=annobin -fplugin-arg-annobin-no-active-checks"
 export CFLAGS="${CFLAGS} ${common_sanitizer_flags}"
--- a/Containerfile
+++ b/Containerfile
@@ -23,7 +23,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
     rpm-build \
     gcc gcc-c++ make cmake \
     cyrus-sasl-devel openssl-devel libuuid-devel \
-    python39-devel python39-pip \
+    python39-devel python39-pip swig \
     libnghttp2-devel \
     wget tar patch findutils git libasan libubsan libtsan \
  && microdnf clean all -y
--- a/Containerfile
+++ b/Containerfile
@@ -23,7 +23,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
     rpm-build \
     gcc gcc-c++ make cmake \
     cyrus-sasl-devel openssl-devel libuuid-devel \
-    python39-devel swig \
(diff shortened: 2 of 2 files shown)


.github/scripts/compile.sh
```diff
@@ -133,7 +133,10 @@ do_build () {
 
 # Do a regular build without asan or tsan.
 do_build "" OFF
+# and install Proton Python
+python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/dist/" -name 'python-qpid-proton*.tar.gz')"
```
jiridanek, line 137: the script starts with

```
# https://sipb.mit.edu/doc/safe-shell
set -Eefuxo pipefail
```

so * expansion does not work... I stumbled upon this myself ;(

Containerfile
```diff
@@ -23,7 +23,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
     rpm-build \
     gcc gcc-c++ make cmake \
     cyrus-sasl-devel openssl-devel libuuid-devel \
-    python39-devel swig \
+    python39-devel python39-pip \
```
jiridanek, line 26: This will work only with Proton main. All versions released up to now need swig. Should we remove swig here now, or do I need to wait for something?

astitcher: I'd leave swig until 0.39 is released, should still need pip though and also wheel I think if you want to avoid compiling the library in the run container

.github/scripts/compile.sh
astitcher: I think you'd be better using the wheel if you want to avoid needing the compile chain in the run container - the pattern would give you `python_qpid_proton-0.39.0.dev0-cp311-cp311-linux_x86_64.whl` for 0.39-dev0 version for cpython 3.11.
The .tar.gz is more generally useful if you might have a different python in the build container and run container though.

astitcher: Also you might want ot use a multiple roots for find that will work for 0.38 and 0.39 to ease the transition. For 0.38 the path would be `python/pkgs` for 0.39 as you use here its `python/dist`.

astitcher: Sorry I misunderstood this change!
I think it would really be better to copy the pkg to the run container and use the active python there to install - that will make sure that you don't need to worry about where the prefix is. Just make sure that te python that runs `python -m pip install /..../python-qpid-proton...whl` or `....tar.gz` is the same one that will run the scripts that use proton.

fgiorgetti (approved):
It seems to fix the issue.

.github/scripts/compile.sh
jiridanek: > I think you'd be better using the wheel if you want to avoid needing the compile chain in the run container

I am quite happy with doing `--prefix="$PROTON_INSTALL_DIR/usr"` and copying that to the run container under /usr (that's what tar is used later for).

> Also you might want ot use a multiple roots for find that will work for 0.38 and 0.39 to ease the transition.

@ganeshmurthy I guess we want that...

> I think it would really be better to copy the pkg to the run container and use the active python there to install

It would be more foolproof, the pip install for the wheel would check that versions of Python match, which here is not checked... On the other hand, the python version is in the site-packages path, so actually, there is sort of check for that. Using wheel means one extra dependency for the build container.

> The .tar.gz is more generally useful if you might have a different python in the build container and run container though.

The wheel could also be multiversion, one day. It is a feature that I think is possible to implement, https://issues.apache.org/jira/browse/PROTON-2678

Containerfile
jiridanek: done

.github/scripts/compile.sh
astitcher: Ah multiversion wheel for linux seems like a very good idea right now - it doesn't seem too hard and that is what you get by default on windows already I'm pretty sure.

astitcher: The wheel python dependency is a very small extra dependency compared to having the compilers and cffi etc.

jiridanek: > Ah multiversion wheel for linux seems like a very good idea right now - it doesn't seem too hard and that is what you get by default on windows already I'm pretty sure.

Linux has also the "manylinux" aspect, IIRC. Not sure if we can have that on Linux, though, because that should mean no dependencies on external .so files, and Proton depends on cyrus sasl, ssl, and so on... But anyways, I might be wrong, and it's offtopic here.

jiridanek: > The wheel python dependency is a very small extra dependency compared to having the compilers and cffi etc.

I need cffi anyways, I think, isn't that a runtime dep now? And currently in this PR I build the proton package in build container, install it in that --prefix, and tar that and extract in run container (together with the rest of installed proton), so I don't have compiler in run image.

astitcher: I think the abi3 supercedes manylinux. manylinux is tricky for us because it relies on a fixed (and somewhat old) build environment to make sure that the c libraries link to the system glibc.

.github/scripts/compile.sh
```diff
@@ -134,7 +134,7 @@ do_build () {
 # Do a regular build without asan or tsan.
 do_build "" OFF
 # and install Proton Python
-python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/dist/" -name 'python-qpid-proton*.tar.gz')"
+python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/{pkgs,dist}/" -name 'python-qpid-proton*.tar.gz')"
```
astitcher, line 137: The {pkgs,dist} syntax is not recognised by find or expanded by the shell and causes the container build to fail.
I think you can just use multiple directories directly:
"$PROTON_BUILD_DIR/python/dist/" "$PROTON_BUILD_DIR/python/pkgs/"

astitcher (changes requested):
The container build is still failing

.github/scripts/compile.sh
jiridanek: Ouch. I guess I should've looked after pushing, if it works. I got distracted by something else.

That `{pkgs,dist}` would've been expanded by shell if I removed the quotes around it. (I tried it in shell without quotes, and thought that they don't make a difference.) Ok, I'll put there the dirs directly, that is resilient towards spaces in the paths...

jiridanek: > "$PROTON_BUILD_DIR/python/dist/" "$PROTON_BUILD_DIR/python/pkgs/"

This would not work either, because find fails if given nonexisting path. I'm going to give it `"$PROTON_BUILD_DIR/python`, that seems to work ok.

.github/scripts/compile.sh
```diff
@@ -133,7 +133,10 @@ do_build () {
 
 # Do a regular build without asan or tsan.
 do_build "" OFF
+# and install Proton Python
+python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
```
jiridanek, line 137: @astitcher Is there a reason why the wheel is named python_qpid_proton and tar.gz is python-qpid-proton? It is a bit weird.

astitcher: It seems that is just what the tooling does - I don't knwo why.

astitcher (approved):
lgtm and this change isn't related to the build failures anymore.

jiridanek:
> lgtm and this change isn't related to the build failures anymore.

@astitcher Thanks. I really don't want to use the wheel here. The tar.gz seems to me the most similar to how things worked before (with SYSINSTALL).

PR #942, opened 2023-02-03
Fixes #941: ignore CI errors due to missing LastTestsFailed.log

### LINKED ISSUE #941: Avoid failing log cleanup in CI if test run fails early
If the GHA CI fails early - for example during proton build - the step that tries to purge logs will also fail with a FileNotFound exception.

This causes the GHA UI to point to the FileNotFound error when first navigating to the test error logs. This causes confusion because the actual failure is from a previous step.

kgiusti:
Closes: #941

### DIFF
--- a/scripts/gha_purge_successful_test_logs.py
+++ b/scripts/gha_purge_successful_test_logs.py
@@ -44,7 +44,7 @@ def main() -> int:
 
     log_file = pathlib.Path(build_dir, 'Testing', 'Temporary', 'LastTestsFailed.log')
     if not log_file.exists():
-        logging.debug(f"File {log_file} does not exist - nothing to do")
+        logging.info(f"File {log_file} does not exist - nothing to do")
         return 0
 
     with log_file.open('rt') as f:
@@ -54,7 +54,7 @@ def main() -> int:
 
     system_tests_dir = pathlib.Path(build_dir, 'tests', 'system_test.dir', 'tests')
     if not system_tests_dir.exists():
-        logging.info("system_test.dir does not exits, nothing to do")
+        logging.info("system_test.dir does not exist, nothing to do")
         return 0
 
     for test_dir in system_tests_dir.iterdir():
--- a/scripts/gha_purge_successful_test_logs.py
+++ b/scripts/gha_purge_successful_test_logs.py
@@ -42,11 +42,15 @@ def main() -> int:
         logging.error("Build dir %s does not exist or is not a directory", build_dir)
         return 1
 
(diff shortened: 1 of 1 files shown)


scripts/gha_purge_successful_test_logs.py
```diff
@@ -42,11 +42,15 @@ def main() -> int:
         logging.error("Build dir %s does not exist or is not a directory", build_dir)
         return 1
 
-    with pathlib.Path(build_dir, 'Testing', 'Temporary', 'LastTestsFailed.log').open('rt') as f:
-        failed_tests = gha_tools.parse_last_tests_failed_log(f)
+    pathy_mcpathface = pathlib.Path(build_dir, 'Testing', 'Temporary', 'LastTestsFailed.log')
```
ganeshmurthy, line 45: pathy_mcpathface ?

scripts/gha_purge_successful_test_logs.py
```diff
@@ -42,11 +42,15 @@ def main() -> int:
...
+    if not pathy_mcpathface.exists():
+        logging.debug(f"File {pathy_mcpathface} does not exist - nothing to do")
+        return 0
 
+    with pathy_mcpathface.open('rt') as f:
+        failed_tests = gha_tools.parse_last_tests_failed_log(f)
     failed_test_names = {test[1] for test in failed_tests}
-    logging.debug("Failed tests: %s", failed_test_names)
+    logging.debug(f"Failed tests: %s", failed_test_names)
```
ganeshmurthy, line 53: logging.debug(f"Failed tests: ${failed_test_names}")

kgiusti: Ah, good catch - fix coming up

scripts/gha_purge_successful_test_logs.py
kgiusti: I really have to stop coding after happy hour....

ganeshmurthy: LOL

ganeshmurthy (approved):

jiridanek (approved):
LGTM, it will make reviewing the type of fails we had with Proton compilation easier.

I'd prefer to make the `logging.debug` at _info_ level instead, because that way it gets printed into the GHA output for everyone to see what happened, should anyone investigate. Also, that would follow the same logic as the

```
        logging.info("system_test.dir does not exits, nothing to do")
```

lower down in that file. Hah, there's a typo in the info log msg, I now see

PR #951, opened 2023-02-09
Possible fix for raw event change in Proton

astitcher:


astitcher:
Creating PR to make sure this doesn't break anything before makeing the corresponding change in Proton

astitcher:
@kgiusti I've tried this change against the new version of proton on my own machine and I'm still getting the qd_adaptor_buffer_t leaks so I guess it's not as simple as just this change,

astitcher:
ok I'm getting the same leak run against an unmodified proton too. so I'm confused

PR #964, opened 2023-02-21
Attempt to fix # 957. Try increasing the number of unaccepted connect…

ganeshmurthy:
…ions to 25

### DIFF
--- a/tests/TCP_echo_server.py
+++ b/tests/TCP_echo_server.py
@@ -165,7 +165,11 @@ class TcpEchoServer:
                     sck = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
                     sck.bind((self.HOST, int(self.port)))
                     self.sock = context.wrap_socket(sck, server_side=True)
-                    self.sock.listen(5)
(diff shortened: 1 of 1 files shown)


tests/TCP_echo_server.py
```diff
@@ -165,7 +165,11 @@ def run(self):
                     sck = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
                     sck.bind((self.HOST, int(self.port)))
                     self.sock = context.wrap_socket(sck, server_side=True)
-                    self.sock.listen(5)
+                    # self.sock.listen(n) - where n is the number of unaccepted connections that the
+                    # system will allow before refusing new connections.
+                    # Since the CI system is low on resources, it might take time for the echo server
+                    # to accept TLS connections. Test setting the value of n to 25
+                    self.sock.listen(25)
```
kgiusti, line 172: Alternatively, you can set it to socket.SOMAXCONN, which is the system default maximum.

ganeshmurthy: Done ! Thanks.

PR #965, opened 2023-02-23
Always run `clang-format` against diff with upstream/main branch, even in user forks

jiridanek:
Previously, when running GHA in a user's project on a branch, clang-format would compare that branch with user's main branch. When that copy of the main branch is behind the changed branch, there will be some random diff in C code due to this, and clang-format will start reporting issues:

image

The solution seems to be to ensure that clang-format is always doing a comparison with the upstream main (at skupperproject/skupper-router). Then, the very same branch passes green.

image

This should reduce the e-mail noise from github caused by failing clang-format when it is actually running on nonsensical diffs.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -611,8 +611,11 @@ jobs:
     steps:
 
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
+
+      - name: Fetch upstream main reference
+        run: |
+          git remote add upstream https://github.com/skupperproject/skupper-router.git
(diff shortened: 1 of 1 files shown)


mgoulish (approved):
Excellent. This will be a big help.
Nihil Obstat.
Imprimatur.

jiridanek:
@mgoulish I guess I've grown desensitized to that fail and learned to ignore it as a matter of routine... and I somehow presumed that's how everyone feels, cause nobody neither complained (to me) nor filled issue nor fixed it themselves

PR #968, opened 2023-02-23
Fixes #967: Introduce LTO in router builds

### LINKED ISSUE #967: Enable Link Time Optimization in the skupper-router build
https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

ganeshmurthy:


CMakeLists.txt
```diff
@@ -42,9 +43,24 @@ set(C_WARNING_Clang -Wall -Wextra -Werror -Wpedantic
...
-# Additional compiler-specific flags
-#  -gdwarf-aranges to workaround https://sourceware.org/bugzilla/show_bug.cgi?id=22288 (backtrace line printing with libdw and clang)
 set(C_EXTRA_GNU )
+if (ENABLE_LINKTIME_OPTIMIZATION)
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT ipo_supported OUTPUT output)
+    if(ipo_supported)
+      message(STATUS "Interprocedural optimization is supported")
+      set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+      set(C_EXTRA_GNU -fno-fat-lto-objects)
```
jiridanek, line 53: Isn't this the default? Judging from https://gitlab.kitware.com/cmake/cmake/-/issues/21696 and https://gitlab.kitware.com/cmake/cmake/-/blob/eaea031fb01d0c97b132e9809702da7eb0880567/Modules/Compiler/GNU.cmake#L99

I guess I would not worry about the flag either way and just let CMake to set its default LTO flags. This seems possibly important for libraries, and the router is not one.

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 set(CMAKE_ENABLE_EXPORTS TRUE)
+option(ENABLE_LINKTIME_OPTIMIZATION "Perform link time optimization" ON)
```
jiridanek, line 35: In general, I think that the default and options in cmake scripts should work differently

## What it is now

| User preference     | IPO available | IPO not available    |
|---------------------|---------------|----------------------|
| None (default)      | ON            | OFF (w/ warning)     |
| Explicitly enabled  | ON            | **OFF (w/ warning)** |
| Explicitly disabled | OFF           | OFF                  |

## What it should be

| User preference     | IPO available | IPO not available |
|---------------------|---------------|-------------------|
| None (default)      | ON            | OFF (maybe w/out warning)  |
| Explicitly enabled  | ON            | **Error**         |
| Explicitly disabled | OFF           | OFF               |

That is, if user enabled it and we cannot deliver, then error out.

jiridanek (approved):
One more thing. I think that the option in the router should match whatever @astitcher is going to do in Proton, because it would be nicer that way.

CMakeLists.txt
ganeshmurthy: Done ! Thanks.

CMakeLists.txt
kgiusti: @ganeshmurthy my $0.02US - I agree with @jiridanek that if the user has explicitly configured IPO and it isn't provided it should fail. That prevents me from inadvertently benchmarking IPO vs non-IPO and getting misleading results. 

Again, just my humble opinion.

ganeshmurthy: @jiridanek @kgiusti I have made the requested changes, please take a look. Thanks.

ganeshmurthy: I have to make a couple more changes to this PR, hang on.

.github/scripts/compile.sh
```diff
@@ -122,7 +122,7 @@ do_build () {
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DRUNTIME_CHECK="${runtime_check}" \
     -DSANITIZE_PYTHON=OFF \
-    -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON \
+    -DENABLE_LINKTIME_OPTIMIZATION=ON \
```
ssorj, line 125: I think we should align on the cmake name for this, CMAKE_INTERPROCEDURAL_OPTIMIZATION (so what you had originally).

See also the note about policy CMP0069 here:

https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html

ganeshmurthy: Proton uses `ENABLE_LINKTIME_OPTIMIZATION` so I used the same in skupper-router.

If you look at the Examples section of policy CMP0069 - https://cmake.org/cmake/help/latest/policy/CMP0069.html#policy:CMP0069 - it suggests the use of CMake's `check_ipo_supported()` which we use here. If `check_ipo_supported()` is true, we call `set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)`

CMakeLists.txt
```diff
@@ -42,9 +43,23 @@ set(C_WARNING_Clang -Wall -Wextra -Werror -Wpedantic
...
         -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-language-extension-token)
 
-# Additional compiler-specific flags
-#  -gdwarf-aranges to workaround https://sourceware.org/bugzilla/show_bug.cgi?id=22288 (backtrace line printing with libdw and clang)
 set(C_EXTRA_GNU )
+if (ENABLE_LINKTIME_OPTIMIZATION)
+    include(CheckIPOSupported)
+    check_ipo_supported(RESULT ipo_supported OUTPUT output)
+    if(ipo_supported)
+      message(STATUS "Interprocedural optimization is supported")
```
jiridanek, line 51:
```suggestion
      message(STATUS "Interprocedural optimization is enabled")
```

me thinks that makes more sense as a status message in cmake output

README.adoc
```diff
@@ -363,7 +363,7 @@ Other options include `Debug` (disables optimizations) and `Coverage`.
 |Enables Leak Sanitizer suppressions for libpython3.
 Use it to disable sanitization with older Python version (< 3.9).
 
-|`-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`
+|`-DENABLE_LINKTIME_OPTIMIZATION=ON`
 |With CMake 3.9+, compiles the project with LTO (Link Time Optimization) enabled.
```
jiridanek, line 367: Maybe remove the "With CMake 3.9+" , and "Older versions of CMake", because we now require at least 3.20 in any case

https://github.com/skupperproject/skupper-router/blob/45d75bbfd428ef0ec03239d127125d786f11aa9c/CMakeLists.txt#L20

.github/scripts/compile.sh
ssorj: Last I spoke to Andrew, he was planning to switch Proton to the standard CMake name.

CMakeLists.txt
ganeshmurthy: Done !

README.adoc
ganeshmurthy: Done !

.github/scripts/compile.sh
ganeshmurthy: Switched to use the standard CMake name `CMAKE_INTERPROCEDURAL_OPTIMIZATION`

ganeshmurthy:
@jiridanek @ssorj can you please take a look at my latest changes in this PR and critique/approve this PR. Thanks.

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 set(CMAKE_ENABLE_EXPORTS TRUE)
+option(CMAKE_INTERPROCEDURAL_OPTIMIZATION "Perform link time optimization" OFF)
```
jiridanek, line 35: I thought the intention is to default this to ON, same as in Qpid Proton.

Otherwise, users can just set `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` and there is no need for any additional CMakeLists code to support this thing.

CMakeLists.txt
```diff
@@ -42,9 +43,23 @@ set(C_WARNING_Clang -Wall -Wextra -Werror -Wpedantic
         -Wstrict-prototypes -Wold-style-definition
         -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-language-extension-token)
 
-# Additional compiler-specific flags
-#  -gdwarf-aranges to workaround https://sourceware.org/bugzilla/show_bug.cgi?id=22288 (backtrace line printing with libdw and clang)
 set(C_EXTRA_GNU )
```
jiridanek, line 46: Why did you put the `if (CMAKE_INTERPROCEDURAL_OPTIMIZATION)` in between the `C_EXTRA_GNU` and `C_EXTRA_Clang` variables? The way I feel it, the C_EXTTRA stuff should be kept together and INTERPROCEDURAL_OPTIMIZATION can be either above it or below it... But this is just a nit.

jiridanek:
@ganeshmurthy Fine with me. It works, I won't hold you back from merging. If @astitcher changes something LTO-related in Proton that would justify revisiting it here, then we can just revisit it there and then.

CMakeLists.txt
ganeshmurthy: Good catch. I moved it up before the if statement because I *was* setting the `-fno-fat-lto-objects` explicitly but that was later removed since cmake defaults it anyways. I will fix this.

CMakeLists.txt
astitcher: I'm not entirely sure that there is any work to allow LTO now that cmake supports it.

The only reason there is specific work in proton is that we supported it on versions of cmake too early to support it directly.

We're moving to remove any specific support and just to rely on the CMAKE_INTERPROCEDURAL_OPTIMIZATION cmake setting. The support for ENABLE_LINKTIME_OPTIMIZATION that is still there is purely for backwards  compatiblilty of build scripts.

astitcher: What I'm saying is that the only thing you should need to do in a recent version of cmake (proton now requries 3.16 or later) is to set CMAKE_INTERPROCEDURAL_OPTIMIZATION=On. In a build script all you need to do is add "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=On" to the cmake line.

CMakeLists.txt
astitcher: FWIW Proton does not make this an option - the intent is just to rely on the underlying cmake support and to remove the specific option.

ganeshmurthy: Proton does seem to have an option called `ENABLE_LINKTIME_OPTIMIZATION`
`option(ENABLE_LINKTIME_OPTIMIZATION "Perform link time optimization" ${DEFAULT_LINKTIME_OPTIMIZATION})`

https://github.com/apache/qpid-proton/blob/main/CMakeLists.txt#L149

astitcher: As I said above that is for backwards compatibility which you don't need as you have no previous setting for this.

by 'make this an option' I mean literally use the option() command in cmake - you can just use a command line setting to turn this on unless you want to default this to on in which case you will need to use some way of setting it.

ganeshmurthy: > I thought the intention is to default this to ON, same as in Qpid Proton.
> 
> Otherwise, users can just set `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` and there is no need for any additional CMakeLists code to support this thing.

You got your wish.

PR #977, opened 2023-03-08
Fixes #539: remove pn_ leak suppressions after upgrading to Proton 0.38.0

### LINKED ISSUE #539: Leak of `pn_raw_connection` in h2spec tests with Proton 0.38.0
> The pn_raw_connection has a pn_collector which is linked to pn_event if there is an outstanding pn_event with a raw connection event then that will have a reference to the original pn_connection itself
> so I think we must have exited the event loop somehow without clearing the event

> probably not really an issue - it'll only leak in some slightly rare cases where the connection gets made but never used at all
> as in the connection is completed and generates the 'connected' event but nother ever takes that event from the event loop

This leak disappears when Proton main (what's to be Proton 0.38.0) is used

> So maybe we can remove the other suppressions when we release 0.38

Task is to remove suppression when Proton 0.37.0 CI jobs are updated to 0.38.

jiridanek:


astitcher:
If removing these suppressions still gives clean results then definitely go for it. It would be good not to suppress these things as they shouldn't leak.

ganeshmurthy (approved):

Issue #966, opened 2023-02-23
SEGFAULT in _vflow_set_int_TH during Http1AdaptorBadEndpointsTest::test_08_bad_chunked_body test


kgiusti:
https://github.com/skupperproject/skupper-router/actions/runs/4244471490/jobs/7378637317#step:30:3233

This test simulates badly formatted HTTP/1 messages with chunked body encoding. The router will respond by closing the associated connection when the parse error is hit. The crash occurs shortly after the last connection closes.

The fault appears to occur during _vflow_find_attribute():

```c
  static vflow_attribute_data_t* _vflow_find_attribute(vflow_record_t *record, vflow_attribute_t attr)
  {
    vflow_attribute_data_t *data = DEQ_TAIL(record->attributes);
    while(!!data) {
     if (data->attribute_type <= attr) {  <<<<<< HERE

```

```
2023-02-22T16:03:46.1800912Z Program terminated with signal SIGSEGV, Segmentation fault.
2023-02-22T16:03:46.1801241Z #0  0x00007fd690d7eb8f in raise () from /lib64/libpthread.so.0
2023-02-22T16:03:46.1953650Z [Current thread is 1 (Thread 0x7fd6895c6700 (LWP 5506))]
2023-02-22T16:03:46.1953843Z 
2023-02-22T16:03:46.1953962Z Thread 7 (Thread 0x7fd6865c3700 (LWP 5509)):
2023-02-22T16:03:46.1954258Z #0  0x00007fd690d7db47 in close () from /lib64/libpthread.so.0
2023-02-22T16:03:46.1954992Z #1  0x00007fd6915ebd3b in pclosefd (fd=<optimized out>, p=0x2d3ce30) at /home/runner/work/skupper-router/skupper-router/qpid-proton/c/src/proactor/epoll.c:755
2023-02-22T16:03:46.1955751Z #2  praw_connection_cleanup (prc=<optimized out>, prc=<optimized out>) at /home/runner/work/skupper-router/skupper-router/qpid-proton/c/src/proactor/epoll_raw_connection.c:155
2023-02-22T16:03:46.1956517Z #3  0x00007fd6915efd36 in pni_raw_connection_done (rc=0x3acd1c0) at /home/runner/work/skupper-router/skupper-router/qpid-proton/c/src/proactor/epoll_raw_connection.c:431
[...]

PR #997, opened 2023-03-13
Fixes #996 - Use test-and-set result to avoid spurious wakes in the H…

### LINKED ISSUE #996: Conditionally wake on q2_restart
Since sys_atomic_set performs an atomic test-and-set, spurious wakes can be avoided by waking the connection only when the q2_restart condition transitions from 0 to 1.

ted-ross:
…TTP1 adaptor.


kgiusti (approved):
@ted-ross I say we merge.  The two http1 failures are other unrelated known issues: #1000 (fix available) and #966. And the patch is fairly obvious Good To Have.

PR #999, opened 2023-03-13
ISSUE #998: Increase the qd_buffer_t payload size to 4096 (less overh…

kgiusti:
…ead)

Fixes: #998


ssorj:
LGTM.  I see a ~15% improvement in TCP throughput with Flimflam's builtin workload.  Otherwise, memory and CPU and HTTP workloads seem to be unaffected by the change.

Further testing with 100 concurrent client sessions gets me a substantial boost in iperf3 (looks like more than 30%) and reduced memory usage (relative to main) with the builtin workload.

ganeshmurthy (approved):

PR #1008, opened 2023-03-22
Use Fedora 38 in GitHub Actions CI builds

jiridanek:
Fedora 38 is currently a Beta, so it's I think a good time to start considering a switch in CI. https://fedorapeople.org/groups/schedule/f-38/f-38-key-tasks.html

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -313,7 +313,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         container: ['fedora']
-        containerTag: ['37']
+        containerTag: ['38']
         cc: [gcc]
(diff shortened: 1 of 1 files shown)


ganeshmurthy:
Looks like F38 is going to be released on April 18th.

ganeshmurthy:
@jiridanek Very much shocked/surprised that *every single* test CI failure is related to http2 tests.

jiridanek:
> @jiridanek Very much shocked/surprised that *every single* test CI failure is related to http2 tests. 

And now we know how/why!

The spirit of th pr lives on in

* 
https://github.com/skupperproject/skupper-router/pull/1374

May it find more luck in the next life

Issue #339, opened 2022-04-15
Print source file and line number in alloc_pool leak backtraces automatically


jiridanek:
Currently, the code uses the [`backtrace_symbols`](http://manpages.ubuntu.com/manpages/jammy/en/man3/backtrace.3.html) function to format backtraces. This has several advantages, chiefly that it does not even require debug information to be present, compiling with `-rdynamic` is enough.

https://github.com/skupperproject/skupper-router/blob/e5b28d2fdbade7404d4dc32d847a6d88c4e003e1/CMakeLists.txt#L34

Problem is that the output is rather uninspiring and to obtain code locations, manual processing with a 3rd party symbolizer (such as [`addr2line`](http://manpages.ubuntu.com/manpages/jammy/en/man1/addr2line.1.html) is necessary). This can of course be only done for builds with debug symbols.

```
Leak:  type: xxx address: 0x7ffead01f400
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests(b_stores_backtrace+0x1d) [0x5acde6]
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests(a_calls_b+0x18) [0x5ace08]
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests() [0x5ad1b2]
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests(_ZN7doctest7Context3runEv+0x1083) [0x561e8d]
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests(main+0x98) [0x562a3a]
/lib64/libc.so.6(+0x2d560) [0x7f586db1e560]
/lib64/libc.so.6(__libc_start_main+0x7c) [0x7f586db1e60c]
/home/jdanek/repos/skupper-router/cmake-build-debug/tests/c_unittests/c_unittests(_start+0x25) [0x4b71e5]
```

What we want instead, is something like this

![image](https://user-images.githubusercontent.com/442720/163545426-cc539a41-7f92-4127-a15d-d01eaa3e8e12.png)

There are some libraries that provide a high level API to get such nice stacktraces

* https://github.com/ianlancetaylor/libbacktrace (stand-alone version of [code used in e.g. GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66570) AddressSanitizer)
[...]

PR #1018, opened 2023-03-24
Add Clang GitHub Actions CI job

jiridanek:
This required fixing https://issues.apache.org/jira/browse/PROTON-2691 first, so only Proton main can be compiled with clang currently.

* Also, merging https://github.com/skupperproject/skupper-router/pull/1020 is required to get passing tests with LTO

PR #1020, opened 2023-03-27
Issue #339: use the bombela/backward-cpp scan for symbolized backtrace clang compatibility

jiridanek:
This is necessary if LTO is used on Clang, because then the `-gdwarf-aranges` option stops working.


jiridanek:
I'm considering this approved, with the approve on a superset PR https://github.com/skupperproject/skupper-router/pull/1018

Issue #1015, opened 2023-03-24
Runtime Debug tests to validate proper thread execution


kgiusti:
There are execution paths throughout the router codebase that are thread-specific. These codepaths are restricted to executing only on a particular type of thread. This features provides a way to annotate those codepaths. For debug builds these annotations will trigger process abort when threading violations occur.

PR #1036, opened 2023-03-30
Add Clang Thread Safety Annotations (locks-only)

jiridanek:
* More easily digestible subset of https://github.com/skupperproject/skupper-router/pull/212

Uses https://clang.llvm.org/docs/ThreadSafetyAnalysis.html (with all the limitations that stem from it, namely, works only in clang or IDE based on clangd, and the GUARDED_BY annotation cannot be used in C structs, https://github.com/llvm/llvm-project/issues/20777)

# How it works in action

Example of a text-only compile failure message: https://github.com/jiridanek/skupper-router/actions/runs/4555107395/jobs/8033958647#step:25:282

Example of the clangd IDE integration:

## Calling a `_LH` function without holding the lock

warning_should_hold_lock

## Unlocking something that is not locked

releasing mutex not held

## Returning without unlocking

return_without_unlock

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ set(C_WARNING_GNU -Wall -Wextra -Wpedantic -pedantic-errors
         -Wno-empty-body -Wno-unused-parameter  -Wno-missing-field-initializers -Wno-sign-compare -Wno-type-limits)
 set(C_WARNING_Clang -Wall -Wextra -Wpedantic
         -Wstrict-prototypes -Wold-style-definition
-        -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-language-extension-token
+        -Wthread-safety -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-language-extension-token
         -Wno-gnu-statement-expression-from-macro-expansion)
 
 # Set default build type. Must use FORCE because project() sets default to ""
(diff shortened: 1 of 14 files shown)


jiridanek:
Related to https://github.com/skupperproject/skupper-router/issues/1015

jiridanek:
I'm reasonably confident that this will be low-noise and will not be causing annoying problems.

Judging from past history, the only way to get this in is probably to just push it to main and then see whether my confidence was justified, or it does cause some problems and needs to be reverted.

I still have the previous full PR (which contains additional annotation for running on the core thread) in the queue

* https://github.com/skupperproject/skupper-router/pull/212

ganeshmurthy:
> I'm reasonably confident that this will be low-noise and will not be causing annoying problems.
> 
> Judging from past history, the only way to get this in is probably to just push it to main and then see whether my confidence was justified, or it does cause some problems and needs to be reverted.
> 
> I still have the previous full PR (which contains additional annotation for running on the core thread) in the queue
> 
>     * [Add Clang Thread Safety Annotations #212](https://github.com/skupperproject/skupper-router/pull/212)

This is a big enough change that I would at least like @kgiusti to sign off on it before this lands on main

jiridanek:
> This is a big enough change that I would at least like @kgiusti to sign off on it before this lands on main

He said this about it (long time ago) https://github.com/skupperproject/skupper-router/pull/212#pullrequestreview-921673100

I really think this will be low friction in practice. That was the whole point of the "reduced" PR, after all. And if not, I'm ready with my finger on the revert button. Either way, I really want to conclude this "initiative" in some way. At least this way it will be shown if thread-safety-annotations are beneficial or painful.

PR #1044, opened 2023-04-03
Fixes #1031: Added new option for PGO called ENABLE_PROFILE_GUIDED_OP…

### LINKED ISSUE #1031: Enable Profile Guided Optimization for gcc in skupper-router
There are many optimize options available in GNU GCC. One of them is Profile Guided Optimization

Introduce a new option called ENABLE_PROFILE_GUIDED_OPTIMIZATION. When this option is turned on - 

1. Compile the source code with the appropriate flags (`-fprofile-generate` in the case of GCC) which
will create an instrumented skrouterd binary.
2. Run the tests using the instrumented skrouterd binary which will produce the profiling data.
3.  Feed back the generated profiling data by using the `-fprofile-use` option of gcc

ganeshmurthy:
…TIMIZATION.

### DIFF
--- a/.github/scripts/compile.sh
+++ b/.github/scripts/compile.sh
@@ -130,6 +130,13 @@ do_build () {
   local suffix=${1}
   local runtime_check=${2}
 
+  if [ "$runtime_check" == "OFF" ]; then
+     # Turn on PGO only when we are not doing asan or tsan
+     BUILD_OPTS="-DENABLE_PROFILE_GUIDED_OPTIMIZATION=ON"
+  else
+     BUILD_OPTS="-DBUILD_TESTING=OFF"
+  fi
+
   cmake -S "${PROTON_DIR}" -B "${PROTON_BUILD_DIR}${suffix}" \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DRUNTIME_CHECK="${runtime_check}" \
@@ -142,21 +149,31 @@ do_build () {
 
   DESTDIR="$PROTON_INSTALL_DIR${suffix}" cmake --install "$PROTON_BUILD_DIR${suffix}"
 
+   if [ "$runtime_check" == "OFF" ]; then
+     # This will install the proton python libraries in sys.path so the tests using
+     # proton can be run successfully.
+     python3 -m pip install "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
+   fi
+
   cmake -S "${SKUPPER_DIR}" -B "${SKUPPER_BUILD_DIR}${suffix}" \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DRUNTIME_CHECK="${runtime_check}" \
     -DProton_USE_STATIC_LIBS=ON \
     -DProton_DIR="${PROTON_INSTALL_DIR}${suffix}/usr/lib64/cmake/Proton" \
-    -DBUILD_TESTING=OFF \
+    ${BUILD_OPTS} \
     -DVERSION="${VERSION}" \
     -DCMAKE_INSTALL_PREFIX=/usr
   cmake --build "${SKUPPER_BUILD_DIR}${suffix}" --verbose
 }
 
+# This is required to install the python packages that the system tests use.
+python3 -m pip install -r "${SKUPPER_DIR}"/requirements-dev.txt
+
 # Do a regular build without asan or tsan.
 do_build "" OFF
-# and install Proton Python
-python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
+
+# Install Proton Python
+python3 -m pip install --ignore-installed --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
 
 # Then perform sanitized builds of Proton and the Router.
 # talking to annobin is not straightforward, https://bugzilla.redhat.com/show_bug.cgi?id=1536569
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 set(CMAKE_ENABLE_EXPORTS TRUE)
 option(CMAKE_INTERPROCEDURAL_OPTIMIZATION "Perform link time optimization" ON)
 option(ENABLE_WARNING_ERROR "Consider compiler warnings to be errors" ON)
+option(ENABLE_PROFILE_GUIDED_OPTIMIZATION "Perform profile guided optimization" OFF)
 
 # preserve frame pointers for ease of debugging and profiling
 #  see https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer
@@ -46,6 +47,18 @@ set(C_WARNING_GNU -Wall -Wextra -Wpedantic -pedantic-errors
 set(C_WARNING_Clang -Wall -Wextra -Wpedantic
         -Wstrict-prototypes -Wold-style-definition
         -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-language-extension-token)
+
+# Set default build type. Must use FORCE because project() sets default to ""
+if (NOT CMAKE_BUILD_TYPE)
+  set (CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
+    "Build type: Debug, Release, RelWithDebInfo, MinSizeRel or Coverage (default RelWithDebInfo)" FORCE)
+endif(NOT CMAKE_BUILD_TYPE)
+if (CMAKE_BUILD_TYPE MATCHES "Deb|Cover")
+  set (has_debug_symbols " (has debug symbols)")
+endif (CMAKE_BUILD_TYPE MATCHES "Deb|Cover")
+message(STATUS "Build type is \"${CMAKE_BUILD_TYPE}\"${has_debug_symbols}")
+# This is commonly needed so define it before we include anything else.
+string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
 
 if (CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     include(CheckIPOSupported)
@@ -66,6 +79,26 @@ endif (CMAKE_INTERPROCEDURAL_OPTIMIZATION)
 set(C_EXTRA_GNU )
 set(C_EXTRA_Clang )
 
+if (ENABLE_PROFILE_GUIDED_OPTIMIZATION)
+    message(STATUS "Profile guided optimization is enabled on ${CMAKE_C_COMPILER_ID} compiler")
+    # Set the build type to coverage since we need coverage related
+    # build options when -fprofile-generate is used.
+    if("GNU" STREQUAL ${CMAKE_C_COMPILER_ID})
+        # Temporarily modify the CMAKE_BUILD_TYPE to Coverage. Note that we are not setting the
+        # CMAKE_BUILD_TYPE to Coverage using the CACHE option. This means that when this build is over,
+        # we will retain the original value of CMAKE_BUILD_TYPE
+        set(CMAKE_BUILD_TYPE Coverage)
+        #  You must use -fprofile-generate both when compiling *and* when linking your program
+        set(C_EXTRA_GNU ${C_EXTRA_GNU} -fprofile-generate=${CMAKE_BINARY_DIR}/profile-data -fprofile-correction)
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fprofile-generate=${CMAKE_BINARY_DIR}/profile-data")
+    endif("GNU" STREQUAL ${CMAKE_C_COMPILER_ID})
+    if("Clang" STREQUAL ${CMAKE_C_COMPILER_ID})
+        # We are currently not supporting PGO on Clang although clang has good options to do PGO.
+        # Any contributions to skupper-router in this regard is welcome.
+        message(FATAL_ERROR "Profile Guided optimization of skupper-router not supported on ${CMAKE_C_COMPILER_ID}. Please consider contributing this feature")
+    endif("Clang" STREQUAL ${CMAKE_C_COMPILER_ID})
+endif (ENABLE_PROFILE_GUIDED_OPTIMIZATION)
+
 set(C_WARNING_FLAGS ${C_WARNING_${CMAKE_C_COMPILER_ID}})
 set(C_EXTRA_FLAGS ${C_EXTRA_${CMAKE_C_COMPILER_ID}})
 add_compile_options(${C_WARNING_FLAGS} ${C_EXTRA_FLAGS})
@@ -73,18 +106,6 @@ add_compile_options(${C_WARNING_FLAGS} ${C_EXTRA_FLAGS})
 if (ENABLE_WARNING_ERROR)
     add_compile_options(-Werror)
 endif (ENABLE_WARNING_ERROR)
-
-# Set default build type. Must use FORCE because project() sets default to ""
-if (NOT CMAKE_BUILD_TYPE)
-  set (CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
-    "Build type: Debug, Release, RelWithDebInfo, MinSizeRel or Coverage (default RelWithDebInfo)" FORCE)
-endif(NOT CMAKE_BUILD_TYPE)
-if (CMAKE_BUILD_TYPE MATCHES "Deb|Cover")
-  set (has_debug_symbols " (has debug symbols)")
-endif (CMAKE_BUILD_TYPE MATCHES "Deb|Cover")
-message(STATUS "Build type is \"${CMAKE_BUILD_TYPE}\"${has_debug_symbols}")
-# This is commonly needed so define it before we include anything else.
-string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
 
 if(uppercase_CMAKE_BUILD_TYPE STREQUAL "DEBUG" )
     option(QD_ENABLE_ASSERTIONS "Enable assertions" ON)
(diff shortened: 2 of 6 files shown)


Containerfile
```diff
@@ -48,6 +48,8 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
     cyrus-sasl-lib cyrus-sasl-plain cyrus-sasl-gssapi openssl \
     python3 \
     libnghttp2 \
+    curl \
+    nmap-ncat \
```
jiridanek, line 52: maybe put these on the same line? bash does not really allow to put comments between these lines, so the grouping is the only way to provide an idea what the deps are for. and since both are test dependencies, i"d put them on the same line
```suggestion
    curl nmap-ncat \
```

tests/system_test.py
```diff
@@ -1139,7 +1139,8 @@ def rmtree(self):
     def setup(self):
         """Called from test setup and class setup."""
         if self.directory:
-            os.makedirs(self.directory)
+            if not os.path.isdir(self.directory):
+                os.makedirs(self.directory)
```
jiridanek, line 1143:
```suggestion
            os.makedirs(self.directory, exist_ok=True)
```

.github/scripts/compile.sh
```diff
@@ -106,33 +106,49 @@ do_build () {
...
   DESTDIR="$PROTON_INSTALL_DIR${suffix}" cmake --install "$PROTON_BUILD_DIR${suffix}"
 
+   if [ "$runtime_check" == "OFF" ]; then
+     # Install Proton Python
+     python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
+
+     # This will install the proton python libraries in sys.path so the tests using
+     # proton can be run successfully.
+     python3 -m pip install "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
+   fi
```
jiridanek, line 135: I'd place this unconditionally before the

```
# Do a regular build without asan or tsan.
do_build "" OFF
```

(at line 133/151)

.github/scripts/compile.sh
```diff
@@ -106,33 +106,49 @@ do_build () {
...
+  else
+     BUILD_OPTS="-DBUILD_TESTING=OFF"
+  fi
+
   cmake -S "${PROTON_DIR}" -B "${PROTON_BUILD_DIR}${suffix}" \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DRUNTIME_CHECK="${runtime_check}" \
     -DENABLE_LINKTIME_OPTIMIZATION=ON \
-    -DCMAKE_POLICY_DEFAULT_CMP0069=NEW -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON \
+    -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON \
```
jiridanek, line 120: Does this change still work to produce LTO build of Proton? Looking at the build log, I'm surprised that it actually does, https://github.com/skupperproject/skupper-router/actions/runs/4599414496/jobs/8124756301?pr=1044#step:3:1666

```
[  1%] Building C object c/CMakeFiles/qpid-proton-tls-static.dir/src/tls/openssl.c.o
cd /build/proton/app/build/c && /usr/bin/gcc -DPROTON_DECLARE_STATIC -I/build/proton/app/c/include -I/build/proton/app/c/src -I/build/proton/app/build/c/include -I/build/proton/app/build/c/src -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong   -m64 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fvisibility=hidden -O2 -g -DNDEBUG -flto -fno-fat-lto-objects -std=c99 -Werror -Wall -pedantic-errors -Wno-unused-parameter -Wstrict-prototypes -Wvla -Wsign-compare -Wwrite-strings  -MD -MT c/CMakeFiles/qpid-proton-tls-static.dir/src/tls/openssl.c.o -MF CMakeFiles/qpid-proton-tls-static.dir/src/tls/openssl.c.o.d -o CMakeFiles/qpid-proton-tls-static.dir/src/tls/openssl.c.o -c /build/proton/app/c/src/tls/openssl.c
```

I guess this is because minimal required cmake version in proton was increased and so this CMP0069 policy no longer applies.

.github/scripts/compile.sh
jiridanek: The `# Install Proton Python` can remain where it is, after the builds. That installs proton for the output image.

jiridanek:
```
error: Could not depsolve transaction; 1 problem detected:
 Problem: problem with installed package curl-minimal-7.76.1-19.el9_1.1.x86_64
  - package curl-minimal-7.76.1-19.el9_1.1.x86_64 conflicts with curl provided by curl-7.76.1-19.el9_1.1.x86_64
  - conflicting requests
```

https://github.com/skupperproject/skupper-router/actions/runs/4599414496/jobs/8124756301?pr=1044#step:3:10276

This happens on centos 9 as well, it is the reason for the line 603 in

https://github.com/skupperproject/skupper-router/blob/18f89c81dd74cbb39023037d26487c12fba6a2a0/.github/workflows/build.yaml#L600-L604

I guess on ubi9 you can install `curl-minimal`, which supports http2.

.github/scripts/compile.sh
jiridanek: Note to /self: check that the proton static lib also gets lto flags

Containerfile
ganeshmurthy: Will do. I noticed that even after including nmap-ncat, the ncat tcp adaptor tests don't run. 

On an unrelated note, I am worried that including the packages needed for testing will increase the likelihood that the resulting skupper-router image might be involved in unneeded CVEs that might arise from these packages.

tests/system_test.py
ganeshmurthy: Done ! Thanks

.github/scripts/compile.sh
ganeshmurthy: Yes, it looks like `CMP0069` no longer applies. It showed up as an error when running cmake. The error went away after I removed `-DCMAKE_POLICY_DEFAULT_CMP0069=NEW`

Containerfile
jiridanek: > On an related note, I am worried that including the packages needed for testing will increase the likelihood that the resulting skupper-router image might be involved in unneeded CVEs that might arise from these packages.

It won't, we use https://docs.docker.com/build/building/multi-stage/ in our containerfile

CMakeLists.txt
```diff
@@ -62,22 +74,30 @@ endif (CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
+    message(STATUS "Profile guided optimization is enabled on ${CMAKE_C_COMPILER_ID} compiler")
+    # Set the build type to coverage since we need coverage related
+    # build options when -fprofile-generate is used.
+    if("GNU" STREQUAL ${CMAKE_C_COMPILER_ID})
+        # Temporarily modify the CMAKE_BUILD_TYPE to Coverage. Note that we are not setting the
+        # CMAKE_BUILD_TYPE to Coverage using the CACHE option. This means that when this build is over,
+        # we will retain the original value of CMAKE_BUILD_TYPE
+        set(CMAKE_BUILD_TYPE Coverage)
+        #  You must use -fprofile-generate both when compiling *and* when linking your program
+        set(C_EXTRA_GNU ${C_EXTRA_GNU} -fprofile-generate=${CMAKE_BINARY_DIR}/profile-data -fprofile-correction)
```
jiridanek, line 87: I happened to run across https://go.googlesource.com/proposal/+/master/design/55022-pgo.md, which discussed two different methods of doing PGO. Either tracing coverage (what this PR does), or a newer approach of sampling coverage in production, which is what the linked document also explains in detail.

See the heading "AutoFDO vs. instrumentation-based FDO".

My conclusion from reading the Go proposal is that for skupper-router, AutoFDO is not necessary and that `-fprofile-generate` is good enough.

Interesting quote:

> Google‘s AutoFDO (for C/C++ programs) is built on LLVM’s sample-based PGO, along with other toolings and mechanisms such as [Google-wide profiling (GWP)](https://research.google/pubs/pub36575). [AutoFDO](https://research.google/pubs/pub45290) improves the performance of C/C++ programs by 5–15% in Google datacenters.

.github/scripts/compile.sh
jiridanek: Even static lib gets now LTO. Ok, it really does work.

router/CMakeLists.txt
```diff
@@ -35,4 +35,30 @@ set(router_SOURCES
...
+            # For now, we are only running the adaptor based tests to keep it quick.
+            # Even if one of the adaptor tests fail, we will go ahead and create the final executable.
+            COMMAND ${CMAKE_CTEST_COMMAND} -j4 -V -R tcp || (exit 0)
+            COMMAND ${CMAKE_CTEST_COMMAND} -j4 -V -R http || (exit 0)
+            # Set the ENABLE_PROFILE_GUIDED_OPTIMIZATION=OFF so that we don't once again create profile data.
+            COMMAND ${CMAKE_COMMAND} ${CMAKE_BINARY_DIR} -DENABLE_PROFILE_GUIDED_OPTIMIZATION=OFF -DCMAKE_EXE_LINKER_FLAGS=-fprofile-use=${CMAKE_BINARY_DIR}/profile-data
+
+            # Feed the profile data we created from the above test runs using the -fprofile-use compiler option.
+            # Use the -fprofile-correction option to smoothen the counter updates when using multi-threaded applications.
+            COMMAND ${CMAKE_COMMAND} ${CMAKE_BINARY_DIR} -DC_EXTRA_GNU=-fprofile-use=${CMAKE_BINARY_DIR}/profile-data\\\;-fprofile-correction
```
jiridanek, line 53: According to a random suse blog, you are suppose to use `-⁠fprofile-partial-training` at this point, if the tests were not entirely representative of the production load, which _might_(?) be the case.

> It is important that the train run exhibits the same characteristics as the real workload. Unless you use the option -⁠fprofile-partial-training in the second build, it needs to exercise the code that is also the most frequently executed in real use, otherwise it will be optimized for size and PGO would make more harm than good. With the option, GCC reverts to guessing properties of portions of the projects not exercised in the train run, as if they were compiled without profile feedback. This however also means that the code size will not perform better or shrink as much as one would expect from a PGO build.

https://documentation.suse.com/sbp/server-linux/single-html/SBP-GCC-11/index.html#sec-gcc11-pgo

jiridanek: But the PGO build is currently running only a small portion of the tests, right? If you can be sure that "every function that needs to be fast" was executed during training, then the option is not helpful.

I don't insist on `-⁠fprofile-partial-training`, I'm only mentioning it as interesting possibility.

https://clearlinux.org/blogs-news/major-improvements-gcc-101

jiridanek: > I will add the -fprofile-partial-training since it seems to only help.

It can hurt, in the articles they say if you really have some truly unused function, adding the option will make resulting binary bigger. I guess binary size is another thing to look for when experimenting, since one benefit of PGO can be smaller binaries.

But yeah, I guess I'd turn it on too, if it was my pr ;p

jiridanek: ```\\\;``` as the option separator? I did not try myself, but I think you need that, not just a `;`

jiridanek (approved):
@ganeshmurthy lgtm, nothing to add from me

ganeshmurthy:
> @ganeshmurthy lgtm, nothing to add from me

The container build seems to have just hung. Also many test failures in the new PGO job. Not merging them until these issues are root caused.

ganeshmurthy:
> The container build seems to have just hung. Also many test failures in the new PGO job. Not merging them until these issues are root caused.

Correction. No failure in the PGO job but the Container build CI job is stuck and is spinning.

ganeshmurthy:
Many errors like this - https://github.com/skupperproject/skupper-router/actions/runs/5070497234/jobs/9105609474?pr=1044#step:3:6737

jiridanek:
> Many errors like this - https://github.com/skupperproject/skupper-router/actions/runs/5070497234/jobs/9105609474?pr=1044#step:3:6737

I'm going to retrigger that on main, if it happens there as well. I can't see old logs from a CI step that is currently running, so I see nothing on that link.

Rerun on main, https://github.com/skupperproject/skupper-router/actions/runs/5060843474/jobs/9108390401

jiridanek:
> > @ganeshmurthy lgtm, nothing to add from me
> 
> The container build seems to have just hung. Also many test failures in the new PGO job. Not merging them until these issues are root caused.

PGO in the container build runs fine with podman, https://github.com/jiridanek/skupper-router/actions/runs/5077892340/jobs/9121720794#step:3:6296

```
The following tests passed:
	system_tests_http1_over_tcp
	system_tests_tcp_adaptor_tls
	system_tests_tcp_adaptor
100% tests passed, 0 tests failed out of 3
```

If I do `docker build` on my laptop, i don't have ipv6 available, but if i do `podman build`, then ipv6 is there, https://privatebin.corp.redhat.com/?7fa1df7279df572a#6EeAxUUKGNFbc3XFcTM42i7kDJcMErcuNc39EwHWN8kC

jiridanek:
> > > @ganeshmurthy lgtm, nothing to add from me
> > 
> > 
> > The container build seems to have just hung. Also many test failures in the new PGO job. Not merging them until these issues are root caused.
> 
> PGO in the container build runs fine with podman, https://github.com/jiridanek/skupper-router/actions/runs/5077892340/jobs/9121720794#step:3:6296
> 
> ```
> The following tests passed:
> 	system_tests_http1_over_tcp
> 	system_tests_tcp_adaptor_tls
> 	system_tests_tcp_adaptor
> 100% tests passed, 0 tests failed out of 3
> ```

The GHA I linked eventually failed, but that is after the image was successfully built, because the image was built in podman but the test step afterwards expects it in docker. That would not be too hard to resolve, either by copying the image from podman to docker, or by running the test step in podman.

.github/scripts/compile.sh
```diff
@@ -106,32 +106,49 @@ do_build () {
...
 
+# This is required to install the python packages that the system tests use.
+python3 -m pip install -r "${SKUPPER_DIR}"/requirements-dev.txt
+
 # Do a regular build without asan or tsan.
 do_build "" OFF
-# and install Proton Python
+
+# Install Proton Python
 python3 -m pip install --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
```
jiridanek, line 152:
```suggestion
python3 -m pip install --ignore-installed --prefix="$PROTON_INSTALL_DIR/usr" "$(find "$PROTON_BUILD_DIR/python/" -name 'python-qpid-proton-*.tar.gz')"
```

@ganeshmurthy this needs that `--ignore-installed` flag, as I had in https://github.com/jiridanek/skupper-router/actions/runs/5072575450/jobs/9110449383

ganeshmurthy: Forgot about that, adding it now.

ganeshmurthy: After I rebased with the podman changes, the ipv6 related errors have disappeared but I see this test failure - https://github.com/skupperproject/skupper-router/actions/runs/5084198540/jobs/9136182240?pr=1044#step:3:6634

I wonder if the CI is just slow and the 60 second timeout was hit but the same test did pass in other cases.

jirkadanek: > After I rebased with the podman changes, the ipv6 related errors have disappeared but I see this test failure - https://github.com/skupperproject/skupper-router/actions/runs/5084198540/jobs/9136182240?pr=1044#step:3:6634
> 
> I wonder if the CI is just slow and the 60 second timeout was hit but the same test did pass in other cases.

Regular CI jobs have 300 second timeout or so, set by environment variable. The tests running in Containerfile build have the default 60 second limit because there the variable is not set.

I tried to rettriger the CI job to see what happens. Defining the variable also for Containerfile might make sense.

jirkadanek: https://github.com/skupperproject/skupper-router/blob/4201a1ac7df591993b0082d26fa6e801daab1634/.github/workflows/build.yaml#LL459C21-L459C21

ganeshmurthy: Are you suggesting declaring the QPID_SYSTEM_TEST_TIMEOUT in the Containerfile like this - 

 ```diff
 ENV LIBUNWIND_VERSION=v1.6.2
 ENV LWS_SOURCE_URL=${LWS_SOURCE_URL:-https://github.com/warmcat/libwebsockets/archive/refs/tags/${LWS_VERSION}.tar.gz}
 ENV LIBUNWIND_SOURCE_URL=${LIBUNWIND_SOURCE_URL:-https://github.com/libunwind/libunwind/archive/refs/tags/${LIBUNWIND_VERSION}.tar.gz}
+ENV QPID_SYSTEM_TEST_TIMEOUT=300
 
 ARG VERSION=UNKNOWN
 ENV VERSION=$VERSION
``` 

Instead of doing in the Containerfile, we could also do this - 

```diff
-            COMMAND ${CMAKE_CTEST_COMMAND} -j4 -V -R tcp || (exit 0)
-            COMMAND ${CMAKE_CTEST_COMMAND} -j4 -V -R http || (exit 0)
+            COMMAND QPID_SYSTEM_TEST_TIMEOUT=300 ${CMAKE_CTEST_COMMAND} -j4 -V -R tcp || (exit 0)
+            COMMAND QPID_SYSTEM_TEST_TIMEOUT=300 ${CMAKE_CTEST_COMMAND} -j4 -V -R http || (exit 0)
```

jiridanek: It seems to me that the extended timeout is necessary, otherwise the tests don't finish on CI, at least that's what the situation was when i last looked.

Whether to add it as Containerfile ENV or just for the command in CMakeLists.txt, I don't have a strong preference. I'd do the Containerfile ENV, but I see why the second option may look good too.

.github/scripts/compile.sh
```diff
@@ -130,6 +130,13 @@ do_build () {
   local suffix=${1}
   local runtime_check=${2}
 
+  if [ "$runtime_check" == "OFF" ]; then
+     # Turn on PGO only when we are not doing asan or tsan
+     BUILD_OPTS="-DENABLE_PROFILE_GUIDED_OPTIMIZATION=ON"
```
jiridanek, line 135: Maybe not for this PR but for a subsequent one, but it seems useful to be able to turn off PGO. Also, possibly option to turn off the sanitizers as well would come handy.

Imagine wanting to try something with the image, and waiting for PGO, LTO, and sanitizers to build makes this slow to try things and iterate.

.github/scripts/compile.sh
ganeshmurthy: > It seems to me that the extended timeout is necessary, otherwise the tests don't finish on CI, at least that's what the situation was when i last looked.
> 
> Whether to add it as Containerfile ENV or just for the command in CMakeLists.txt, I don't have a strong preference. I'd do the Containerfile ENV, but I see why the second option may look good too.

I will add the QPID_SYSTEM_TEST_TIMEOUT before the call to ctest. If there are too many tests that need using the QPID_SYSTEM_TEST_TIMEOUT, we will move it to the Containerfile

.github/scripts/compile.sh
ganeshmurthy: > Maybe not for this PR but for a subsequent one, but it seems useful to be able to turn off PGO. Also, possibly option to turn off the sanitizers as well would come handy.

Agreed !

ganeshmurthy: @jiridanek I am going to merge this to main if you don't have any further comments. Please let me know. Thx

jirkadanek: Ship it! Or, scupper it!

PR #1048, opened 2023-04-04
Fixes #1058: Track system-tests coverage in Codecov.io

### LINKED ISSUE #1058: Track system-tests coverage in Codecov.io
See https://github.com/skupperproject/skupper-router/pull/1048 for why this is ugly and horrible thing to do, but we still want this.

jiridanek:
https://app.codecov.io/github/skupperproject/skupper-router/pull/1048

Doing code coverage on system_tests is cheating. It runs a lot of code without asserting anything about the vast majority of it. So, counting all the code that was run as "covered by tests" is untrue.

The better way to measure system tests coverage is not by counting source code lines, but by counting covered user-facing features and scenarios.

@ganeshmurthy lets see what this does; I have great expectations, but maybe I'm imagining it wrong, what this is going to do

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -31,10 +31,5 @@ component_management:
           name: systemtests
           flag_regexes:
               - ".*systemtests"
-flags:
-    pysystemtests:
-        joined: false
-    systemtests:
-        joined: false
 github_checks:
     annotations: false
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -325,6 +325,8 @@ jobs:
             containerTag: stream9
             cc: gcc
             cxx: g++
+            buildType: Coverage
+            covType: system
             runtimeCheck: OFF
             protonGitRef: ${{ github.event.inputs.protonBranch || 'main' }}
             shard: 1
@@ -334,6 +336,8 @@ jobs:
             containerTag: stream9
             cc: gcc
             cxx: g++
+            buildType: Coverage
+            covType: system
             runtimeCheck: OFF
             protonGitRef: ${{ github.event.inputs.protonBranch || 'main' }}
             shard: 2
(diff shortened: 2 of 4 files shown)


jiridanek:
Looks like we can't collect python coverage while running tests in parallel, https://github.com/jiridanek/skupper-router/actions/runs/4609453728/jobs/8146605198#step:34:498

* https://github.com/nedbat/coveragepy/issues/1514

Also, we need to decide what to do with regards to coverage if tests fail, since CentOS tests currently rarely pass. There is some setting in codecov about this

```
codecov:
  require_ci_to_pass: true
```

and also the uploading github action step gets skipped if the previous step failed, so that codecov setting does not even get to play a role, currently

@ganeshmurthy thoughts?

jiridanek:
I expected some way in the codecov.io web ui to be able to show lines covered by tests uploaded under a particular flag

image

but, that does not seem to be the case, and all are lumped under https://app.codecov.io/github/skupperproject/skupper-router/commit/b2095422f08b6d4e5b564e662a819a97a3a69210/tree

It appears that I can filter out the flags in yaml, https://docs.codecov.com/docs/flags#hide-builds-eg-nightly-builds

```
flags:
  nightly:
    joined: false
```

jiridanek:
Hmm, looks like I won't be able to guess it. Let me try locally, hopefully I can put together something that works to report python coverage reliably.

Still, I don't want to indiscriminately squash systemtests and unittest coverage when showing the line-per-line coverage in codecov.io web ui...

jiridanek:
I wrote to Codecov forum about the displaying of covered lines per flag, https://community.codecov.com/t/can-i-see-line-by-line-coverage-in-file-explorer-uploaded-from-a-specific-flag-or-a-subset-of-flags/4251

jiridanek:
@ganeshmurthy ok, this should be it. first ci run failed on network issue uploading to codecov.io, so I'm rerunning it again now, so that we see the results

jiridanek:
Now the only coverage missing is what the embedded python interpreter runs from inside skrouterd. I don't consider resolving that in scope of this PR.

ganeshmurthy (approved):
Let's get this merged. Thanks for doing this.

jiridanek:
> Can you please add an issue for this and change the commit message to point to the new issue ? Thx

I'll do that, but in general, it seems fairly superfluous thing to do on GitHub, where PRs and Issues work very similarly and you can link PRs to releases, same as (or instead of) the issues.

For example, what's the point of having both https://github.com/skupperproject/skupper-router/issues/1053 and https://github.com/skupperproject/skupper-router/pull/1054 when you can link the PR to release and be done with it.

(On the other hand, issue like https://github.com/skupperproject/skupper-router/issues/1055 is helpful because it explains the background for the change. But, that can be just as well explained right in the PR comment https://github.com/skupperproject/skupper-router/pull/1056)

What does make sense, IMO, is to create issue when you don't want to immediately create the PR. But otherwise, just the PR suffices.

PR #1054, opened 2023-04-07
Fixes #1053: Use SET_ATOMIC_BOOL to set receive_complete flag on mess…

### LINKED ISSUE #1053: In qd_message_receive(), the content->receive_complete flag needs to be set with atomic function
Replace `content->receive_complete = true;` with `SET_ATOMIC_BOOL(&content->receive_complete, true);`

ganeshmurthy:
…age content


src/message.c
```diff
@@ -1580,7 +1580,7 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
                     // pending buffer is absent
                 }
 
-                content->receive_complete = true;
+                SET_ATOMIC_BOOL(&content->receive_complete, true);
```
jiridanek, line 1583: Does this clear any `tsan.supp` entry, by any chance?

I noticed

https://github.com/skupperproject/skupper-router/blob/81a196b1c94de870a0f8d85eb77a792c6ab94d76/tests/tsan.supp#L35-L36

but looking at https://issues.apache.org/jira/browse/DISPATCH-2127, it does appear that atomically setting this flag will allow getting rid of the suppression.

ganeshmurthy: Thanks @jiridanek , I will take a look

kgiusti (approved):
Provisional approval once Ganesh checks if tsan.supp update is needed.

src/message.c
ganeshmurthy: The removal of the suppression is dependent upon these three flags being accessed safely -
```
content->receive_complete
content->q2_input_holdoff
content->aborted
```

I have added additional code to this PR so that access to content->q2_input_holdoff is secured after holding `content->lock`

Let's see what the CI thinks.

ganeshmurthy: After removing the suppression, I see no new test failures on this PR. I am going to kick off the tests one more time to see what happens.

ganeshmurthy: Removing the suppression did not have any side effects. @jiridanek @kgiusti please take a look at this PR one more time and approve/comment.

src/message.c
```diff
@@ -2392,11 +2392,13 @@ qd_message_t *qd_message_compose(qd_composed_field_t *f1,
         qd_compose_free(fields[idx]);
     }
 
+    LOCK(&content->lock);
     // initialize the Q2 flag. Note that we do not need to hold the content
```
jiridanek, line 2396: The comment is out-of-date, now?

ganeshmurthy: I don't think this comment is out-of-date. I agree with the comments and am inclined to remove the lock I added. @kgiusti can you please confirm ?

kgiusti: Yeah this lock shouldn't be necessary IIUC because the message hasn't yet been shared with any other thread - it was just created in this function and we're just setting the q2_input_holdoff initial value.

Fixing the tsan warning was not the purpose of the original patch, which is to simply corrected a missing atomic operation. I'd recommend committing the original one-line change only without removing the tsan suppression since IIUC the suppression is still necessary after the one-line change is made.

ganeshmurthy: I believe that the suppression can be removed with this change. I will take out the lock that I added (since it is not necessary) and take out the suppression as well and see if this passed CI. if it does, the suppression can be safely removed

ganeshmurthy: The CI has passed. I am merging this PR to main, thanks.

PR #1060, opened 2023-04-19
Fixes #925: add Containerfile test that runs `skstat -l`

### LINKED ISSUE #925: Test for the Containerfile does not attempt to use skstat/skmanage, misses issues with Proton Python install

jiridanek:


tests/testcontainers/test_h2spec.rs
```diff
@@ -225,8 +288,69 @@ async fn test_h2spec() {
...
+}
+
+impl LogPrinter {
+    fn new() -> Self {
+        Self {
+            tasks: vec![],
+        }
+    }
+
+    fn print(&mut self, log_stream: Fuse<(impl Stream<Item=String> + Unpin + Sized + Send + 'static)>) {
```
jiridanek, line 311: yes, the type signature is kind of crazy, and even IDE plugin has some trouble with it, turns out

* https://github.com/intellij-rust/intellij-rust/issues/10363

ganeshmurthy:
I am not knowledgeable in Rust. If the skstat -l command fails for some reason, do you think the container build should fail as well ?
Besides that, feel free to commit.

ganeshmurthy (approved):

jiridanek:
> I am not knowledgeable in Rust.

Me neither. When you see me type Rust, it is just the last tutorial I gulped trying to get out, all this mediated through IntelliSense and IDE auto fixes.

And writing code is actually easier than reading it, in some cases.

> If the skstat -l command fails for some reason, do you think the container build should fail as well ?

I guess. If the image is going to be broken, better  not push it. That should be a good enhancement, yes.

* https://github.com/skupperproject/skupper-router/issues/1063

PR #1064, opened 2023-04-21
ISSUE-1059: part 1: replace tcp echo server with asyncio implementation

kgiusti:
This patch replaces the current non-blocking TCP echo server with a much simpler asyncio based implementation as recommended by the python docs (see the see-also block):

https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets

This fixes the TCP test failures that involved SSL with the echo server. This does not fix the issues with the nginx TLS tests. That will need a followup fix.

### DIFF
--- a/tests/TCP_echo_server.py
+++ b/tests/TCP_echo_server.py
@@ -19,10 +19,13 @@
 # under the License.
 #
 
+#
+# A simple TCP echo server that supports TLS connections
+#
+
 import argparse
-import selectors
+import asyncio
 import signal
-import socket
 import sys
 import ssl
 import time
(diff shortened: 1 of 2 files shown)


ganeshmurthy:
@kgiusti can you please add back the log messages from the old version so that the new echo server puts out similar log messages like the old version ? The older log messages were very useful when debugging the echo server

kgiusti:
@ganeshmurthy the new async implementation drops a lot of code that is no longer necessary, so there are some log messages that no longer apply.

The new impl still logs all the protocol events including data traces - is there anything specific that I missed? Lemme know, thanks.

ganeshmurthy (approved):

Issue #973, opened 2023-03-02
InvalidVersion error thrown during Python package  build: version number is not PEP-440 compliant


kgiusti:
Router build issues the following very scary message:

/home/kgiusti/.local/lib/python3.11/site-packages/setuptools/_normalization.py:93: SetuptoolsDeprecationWarning: Invalid version: '2.1.0-149-g45d75bbf'.
        !!
        ###################
        # Invalid version #
        ###################
        '2.1.0-149-g45d75bbf' is not valid according to PEP 440.
 
        Please make sure specify a valid version for your package.
        Also note that future releases of setuptools may halt the build process
        if an invalid version is given.
        
 
!!
  warnings.warn(cleandoc(msg), SetuptoolsDeprecationWarning)

See https://peps.python.org/pep-0440/#version-scheme

Need to reformat the version to something compliant: maybe 2.1.0.dev149+g45d75bbf ??

jiridanek:
also, https://github.com/jiridanek/skupper-router/actions/runs/4514232572/jobs/7950069120#step:25:661

```
/usr/lib/python3.11/site-packages/pkg_resources/__init__.py:123: PkgResourcesDeprecationWarning: UNKNOWN is an invalid version and will not be supported in a future release
```

jiridanek:
and, in copr rpm builds on fedora 38 and rawhide (but not on fedora 37)

```
Processing files: skupper-router-0.0.git.3973.71ba5332-2.20230328192824435834.pr1014.172.g71ba5332.fc38.s390x
Traceback (most recent call last):
  File "/usr/lib/rpm/pythondistdeps.py", line 622, in <module>
    main()
  File "/usr/lib/rpm/pythondistdeps.py", line 481, in main
    if RpmVersion(version).is_zero():
       ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/rpm/pythondistdeps.py", line 135, in __init__
    version = parse(version_id)
              ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/packaging/version.py", line 52, in parse
    return Version(version)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '0.0.git.3973.71ba5332'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #1089, opened 2023-05-23
Fixes #1088: Use the docker tag as the skupper-router version

### LINKED ISSUE #1088: Use the DOCKER_TAG as the skupper-router version in image-manual.yml
Currently skupper-router release candidate builds are done using image-manual.yml. The skupper-router version gets set to the github commit sha. Set the skupper-router version to the same value as the image version.

ganeshmurthy:


.github/workflows/image-manual.yml
```diff
@@ -56,8 +56,8 @@ jobs:
       - run: |
           # build, run tests and if all worked publish a custom image
           export PROJECT_TAG="${DOCKER_TAG}"
-          # Use {GITHUB_SHA::8} for shorter SHA
-          export VERSION="${GITHUB_SHA}"
+          # The version of skupper-router should be the same as the version of the image.
+          export VERSION="${DOCKER_TAG}"
```
jiridanek, line 60: It would be nice IMO to keep first 6 or 8 characters of the commit hash in the version. But that is for separate enhancement. It might also clash with Python versioning, so https://github.com/skupperproject/skupper-router/issues/973 would need to be looked into at the same time

jiridanek (approved):
afaict looks like what @ted-ross asked for!

> i.e. why doesn't the version say "1.4.0-rc1"?

✅ 

> we could use the official process for the RC but the official process does a bunch of things like push the :latest tag, create release notes, create a github release (which sends emails) etc. So, we will have to add a bunch of if statement around the official release code

@ganeshmurthy we could split the https://github.com/skupperproject/skupper-router/blob/main/.github/workflows/release.yml into two jobs, one job builds image, the other one pushes release. And then for rebuilds, you'd only rerun the image job...

ganeshmurthy:
> afaict looks like what @ted-ross asked for!
> 
> > i.e. why doesn't the version say "1.4.0-rc1"?
> 
> white_check_mark
> 
> > we could use the official process for the RC but the official process does a bunch of things like push the :latest tag, create release notes, create a github release (which sends emails) etc. So, we will have to add a bunch of if statement around the official release code
> 
> @ganeshmurthy we could split the https://github.com/skupperproject/skupper-router/blob/main/.github/workflows/release.yml into two jobs, one job builds image, the other one pushes release. And then for rebuilds, you'd only rerun the image job...

With regards to the image creation part, we should be able to modify the current release.yaml that can be shared by the regular release and rc release. When I push a tag that is of the form x.y.z or x.y.z-rc(n) an image must be *automatically* pushed to quay.io that that matches the tag. For a regular release (without rc in in its tag, the "Create draft release" part should be executed and not executed if rc is in the tag.

PR #1090, opened 2023-05-23
ISSUE-1083: add CI testcase for panic handler

kgiusti:


include/qpid/dispatch/router.h
```diff
@@ -141,6 +141,11 @@ qd_router_forwarder_t *qd_router_get_forwarder(qd_address_treatment_t t);
 /** Retrieve the routers current memory usage (in bytes) */
 uint64_t qd_router_memory_usage(void);
 
+/** Return true if the router is running in test mode. Test mode can be enabled by passing the '-T' command line option
+ * to the router on startup.
+ */
+bool qd_router_test_mode_enabled(void);
```
jiridanek, line 147: naming nit:

I like the old name better. The `qdr_core_test_hooks_enabled` suggests that there are some additional "hooks" that can be used to poke the router in tests.

The new name `qd_router_test_mode_enabled` suggests that the router itself somehow [works differently](https://jalopnik.com/understanding-the-test-mode-that-let-vw-trick-the-emiss-1732175835), that it has a test mode behavior and nontestmode behavior. And that is not true. Even the "crash me" config is actually a hook like the others.

tests/system_tests_panic_handler.py
```diff
@@ -0,0 +1,113 @@
...
+        regex = re.compile(r"^Process ID: (\d+) (\S+)$", re.MULTILINE)
+        match = regex.search(crash_dump)
+        self.assertIsNotNone(match, "failed to find Process ID:")
+
+        regex = re.compile(r"^!!! libunwind not present: backtrace unavailable !!!$",
+                           re.MULTILINE)
+        if (regex.search(crash_dump) is None):
+            # libunwind present, expect at least one stack frame:
+            regex = re.compile(r"^\[(\d+)\] IP: ", re.MULTILINE)
+            self.assertTrue(len(regex.findall(crash_dump)) > 0,
```
jiridanek, line 104: this could be `self.assertGreater`, just so that failure message is nicer

jiridanek: or assertRegex after all, https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex

src/adaptors/http1/http1_client.c
```diff
@@ -266,6 +266,16 @@ static void _handle_listener_accept(qd_adaptor_listener_t *adaptor_listener, pn_
 //
 qd_http_listener_t *qd_http1_configure_listener(qd_http_listener_t *li, qd_dispatch_t *qd, qd_entity_t *entity)
 {
+    if (qd_router_test_mode_enabled()) {
```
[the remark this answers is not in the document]
kgiusti: Couple 'o reasons why I did it this way:
1) test_adaptor.c doesn't have a management entity, so it cannot be triggered at runtime
2) having an adaptor that immediately crashes the router on startup doesn't stress the traceback as much as having it crash deep in a thread when multi-threading is active.

tests/system_tests_panic_handler.py
kgiusti: changed to assertRegex

ganeshmurthy (approved):

ganeshmurthy:
@jiridanek can you please approve this PR if your concerns have been addressed? We want to get this into main branch ASAP. thanks

PR #1098, opened 2023-05-26
Fixes #1096: Additional fix. Export BUILDAH_FORMAT=docker

### LINKED ISSUE #1096: Use podman instead of docker when creating and testing container images. We currently use docker

ganeshmurthy:


jiridanek (approved):
It's the Red Hat endorsed solution, https://access.redhat.com/solutions/6843481

I don't have docker 1.13 at hand to check, but it looks like it will do the work!

ganeshmurthy:
> It's the Red Hat endorsed solution, https://access.redhat.com/solutions/6843481
> 
> I don't have docker 1.13 at hand to check, but it looks like it will do the work!

Thanks. I hope it works. Keeping my fingers crossed. Merging now

PR #1099, opened 2023-05-26
Moved dataConnectionCount field from connector to router

ganeshmurthy:


ganeshmurthy:
This PR has a couple of bug and some test failures, more work needed but this is a first draft

PR #1100, opened 2023-05-27
Issue #1367: Build arm64/v2 (aarch64) skupper-router docker image

jiridanek:
The PR is a partial implementation for issue #1367 because it does not look into publishing the image and writing the multiarch manifest file into image registry. This is something I intend to do later if the first part of the approach chosen is accepted.

---

There are three ways to build multiarch images

* qemu (easy, slow)
* remote build (harder, fast, needs extra machines)
* cross-compilation (hard, fast)

https://blog.kubesimplify.com/the-secret-gems-behind-building-container-images-enter-buildkit-and-docker-buildx#heading-demo

## Current description

This PR does something in between the qemu-user and cross-compilation. It runs a self-contained (static) host-architecture compiler binary `zig cc` in a cross-compiling mode inside the target-architecture image. It is strange, but it allows to keep most of the changes to `compile.sh` self-contained in an if-branch, compared to qemu-user it is significantly faster, and compared to full cross-compilation it allows me to brush over most of the difficulties (because there is only one system root and not two (host and target), and qemu-user steps in every time I need to run target binaries, so I can directly use those).

## Outdated description, for the first iteration of the PR

This PR tries to go the qemu route, just to feel out the stones while crossing a river.

The arm64 build in qemu takes _ages_. It shows that it is not practical here. Native amd64 build is 7.5 min, arm64 build in qemu is 1.25 hours.

We could use TravisCI, which does have the needed architectures natively, but then there are problems in uploading the images to quay and writing an overall image manifest. (This PR does not yet try to build manifest.)

Or we could do the cross-compilation. I think the decision should depend on what is easiest for downstream Red Hat building.

The Artemis folks (@brusdev and others) in https://github.com/artemiscloud/activemq-artemis-broker-image use https://docs.cekit.io/en/latest/ to build images. They do release multiarch container images already. Maybe that is a viable option here too?

The Strimzi project seems to support multiarch images too https://github.com/strimzi/strimzi-kafka-operator/issues/3357, they have https://github.com/strimzi/strimzi-kafka-operator/blob/d34799cb18d5d1fc5d2f18f3c7b3a85fa346cc18/development-docs/DEV_GUIDE.md?plain=1#L385 and the way the Dockerfile works is that it downloads a previously published release for that particular platform, so there is no compiling https://github.com/strimzi/strimzi-kafka-operator/blob/d34799cb18d5d1fc5d2f18f3c7b3a85fa346cc18/docker-images/kafka-based/kafka/Dockerfile#L40

The skupper PR from @fgiorgetti used cross-compilation combined with qemu (to run microdnf, etc. on $TARGETPLATFORM. Cross-compiling go is easier than cross-compiling C. https://github.com/skupperproject/skupper/pull/1064

The `zig cc` command bundles clang in a way that simplifies its usage for cross compilation, https://zig.news/kristoff/cross-compile-a-c-c-project-with-zig-3599

For CMake, I can imagine an approach which involves installing a cross compiler and mounting arm64 docker image to provide arm64 environment with libraries. https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html


jiridanek:
Thinking about it, the idea of building separately and reusing the artifacts in Containerfile build (the Strimzi way, iiuic) seems really nice. In fact, it is not all that much different from the use of a multistep container build with a `build` container, which skupper-router did from the start.

In upstream, the build for arm64 could happen in packit (https://github.com/skupperproject/skupper-router/issues/1013) and the container build would only have to wait for the RPMs to become available and then install them. I'm not sure if EPEL 9 build profile in COPR corresponds to ubi9, though. If not, then only other option is Travis and some sort of custom artifact hosting, because Travis does not have the artifact storage feature.

My point is that if non-PGO arm64 build under qemu takes 1.25 h, the PGO build will take 3 times as long and there will be problems with test timeouts. So qemu is clearly not the way. And crosscompilation inside container is not easy, and again, for PGO, the tests have to run natively, which means in this case qemu.

Who knows if qemu pgo data are even relevant, or if this skewes the measurement in some way. The same code is executed, but timing is certainly different.

jiridanek:
One cause of the CI fail is that the router is too slow to reopen listening socket (because qemu, no doubt) and the h2spec test does not do retries

```
skrouterd: 2023-05-27 18:14:54.159385 +0000 ROUTER (info) Listener httpListener/0.0.0.0:24162: stopped listening for client connections on 0.0.0.0:24162 (/build/src/adaptors/adaptor_listener.c:170)
h2spec:          [recv] HEADERS Frame (length:69, flags:0x04, stream_id:1)
h2spec:          [recv] DATA Frame (length:1, flags:0x00, stream_id:1)
skrouterd: 2023-05-27 18:14:54.389588 +0000 HTTP_ADAPTOR (info) [C1] Activation triggered, no socket yet so scheduled timer (/build/src/adaptors/http2/http2_adaptor.c:2075)
h2spec: 
h2spec:     ✔ 3: Sends a PRIORITY frame on half-closed (remote) stream
h2spec: 
h2spec:     × 4: Sends a RST_STREAM frame on half-closed (remote) stream
h2spec: 
h2spec: Error: dial tcp 10.89.0.3:24162: connect: connection refused
skrouterd: 2023-05-27 18:14:54.859793 +0000 ROUTER_CORE (info) [C5][L9] Link closed due to connection loss: del=1 presett=0 psdrop=0 acc=0 rej=0 rel=0 mod=0 delay1=1 delay10=0 blocked=no (/build/src/router_core/connections.c:1130)
skrouterd: 2023-05-27 18:14:54.860014 +0000 ROUTER_CORE (info) [C5][L10] Link closed due to connection loss: del=0 presett=0 psdrop=0 acc=0 rej=0 rel=0 mod=0 delay1=1 delay10=0 blocked=no (/build/src/router_core/connections.c:1130)
skrouterd: 2023-05-27 18:14:54.860154 +0000 ROUTER_CORE (info) [C5] Connection Closed (/build/src/router_core/connections.c:1808)
skrouterd: 2023-05-27 18:14:54.926672 +0000 FLOW_LOG (info) FLOW [pZiZ+:10] BEGIN END parent=pZiZ+:9 counterflow=pZiZ+:11 octets=0 latency=7[305](https://github.com/skupperproject/skupper-router/actions/runs/5099845385/jobs/9167791331?pr=1100#step:8:306)368 method=GET octetRate=0 streamId=1 (/build/src/vanflow.c:773)
skrouterd: 2023-05-27 18:14:54.926917 +0000 FLOW_LOG (info) FLOW [pZiZ+:9] END parent=pZiZ+:1 sourceHost=10.89.0.14 sourcePort=34776 octets=0 octetRate=0 windowSize=65536 (/build/src/vanflow.c:773)
skrouterd: 2023-05-27 18:14:55.134791 +0000 HTTP_ADAPTOR (info) [C1] Connecting to nghttpd:8888 (/build/src/adaptors/http2/http2_adaptor.c:2092)
skrouterd: 2023-05-27 18:14:56.000356 +0000 ROUTER_CORE (info) [C1][L13] Link attached: dir=out source={examples expire:link} target={<none> expire:link} (/build/src/router_core/connections.c:2132)
skrouterd: 2023-05-27 18:14:56.001085 +0000 ROUTER (info) Listener httpListener/0.0.0.0:24162: listening for client connections on 0.0.0.0:24162 with backlog 4096 (/build/src/adaptors/adaptor_listener.c:139)
```

https://github.com/skupperproject/skupper-router/actions/runs/5099845385/jobs/9167791331?pr=1100#step:8:298

jiridanek:
Docker/Podman can be used to hook in qemu, alternatively there is Fedora package https://packages.fedoraproject.org/pkgs/qemu/qemu-user-static/

```
      - name: Enable qemu to build foreign architectures
        if: matrix.platform != 'linux/amd64/v2'
        run: sudo podman run --privileged --rm docker.io/tonistiigi/binfmt --install all
```

.github/workflows/build.yaml
```diff
@@ -899,9 +904,14 @@ jobs:
         env:
           HOMEBREW_NO_INSTALL_CLEANUP: 1
 
+      - name: Enable qemu to build foreign architectures
+        if: matrix.platform != 'linux/amd64/v2'
```
jiridanek, line 908:
```suggestion
        if: matrix.platform != 'linux/amd64'
```

router/CMakeLists.txt
```diff
@@ -31,10 +31,17 @@ set(router_SOURCES
...
+
 add_executable(skrouterd ${router_SOURCES})
 target_include_directories(skrouterd PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(skrouterd PRIVATE skupper-router)
-target_link_options(skrouterd PUBLIC LINKER:-Map=skrouterd.map)
+if (linker_map_supported)
+    target_link_options(skrouterd PUBLIC LINKER:-Map=skrouterd.map)
+else ()
+    message(WARNING "Linker map generation not supported, skrouterd.map will not be written out")
+endif ()
```
jiridanek, line 44: * https://github.com/ziglang/zig/issues/18356

ganeshmurthy:
> Regarding functionality, the image works on my raspberry pi 4 as far as I can tell, with the exception of this known issue

How does this libunwind issue manifest itself when running the router ?

jiridanek:
> How does this libunwind issue manifest itself when running the router ?

I've only seen it crash when backtrace() is called. Because libunwind replaces the glibc function backtrace() when libunwind is present on linker path.

I imagine this will also affect the signal handler that Ken created. So if you crash the router, then it will crash again trying to use libunwind to print crash backtrace.

Anyways, the PR does not publish the image anywhere, so it is not like I am asking anyone to publish known-buggy image to quay, or anything.

ganeshmurthy:
> Anyways, the PR does not publish the image anywhere, so it is not like I am asking anyone to publish known-buggy image to quay, or anything.

From you comment in a related ticket, it looks like libunwind 1.8.0rc1 fixes the crash in aarch64 platform ? 
When libunwind 1.8.0 is released (which I am hoping is soon), we will have to update the version of libunwind in the Containerfile to see if that new version is working for both amd64 and aarch64.

That being said, is https://github.com/skupperproject/skupper-router/pull/1339 still applicable ? It is in the Draft state, so I was wondering what is up with that. 
 
You can go ahead and check this PR in as long as this does not fail the regular CI run.

jiridanek:
* https://github.com/skupperproject/skupper-router/pull/1339

That might not be a bad thing to merge, but that's just "keeping up with upstreams". It will not help any with the arm64 crash.

ganeshmurthy:
> That might not be a bad thing to merge, but that's just "keeping up with upstreams". It will not help any with the arm64 crash.

I just remembered @kgiusti trying this PR and having some trouble with the new version. https://github.com/skupperproject/skupper-router/pull/1339 will need to wait for his approval.

PR #1109, opened 2023-06-01
Fixes #1108: Compile libunwind from source.

### LINKED ISSUE #1108: Modify Containerfile and compile.sh to build libunwind from source

ganeshmurthy:


.github/scripts/compile.sh
```diff
@@ -31,11 +31,14 @@ if [ -z "${REMOTE_SOURCES_DIR:-}" ]; then
...
   LWS_DIR="$WORKING_DIR/libwebsockets/app"
+  LIBUNWIND_DIR="$WORKING_DIR/libunwind/app"
   # $REMOTE_SOURCES_DIR was not provided, we will have to download the libwebsockets source from ${LWS_SOURCE_URL}
   # Get the libwebsockets source into a libwebsockets.tar.gz file
   # and untar it into the libwebsockets folder
   wget "${LWS_SOURCE_URL}" -O libwebsockets.tar.gz
-  tar -zxf libwebsockets.tar.gz --one-top-level="${LWS_DIR}" --strip-components 1
+  wget "${LIBUNWIND_SOURCE_URL}" -O libunwind.tar.gz
+  tar -zxf libwebsockets.tar.gz --one-top-level="${LWS_DIR}"     --strip-components 1
+  tar -zxf libunwind.tar.gz     --one-top-level="${LIBUNWIND_DIR}" --strip-components 1
```
jiridanek, line 41: maybe also vertically align the `--strip-components` part, when you are doing the aligning? 🤷🏽

jiridanek (approved):
looks ok to me, but I'd still go with DESTDIR, as suggested inline

.github/scripts/compile.sh
```diff
@@ -88,6 +96,23 @@ DESTDIR="${LWS_INSTALL_DIR}" cmake --install "${LWS_BUILD_DIR}"
 tar -z -C "${LWS_INSTALL_DIR}" -cf /libwebsockets-image.tar.gz usr
 #endregion libwebsockets
 
+#region libunwind
+pushd "${LIBUNWIND_DIR}"
+autoreconf -i
+./configure
+# This make install will install the libunwind in /usr/local/lib
+# and the include files in /usr/local/include and when skupper-router is compiled
+# in the subsequent steps, it can find the libunwind related files in /usr/local
```
jiridanek, line 105: The LWS above does exactly the same thing, installing into two places, first place so it can be used when compiling, and second place so that it can be tarred. It would make more sense to place these extensive comments on the first use of this trick (for libwebsocket), not on the second use here.

Containerfile
```diff
@@ -19,27 +19,29 @@
...
 ENV LWS_SOURCE_URL=${LWS_SOURCE_URL:-https://github.com/warmcat/libwebsockets/archive/refs/tags/${LWS_VERSION}.tar.gz}
+ENV LIBUNWIND_SOURCE_URL=${LIBUNWIND_SOURCE_URL:-https://github.com/libunwind/libunwind/archive/refs/tags/${LIBUNWIND_VERSION}.tar.gz}
 
 ARG VERSION=UNKNOWN
 ENV VERSION=$VERSION
 RUN .github/scripts/compile.sh
-RUN tar zxpf /qpid-proton-image.tar.gz --one-top-level=/image && tar zxpf /skupper-router-image.tar.gz --one-top-level=/image && tar zxpf /libwebsockets-image.tar.gz --one-top-level=/image
+RUN tar zxpf /qpid-proton-image.tar.gz --one-top-level=/image && tar zxpf /skupper-router-image.tar.gz --one-top-level=/image && tar zxpf /libwebsockets-image.tar.gz --one-top-level=/image && tar zxpf /libunwind-image.tar.gz --one-top-level=/image
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
```
kgiusti, line 47: Q: Can you remove the install of the epel-release rpm here as well?  IIRC I added that as part of the libunwind inclusion. It may no longer be necessary (?)

jirkadanek: +1, I thought I saw both epel and lib unwind mentions removed in both places, but clearly I was not attentive enough. It needs to go.

ganeshmurthy:
@jiridanek @kgiusti can you please take a quick look at the most recent commit I made to this PR and let me know if all is good ? thx

jiridanek:
> @jiridanek @kgiusti can you please take a quick look at the most recent commit I made to this PR and let me know if all is good ? thx

lgtm, thanks

PR #1110, opened 2023-06-02
Fixes #1107 - Use a tiered list of read-grant amounts based on percen…

### LINKED ISSUE #1107: Feature - Cap the memory usage for socket buffers

ted-ross:
…tage of buffer capacity that's in use.

### DIFF
--- a/include/qpid/dispatch/alloc_pool.h
+++ b/include/qpid/dispatch/alloc_pool.h
@@ -111,7 +111,8 @@ static inline void *qd_alloc_deref_safe_ptr(const qd_alloc_safe_ptr_t *sp)
     void free_##T(T *p); \
     typedef qd_alloc_safe_ptr_t T##_sp; \
     void set_safe_ptr_##T(T *p, T##_sp *sp); \
-    T *safe_deref_##T(T##_sp sp)
+    T *safe_deref_##T(T##_sp sp); \
+    qd_alloc_stats_t *alloc_stats_##T(void)
 
 /**
  * Define allocator configuration.
--- a/src/adaptors/adaptor_common.c
+++ b/src/adaptors/adaptor_common.c
@@ -102,6 +102,10 @@ void qd_adaptor_common_init(void)
 
 int qd_raw_connection_grant_read_buffers(pn_raw_connection_t *pn_raw_conn)
 {
+    //
+    // Define the allocation tiers.  The tier values are the number of read buffers to be granted
+    // to raw connections based on the percentage of usage of the router-wide buffer ceiling.
+    //
 #define TIER_1 8  // [0% .. 50%)
 #define TIER_2 4  // [50% .. 75%)
 #define TIER_3 2  // [75% .. 85%)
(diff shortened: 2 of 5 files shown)


src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -2234,6 +2234,7 @@ static void qdr_tcp_activate_CT(void *notused, qdr_connection_t *c)
  */
 static void qdr_tcp_adaptor_init(qdr_core_t *core, void **adaptor_context)
 {
+    qd_adaptor_common_init();
```
ganeshmurthy, line 2237: Looks like we are doing this tiered approach only in the tcp adaptor, just making sure ?

include/qpid/dispatch/alloc_pool.h
```diff
@@ -111,7 +111,8 @@ static inline void *qd_alloc_deref_safe_ptr(const qd_alloc_safe_ptr_t *sp)
     void free_##T(T *p); \
     typedef qd_alloc_safe_ptr_t T##_sp; \
     void set_safe_ptr_##T(T *p, T##_sp *sp); \
-    T *safe_deref_##T(T##_sp sp)
+    T *safe_deref_##T(T##_sp sp); \
+    qd_alloc_stats_t *alloc_stats_##T(void)
```
ganeshmurthy, line 115: It would be nice if we could call this `get_alloc_stats_##T` instead of `alloc_stats_##T`

kgiusti:
I think it is safe to suppress the race error, since this doesn't require synchronous access to the held_in_threads counter:

https://github.com/skupperproject/skupper-router/actions/runs/5158636146/jobs/9292547028?pr=1110#step:34:5156

src/adaptors/adaptor_common.c
```diff
@@ -72,24 +78,70 @@ qd_error_t qd_load_adaptor_config(qd_adaptor_config_t *config, qd_entity_t *enti
...
+#define TIER_2 4  // [50% .. 75%)
+#define TIER_3 2  // [75% .. 85%)
+#define TIER_4 2  // [85% .. 100%]
+
     assert(pn_raw_conn);
     pn_raw_buffer_t raw_buffers[RAW_BUFFER_BATCH];
-    size_t          desired = pn_raw_connection_read_buffers_capacity(pn_raw_conn);
-    const size_t    granted = desired;
+    size_t desired = TIER_4;
+    size_t capacity = MIN(pn_raw_connection_read_buffers_capacity(pn_raw_conn), TIER_1);
```
kgiusti, line 113: There may be a problem with this that results in granting more buffers than desired. (or I may be missing something obvious).

If this function is called multiple times before proton actually consumes the buffers I think that the raw connection will be filled to the maximum supported by the raw connection (16 currently) regardless of the memory pressure.

In the TCP adaptor this routine is called every time the delivery continue event is handled.  I've tried setting SKUPPER_MEMORY_CEILING to 65536 which should be hit at 4 buffers, but when I run the CI test it appears as if the raw connection is always getting 16 read buffers anyway.

The problem is we don't have an API to query how many rx bufs the raw connection already has on entry.  We could assume 16 since that's the current raw connection impl.

ted-ross: @kgiusti I believe you are correct here.  To do this right, we need to know either the number of granted read buffers or the maximum number of read buffers.

If we set TIER_1 to 16, I think it would work correctly by virtue of the coincidence that TIER_1 == max_buffers.

src/adaptors/adaptor_common.c
```diff
@@ -72,28 +78,110 @@ qd_error_t qd_load_adaptor_config(qd_adaptor_config_t *config, qd_entity_t *enti
     return qd_error_code();
 }
 
+void qd_adaptor_common_init(void)
+{
+    if (buffer_ceiling > 0) {
+        return;
+    }
+
+    char     *ceiling_string = getenv("SKUPPER_MEMORY_CEILING");
```
ganeshmurthy, line 87: Should this be called `SKUPPER_ROUTER_MEMORY_CEILING` instead of `SKUPPER_MEMORY_CEILING`

ganeshmurthy (approved):

src/adaptors/adaptor_common.c
ted-ross: This is now fixed.

src/adaptors/adaptor_common.c
ted-ross: Yes, probably.

include/qpid/dispatch/alloc_pool.h
ted-ross: Make this a separate issue.  It was already there, I just added it to the declaration.

Issue #910, opened 2023-01-20
Require Python 3.9 (for alignment with Qpid Proton Python)


jiridanek:
Qpid Proton Python needs Python 3.8 as minimum. It makes sense to require Python 3.9 minimum in the Router.

PR #1115, opened 2023-06-06
Set proton version in CI to 0.39.0

ganeshmurthy:


jiridanek:
* Looks like previous update https://github.com/skupperproject/skupper-router/pull/851, so it will be probably correct.

jiridanek:
Also, this

* https://github.com/skupperproject/skupper-router/issues/1116

but I'd be willing to bet it is just revealing some previously existing issue, most likely in tests themselves.

ganeshmurthy:
> Also, this
> 
>     * [ AutoLinkRetryTest.test_auto_link_reattach: AssertionError: 'attaching' != 'failed' #1116](https://github.com/skupperproject/skupper-router/issues/1116)
> 
> 
> but I'd be willing to bet it is just revealing some previously existing issue, most likely in tests themselves.

I am spending some time today looking at some intermittent errors, I will look at this one.

ganeshmurthy:
> > > * Looks like previous update [Fixes #850: Set the proton version to 0.38.0 in build.yaml and CMakeL… #851](https://github.com/skupperproject/skupper-router/pull/851), so it will be probably correct.
> > 
> > 
> > There is a failure here - https://github.com/skupperproject/skupper-router/actions/runs/5189041316/jobs/9353483797?pr=1115#step:8:92
> 
> @ganeshmurthy Uh, I can't read error messages.
> 
> The relevant error message part is
> 
> ```
> CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Python: Found unsuitable version "3.6.8", but required is at
>   least "3.8" (found /usr/bin/python3, found components: Interpreter
>   Development Development.Module Development.Embed)
> ```
> 
> and that comes from new Proton wanting newer Python. We can either use python3.9 in the RPM package, or build the rpm package for RHEL 9. There is PR for the first solution which would need to be rebased and merged
> 
>     * [Issue #910, #48, #102 - Require Python 3.9 in RPM package #398](https://github.com/skupperproject/skupper-router/pull/398)

I *feel* like going the RHEL 9 route (since this is upstream) is the right thing to do, what do you think ?

jiridanek:
> I feel like going the RHEL 9 route (since this is upstream) is the right thing to do, what do you think?

There's nothing wrong in having a RPM that works on RHEL 8 (and variants) as well. It is a bit more work, but the work will be useful to anyone not yet on RHEL 9.

Trying to solve RHEL 7 here would be out of place, I agree. RHEL 8 is fine.

If you look at the RPM PR, you'd see that the change being made also removes the schizophrenic situation where router uses self-compiled proton while router tools use packaged qpid-proton-python. That's another good reason (IMO) to go with the PR.

ganeshmurthy:
> And in that case it might make sense to jump to RHEL 9 straight away.

The other reason I prefer rhel 9 is because this is upstream and we want to generate rpms for the latest rhel. 
I am thinking of merging this PR and you can go ahead and change the rpm generation PR to use rhel 9 ?

jiridanek:
> The other reason I prefer rhel 9 is because this is upstream and we want to generate rpms for the latest rhel.

Rhel is an exception from the "stay on head" upstream rule, I think.

> I am thinking of merging this PR and you can go ahead and change the rpm generation PR to use rhel 9 ?

Fine with me

jiridanek:
> > And in that case it might make sense to jump to RHEL 9 straight away.
> 
> The other reason I prefer rhel 9 is because this is upstream and we want to generate rpms for the latest rhel. I am thinking of merging this PR and you can go ahead and change the rpm generation PR to use rhel 9 ?

* Rhel 9 fix in https://github.com/skupperproject/skupper-router/pull/1120

PR #1125, opened 2023-06-07
Adding build number to published images

fgiorgetti:


fgiorgetti:
> +1, but thinking beyond this PR, there really should be some release steps checklist in https://github.com/skupperproject/skupper-router/tree/main/docs/notes because it is getting complicated.

Release instructions have been updated. @ganeshmurthy @jiridanek Could you please validate?

ganeshmurthy (approved):

RELEASE.adoc
```diff
@@ -57,5 +63,6 @@ under the License
   ** Create a test image using the "Run Workflow" dropdown in the "Manual skupper-router image build"
      page from the Actions tab in github. Make sure the "Push :latest tag?" checkbox is unchecked while creating this test re-spin.
   ** Once the test image is created, use it in skupper-ci and make sure all CI tests pass.
-* Once the skupper CI tests are successful, go back to the "Manual skupper-router image build" page and choose the
-  correct tag (same tag from earlier release, for e.g. 2.3.0) and check the "Push :latest tag?"
+* Once the skupper CI tests are successful, go to the *https://github.com/skupperproject/skupper-router/actions/workflows/release.yml[Release skupper-router]*
+  action page and find the workflow run for the respective tag
```
jiridanek, line 67: for the future: if I understand correctly how GitHub deployments work, this could be implemented as a deployment. The first part of the release workflow would propose the release and the release candidate, and when the "deployment" is then approved (after Skupper CI tests pass with it), it can proceed to push :latest. No rebuild necessary.

If I find an opportunity, I'll try to look into doing this for the future.

https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments

PR #1129, opened 2023-06-09
Fixes #1128 - Development of next-generation TCP adaptor.

### LINKED ISSUE #1128: Cut-through TCP adaptor
This issue tracks a new feature:

This next-generation TCP adaptor has the following attributes:
- It does not use adaptor-buffers and has no buffer copying - it uses qd-buffers in the raw connections.
- It introduces a simpler, more efficient, and lockless through-message buffer flow for unicast streams.
- It does not use the core thread for intermediate frames in a message flow.
- It uses the underlying AMQP framing to delimit stream segments.  It does not use vbin structures in the stream-body.

ted-ross:
Note that this commit occludes/replaces the legacy tcp-adaptor module.

Summary of squashed commits:

- Added new adaptor module and a munch of boilerplate stubs for later development. Listeners and connectors now get core connections and links set up. Preparatory refactoring
- Added the new message/stream API for unicast/cut-through WIP - listener side is implemented up to message stream API WIP - Added connector-side raw-connection establishment. WIP - Added handling of no-server error case.
- WIP - Simplifications to the message headers, implemented server stream Feature: user-streaming links for manual streaming in adaptors WIP - Refactored clinet/server to inbound/outbound to take advantage of symmetries WIP - implemented lifecycle of connections in both LSIDE and CSIDE. WIP - even cleaner version of full connection lifecycle - still leaks messages WIP - fixed the delivery/message leak
- WIP - Added connection cleanup at router shutdown
- WIP - Completed activation for raw connections - one-router transfer tested WIP - Added vFlow metrics for octets and latency
- Added vFlow attribute generation.  Removed debug printf statements. Improved listener and connector deletion:  clean up connections and end the vflow record. Add test-and-set conditions for activation of core, cut-through-inbound, and cut-through-outbound Separate typedef of qd_connection_t into a new header file to eliminate unresolved cross dependencies. Minor updates in preparation for new work.
- Implemented the outgoing cutthrough flow for AMQP links. WIP - incoming cut-through on AMQP side
- Use a null-body to delimit the beginning of the data stream. End-to-end TCP in multi-router cases now operates.  Lots of corner cases to clean up. Fixed tcp connection close issues.  Added more trace logs. Added operStatus updates for listeners.
- Moved to core_connection per tcp_connection to remove mutex issues.  Still need to convert some wakes to direct calls. Removed unimplemented and unused call: qdr_link_delete Major updates to activation and read flow.  Many tests pass. Fixed a bug with cut-through send not handling short-duration connections. Fixed issues with long streaming content.
- Fixed a buffer leak for long streaming messages.  Fixed a delivery leak. Added more stats for listeners and connectors.
- Added a wrapper in "threading" for posix spin-locks. Refactored activation - works much better.  Still a delivery leak to be fixed. Logging improvements
- Fixed delivery leak.
- WIP - Refactoring to provide protection during activation. Stabilized the single-router case.
- Added delayed close for tcplite_connections
- Free all resources on deletion of listeners and connectors. Close window around the raw_opened flag
- Quieted the logs.  Added a pointer check before extracting error conditions. Remove use of spinlocks in the adaptor.
- Recover from Q3-stall
- Hold off streaming until there is a counterflow stream. Convert spinlocks to adaptive spin locks for better user-space performance. Threading/execution cleanup; logging cleanup
- Added assertions for thread-types.
- General suffix and comment cleanup
- Added 'notify' to cutthrough notification function names.  Removed redundant calls to them.  These changes were a result of feedback from team review.
- Properly clean up deliveries on error-dispositions. Use a safe pointer in message-activation context to protext against lost connections. Make connector/listener deletion thread safe.
- Improved the generation of stats for management.
- Merge Fixups
- Updates to the tcp tests
- Ported the tiered buffer allocation scheme from main. Make the normalization function public.  Use the calculated platform memory limit as the default ceiling. Minor updates. Added race-suppression for the new memory stat access. Cleaning up failures in the regression tests.


ted-ross:
It should be noted that not all of the regression tests pass with this initial PR commit:

```The following tests FAILED:
	 64 - system_tests_tcp_adaptor (Failed)
	 65 - system_tests_tcp_adaptor_tls (Failed)
	 66 - system_tests_http2_tls (Failed)
```

PR #1147, opened 2023-06-29
Fixes #1146: Removes trace level logging. Use debug level logging ins…

### LINKED ISSUE #1146: Remove trace logging level in skupper-router
Move all trace level logging to debug and get rid of the trace logging level.

ganeshmurthy:
…tead of trace. Backward compatible, will set logging level to debug if trace is enabled.

### DIFF
--- a/python/skupper_router_internal/management/agent.py
+++ b/python/skupper_router_internal/management/agent.py
@@ -277,9 +277,7 @@ class LogEntity(EntityAdapter):
         # This is for backward compatibility, we will remove this after a few releases.
         if attributes and attributes.get("enable") and "trace" in attributes.get("enable"):
             enabled = attributes.get("enable")
-            if enabled:
-                enabled = enabled.replace("trace", "debug" )
-                attributes["enable"] = enabled
+            attributes["enable"] = enabled.replace("trace", "debug" )
         # Special defaults for DEFAULT module.
         if attributes.get("module") == "DEFAULT":
             defaults = dict(enable="info+", includeTimestamp=True, includeSource=False, outputFile="stderr")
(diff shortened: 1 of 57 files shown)


python/skupper_router_internal/management/agent.py
```diff
@@ -273,8 +273,13 @@ def __str__(self):
 
 
 class LogEntity(EntityAdapter):
-
     def __init__(self, agent, entity_type, attributes=None, validate=True):
+        # This is for backward compatibility, we will remove this after a few releases.
+        if attributes and attributes.get("enable") and "trace" in attributes.get("enable"):
+            enabled = attributes.get("enable")
+            if enabled:
```
kgiusti, line 280: Is this check redundant? attributes.get("enabled") was already checked on line 278

python/skupper_router_internal/router/hello.py
```diff
@@ -18,7 +18,8 @@
 #
 
 from .data import MessageHELLO
-from ..dispatch import LOG_TRACE, LOG_CRITICAL
+from ..dispatch import LOG_DEBUG, LOG_CRITICAL
+from ..dispatch import LOG_DEBUG, LOG_CRITICAL
```
kgiusti, line 22: cut and paste duplicate?

kgiusti (approved):

python/skupper_router_internal/router/hello.py
ganeshmurthy: Removed the dup line, thanks

python/skupper_router_internal/management/agent.py
ganeshmurthy: Yep. Fixed ! Thanks

Issue #1145, opened 2023-06-29
Connection Idle timeout feature for TCP Adaptor


kgiusti:
Proposal: Add the ability to detect and close idle TCP connections.

Add a new configuration attribute to the tcpConnector and tcpListener records - idleTimeoutSecs. This value sets a limit to the time a TCP connection can remain idle - i.e. passing no data - before the router will proactively close the connection. The value is expressed in seconds. A value of zero disables the idle timeout (the connection will not be closed).

If configured on both ends of the connection (both the tcpConnector and the tcpConnection) the lowest value takes precedence.

kgiusti:
I'm thinking that we probably only need to add this to the tcpListener since it doesn't matter which side of the connection is closed.

kgiusti:
Leaving as draft for the moment. As written it will require a fix to #1155 since this patch requires a large number of simultaneously active timers.

ganeshmurthy:
Closing this since the associated PR is closed. Reopen if necessary

Issue #1134, opened 2023-06-21
Router crash on sys_mutex_unlock


ganeshmurthy:
In our Longevity machine, we have a iperf test running for the last 4 days, and we got 11 router restarts
 The test runs in 3 namespaces (public1, public2, private1), created in a single CRC cluster. 
 Each namespace has an iperf server exposed, but the client runs in the private1 NS.
 We send requests for all the 3 services, forever, using this commands : 
```
     while : do
       iperf3 -c iperf3-server-a -n 1G -P 5
       iperf3 -c iperf3-server-b -n 1G -P 5
       iperf3 -c iperf3-server-c -n 1G -P 5
     done  
 ```

 These are the images we are using in this test : 

```
    skupper version -n private1
      client version                 1.4.0.rc3-rh-2
      transport version              xxx.redhat.com/rh-osbs/service-interconnect-skupper-router-rhel9:2.4.0-8 (sha256:1235980cd823)
      controller version             xxx.redhat.com/rh-osbs/service-interconnect-service-controller-rhel9:1.4.0-3 (sha256:63116dbceb05)
      config-sync version            xxx.redhat.com/rh-osbs/service-interconnect-config-sync-rhel9:1.4.0-2 (sha256:030901042633)
      flow-collector version         not-found

This is the status of the pods : 

Namespace: public1
  skupper-router-779c9dff6-k7nqj                2/2     Running   0          4d20h
  skupper-service-controller-69fb6bbf7d-tgkl7   1/1     Running   0          4d20h

Namespace: public2
  skupper-router-54978cd876-56fd2               2/2     Running   0          4d20h
  skupper-service-controller-6ff6576c69-wbkxl   1/1     Running   0          4d20h

Namespace: private1
  skupper-router-7c8fb87d75-fmnnf               2/2     Running   11 (8h ago)   4d20h
  skupper-service-controller-5679f4cf86-s57r8   1/1     Running   0             4d20h

From the pod logs : 

  *** SKUPPER-ROUTER FATAL ERROR ***
  Version: 2.4.0-rc3-rh-8
  Signal: 11 SIGSEGV
  Process ID: 1 (skrouterd)
  Thread ID: 15 (wrkr

_0)

and also : 

  023-06-20 05:09:31.720307 +0000 TCP_ADAPTOR (info) [C47525] PN_RAW_CONNECTION_DISCONNECTED listener, drained_buffers=0
[...]

PR #1148, opened 2023-06-30
ISSUE-1134: Implement idle connection timeout for tcpListener

kgiusti:


ganeshmurthy:
Can we please change the commit message to "Fixes #1134:......"
That way we will have link to the issue right on the commit which will help us navigate to the issue quickly and merging this PR will close the issue automatically

python/skupper_router/management/skrouter.json
```diff
@@ -1254,6 +1254,13 @@
                     "type": ["up", "down"],
                     "description": "The operational status of TCP socket listener: up - the service is active and incoming connections are permitted; down - the service is not active and incoming connection attempts will be refused.",
                     "create": false
+                },
+                "idleTimeoutSeconds": {
```
ganeshmurthy, line 1258: We used to have an attribute called dataConnectionCount which used to be in the connector entity and @grs requested that it be moved to the router entity which we did. @grs, I just wanted to get your thoughts on whether this idleTimeoutSeconds should be in the router entity as well ?

python/skupper_router/management/skrouter.json
```diff
@@ -1254,6 +1254,13 @@
...
+                    "type": "integer",
+                    "default": 300,
+                    "description": "The idle timeout, in seconds, for connections through this listener.  If no data is transferred over the connection for this time interval, the connection shall be closed. A value of zero disables the timeout",
```
ganeshmurthy, line 1261: If users want to turn this feature off by setting 0 for idleTimeoutSeconds, there will have to be a skupper command line command that will need to be added to allow this to be done. Just like how @grs added a command line option to set dataConnectionCount

python/skupper_router/management/skrouter.json
kgiusti: From a code standpoint it's easier to have it in the listener.  Having it in the listener also allows us to disable it per-service type.  More flexibility. You'll also notice that the AMQP listener and connector management object have the same 'idleTimeoutSeconds' attribute, so adding it to the tcpListener is consistent with that.

I don't see a compelling advantage to having a single global value for all services. Is there a use-case where it makes sense?

kgiusti:
> Can we please change the commit message to "Fixes #1134:......" That way we will have link to the issue right on the commit which will help us navigate to the issue quickly and merging this PR will close the issue automatically

I can update the git commit message, but I've also set the issue in the github Development attribute associated with this pull request which will automagically close the related issue.  See left sidebar.

python/skupper_router/management/skrouter.json
kgiusti: I'll add the control-plane devs to the review as well as reach out to them directly.

ganeshmurthy:
> > Can we please change the commit message to "Fixes #1134:......" That way we will have link to the issue right on the commit which will help us navigate to the issue quickly and merging this PR will close the issue automatically
> 
> I can update the git commit message, but I've also set the issue in the github Development attribute associated with this pull request which will automagically close the related issue. See left sidebar.

Yes, update the commit message please. The commit message, when referencing an issue, would be nice if we start the commit message with "Fixes #yyy: ...." as we agreed to do here - https://github.com/skupperproject/skupper-router/blob/main/CONTRIBUTING.adoc

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -1089,6 +1115,21 @@ static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void
     case PN_RAW_CONNECTION_WAKE: {
         qd_log(LOG_TCP_ADAPTOR, QD_LOG_DEBUG, "[C%" PRIu64 "] PN_RAW_CONNECTION_WAKE %s", conn->conn_id,
                qdr_tcp_connection_role_name(conn));
+        if (CLEAR_ATOMIC_FLAG(&conn->check_idle_conn)) {
```
ganeshmurthy, line 1118: Can we do this bit of code only if conn->ingress is true ?

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -1209,9 +1250,14 @@ static qdr_tcp_connection_t *qdr_tcp_connection(qd_tcp_listener_t *listener, qd_
     assert(tcp_stats);
     assert(server);
 
+    if (tc->config->adaptor_config->idle_timeout) {
```
ganeshmurthy, line 1253: Move this into the above if (listener) so we can have idle timers only for connections on the listener side ?

src/adaptors/tcp/tcp_adaptor.c
kgiusti: Good idea! How about we check for existence of the idle_timer? The patch already does that elsewhere since the idle_timer is only allocated if idleTimeoutSeconds != 0.

ganeshmurthy: Cool yo

src/adaptors/tcp/tcp_adaptor.c
kgiusti: We may get a request to add timeout support for the connector side as well. Still not carved in stone but if so I'd rather not conventionalize the code on the type of configuration. We'll just have to revert that change if so.

kgiusti:
> > > Can we please change the commit message to "Fixes #1134:......" That way we will have link to the issue right on the commit which will help us navigate to the issue quickly and merging this PR will close the issue automatically
> > 
> > 
> > I can update the git commit message, but I've also set the issue in the github Development attribute associated with this pull request which will automagically close the related issue. See left sidebar.
> 
> Yes, update the commit message please. The commit message, when referencing an issue, would be nice if we start the commit message with "Fixes #yyy: ...." as we agreed to do here - https://github.com/skupperproject/skupper-router/blob/main/CONTRIBUTING.adoc

Hmmm... ok so I changed the commit message as described (needed to squash/rebase) and I've updated the name of this PR, but now the link to the actual issue in the Development field has gone away. Is this the correct message syntax?

ganeshmurthy:
> Can we please change the commit message to "Fixes #1134:......"

I put in the wrong issue #, the correct issue # is https://github.com/skupperproject/skupper-router/issues/1145
Even after squash rebase, I see the commit message say "ISSUE-1134:
The commit message should say "Fixes #1145: ...."
My apologies.

kgiusti:
Marking this as Draft:

This approach won't scale effectively due to its timer-per-connection implementation. Our timer implementation is the problem: it maintains a linear sorted linked list of timers. As the number of connections scale so do the timers. Scheduling a timer becomes prohibitively expensive timewise as the number of connections scales. In my tests when I scale to 65K timers scheduling a timer can take several milliseconds each (lock held).

A better approach might be to have a timer-per listener and sweep the list of associated connections when the timer expires. This is similar to the approach taken for stuck delivery detection.

I'll need to think more about this.

python/skupper_router/management/skrouter.json
nluaces: (issue created in skupper project)

jiridanek: * https://github.com/skupperproject/skupper/issues/1167

lets link it both ways and from all places :)

python/skupper_router/management/skrouter.json
grs: The counter to that is whether we really envisage users wanting to set different values for this? What will the default be? Could there be a way to configure the default (i.e. a way to change it for all listeners in one place instead of changing the configuration of each individual listener)?

What advice would we give users about when to use this option and what to set it to? And can we make it so that in our opinion the majority of users can ignore it?

PR #1151, opened 2023-07-03
Fixes #1150: Removed NOTICE log level and moved all NOTICE log level …

### LINKED ISSUE #1150: Remove NOTICE log level and move NOTICE level log statements to INFO
As part of the log level cleanup, we are getting rid of the NOTICE log level and moving all NOTICE log statements to INFO log statements.

ganeshmurthy:
…statements to INFO level.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1512,7 +1512,7 @@ void qd_server_run(qd_dispatch_t *qd)
     }
     free(threads);
 
-    qd_log(LOG_ROUTER, QD_LOG_DEBUG, "Shut Down");
+    qd_log(LOG_ROUTER, QD_LOG_INFO, "Shut Down");
 }
 
 void qd_server_stop(qd_dispatch_t *qd)
--- a/src/server.c
+++ b/src/server.c
@@ -839,12 +839,12 @@ static void handle_listener(pn_event_t *e, qd_server_t *qd_server, void *context
             pn_netaddr_str(na, str, sizeof(str));
             // "str" contains the host and port on which this listener is listening.
             if (li->config.name)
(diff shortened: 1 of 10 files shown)


src/server.c
```diff
@@ -1481,7 +1481,7 @@ void qd_server_run(qd_dispatch_t *qd)
     int i;
     assert(qd_server);
     assert(qd_server->container); // Server can't run without a container
-    qd_log(LOG_SERVER, QD_LOG_NOTICE, "Operational, %d Threads Running (process ID %ld)",
+    qd_log(LOG_SERVER, QD_LOG_INFO, "Operational, %d Threads Running (process ID %ld)",
```
kgiusti, line 1484: Not a critique of this patch, but I've found this log message (and the "shutdown" message on line 1515) useful in many cases. In fact the "shutdown" log is very useful to determine if the router had a controlled shutdown ("shutdown" will be in the log) vs. a kill -9 or a truncated logfile ("shutdown" not in the log).

This makes me think that It Would Be Nice (TM) to have status+configuration log messages that only get emitted once at startup and shutdown that should be enabled by default.  But that's probably better left to a separate effort from this patch.

Not a big deal - unless others feel it should stay.

kgiusti (approved):

src/server.c
ganeshmurthy: > useful in many cases. In fact the "shutdown" log is very useful to determine if the router had a controlled

I will leave the router "Shut Down" log message at INFO level so it is visible by default

ganeshmurthy: > But that's probably better left to a separate effort from this patch.

Yes, I agree that the status+configuration one time log messages should be INFO level. I will put out another PR which will try to do this. Thx

PR #1156, opened 2023-07-06
Don't install Proton into the run container image, make it build image only

jiridanek:
Proton does not have to be installed for the run image using cmake. The proton-python binding is installed separately with pip install. Proton libraries are being linked statically so they are not needed at runtime.

This commit fixes the issue where Proton installation brought example certificates to the image, which triggered security warning from a container scanning tool.


ganeshmurthy (approved):
Nice work. I built a container with this PR and looked in the container and there was no /usr/share/proton/ folder. 

LGTM. Approved !

PR #1157, opened 2023-07-06
Run container security scanners on the router image in CI

jiridanek:
I encountered some strange behaviour of the trivy scanner, for which I opened

* https://github.com/aquasecurity/trivy/discussions/4791

Depending on what I learn there, the GHA may need to be run with different options for trivy, maybe.


ganeshmurthy (approved):
Nice work !

PR #1162, opened 2023-07-13
Fixes #1015: Add API to verify thread context at runtime

### LINKED ISSUE #1015: Runtime Debug tests to validate proper thread execution

kgiusti:


cliffjansen (approved):
Looks good. Ship it.

Consider whether in future, at the same time you determine the proactor mode, you also choose a handler type (connection, raw_connection, listener, other).  Then you can separate handlers by functionality and use the same handler until batch_done.

This could be worth while for some code clarity goal.  It would not provide any performance benefit/penalty nor would it speak to the main thrust of this pull request.

src/posix/threading.c
```diff
@@ -131,20 +131,35 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
...
+    char                       name[SYS_THREAD_NAME_MAX + 1];
 };
 
 // initialize the per-thread _self to a non-zero value.  This dummy value will
 // be returned when sys_thread_self() is called from the process's main thread
 // of execution (which is not a pthread).  Using a non-zero value provides a
 // way to distinguish a thread id from a zero (unset) value.
 //
-static sys_thread_t           _main_thread_id = {.name = "main"};
+static sys_thread_t           _main_thread_id = {.role = SYS_THREAD_MAIN, .name = "main"};
```
ganeshmurthy, line 162: This question is not related to this PR. Why is this called _main_thread_id and not just _main_thread ?

src/posix/threading.c
```diff
@@ -131,20 +131,35 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
...
 
 struct sys_thread_t {
     pthread_t thread;
     void *(*f)(void *);
     void *arg;
-    char  name[16];
+    void                      *proactor_context;
+    sys_thread_proactor_mode_t proactor_mode;
+    sys_thread_role_t          role;
+    char                       name[SYS_THREAD_NAME_MAX + 1];
```
ganeshmurthy, line 154: Can this be a pointer into the thread_names array ? That way you can simply point to it instead of needing to strcpy ? I am just trying to find out if what I am suggesting can be done.

src/posix/threading.c
```diff
@@ -176,7 +204,7 @@ sys_thread_t *sys_thread_self(void)
 }
 
 // note: called by signal handler, do not invoke async signal unsafe code in this function!  See man signal(7)
-const char *sys_thread_name(sys_thread_t *thread)
+const char *sys_thread_name(const sys_thread_t *thread)
```
ganeshmurthy, line 207: A pedestrian programmer like me would think that there would be a return statement in this function but I don't see any.

src/posix/threading.c
```diff
@@ -195,3 +223,154 @@ void sys_thread_join(sys_thread_t *thread)
...
 }
+
+sys_thread_role_t sys_thread_role(const sys_thread_t *thread)
+{
+    if (!thread)
+        thread = sys_thread_self();
+    return thread->role;
+}
+
+sys_thread_proactor_mode_t sys_thread_proactor_mode(void)
```
ganeshmurthy, line 234: I see a function called `sys_thread_proactor_set_mode()`, so should this function be called 'sys_thread_proactor_get_mode()' ?

src/posix/threading.c
```diff
@@ -195,3 +223,154 @@ void sys_thread_join(sys_thread_t *thread)
...
+sys_thread_proactor_mode_t sys_thread_proactor_set_mode(sys_thread_proactor_mode_t new_mode, void *context)
+{
+    assert(_self->role == SYS_THREAD_PROACTOR);
+    sys_thread_proactor_mode_t old_mode = _self->proactor_mode;
+    _self->proactor_mode                = new_mode;
+    _self->proactor_context             = context;
+    return old_mode;
+}
+
+void *sys_thread_proactor_context(void)
```
ganeshmurthy, line 248: In the same vein, should this be called 'sys_thread_proactor_get_context()'

src/posix/threading.c
```diff
@@ -195,3 +223,154 @@ void sys_thread_join(sys_thread_t *thread)
...
+    if ((sys_thread_proactor_mode() & SYS_THREAD_PROACTOR_MODE_CONNECTION) == 0
+        || conn != (struct pn_connection_t *) sys_thread_proactor_context()) {
+        fprintf(stderr, "Invalid proactor AMQP connection @ %s:%d\n", file, line);
+        abort();
+    }
+}
+
+// runtime validation of proactor thread raw connection context
+//
+void sys_thread_proactor_check_raw_connection(struct pn_raw_connection_t *raw_conn, const char *file, int line)
```
ganeshmurthy, line 267: Same, can we call this `sys_thread_proactor_check_raw_connection_mode()`

ganeshmurthy (approved):

ganeshmurthy:
@kgiusti please change the commit message to `Fixes #1015:....`

src/posix/threading.c
ganeshmurthy: > I could, except that the thread_names array contains the name prefix in the case of the worker threads

Ah ok, I did not think about the `wrkr_` prefix

ganeshmurthy: > I could, except that the thread_names array contains the name prefix in the case of the worker threads ("wrkr_0", "wrkr_1", ... "wrkr_"N). Those names would need a separate buffer to hold the constructed name.
> 
> I can rename that array to "thread_name_prefixes" or "thread_name_formats" or some other more descriptive name. Or I can update the comment to clarify. Let me know.

Don't bother. I will mark this as Resolved

src/posix/threading.c
kgiusti: It's there: see line 211

src/posix/threading.c
kgiusti: I was trying to keep it consistent with the other simple accessor functions, namely sys_thread_name() and sys_thread_self(), for with the "get" is implied.

src/posix/threading.c
kgiusti: Yeah, keeping consistent with the API.

src/posix/threading.c
```diff
@@ -195,3 +223,154 @@ void sys_thread_join(sys_thread_t *thread)
...
+
+void *sys_thread_proactor_context(void)
+{
+    assert(_self->role == SYS_THREAD_PROACTOR);
+    return _self->proactor_context;
+}
+
+// runtime validation of proactor thread AMQP connection context.
+//
+void sys_thread_proactor_check_connection(struct pn_connection_t *conn, const char *file, int line)
```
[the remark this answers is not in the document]
kgiusti: But it's purpose is to validate that the passed AMQP connection is the same AMQP connection that the calling thread is scheduled for, not the mode.

However the design is that you never invoke these methods directly - you should be using the corresponding macro instead (CHECK_PROACTOR_CONNECTION() in this case).  Let me update the headers to make that clearer.

If you simply want to verify the mode you need to use ASSERT_PROACTOR_MODE().

src/posix/threading.c
kgiusti: Rename done.

src/posix/threading.c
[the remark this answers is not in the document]
kgiusti: I've updated the API comments in threading.h - let me know.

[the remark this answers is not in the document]
ganeshmurthy: > I've updated the API comments in threading.h - let me know.

LGTM. Ship it ! But before shipping please change commit message to `Fixes #1015.....`

PR #1168, opened 2023-07-18
2.4.x ci fixes

kgiusti:
I'm trying to get CI to pass on the 2.4.x branch by backporting CI fixes that are on main but missing from 2.4.x.

With this patch the only test that fails is the Container Image build, but I'm a little unclear as to which commit on main needs to be backported to fix this:

https://github.com/kgiusti/skupper-router/actions/runs/5580430927/jobs/10213011451#step:7:21

Any ideas?  Thanks.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -822,7 +822,7 @@ jobs:
       matrix:
         os: [ 'ubuntu-20.04' ]
         container: [ 'centos' ]
-        containerTag: [ 'stream8' ]
+        containerTag: [ 'stream9' ]
 
     container:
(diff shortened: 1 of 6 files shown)


jiridanek:
@kgiusti here's what comes to mind

* https://github.com/skupperproject/skupper-router/commit/0b2eb4ac13868ee4e45191d9efd71fd50adce6ca

ganeshmurthy:
Can we comment out the h2 spec test just on the 2.4.x branch  to make the Container Image build pass ?

kgiusti:
@jiridanek most excellent - thanks!

jiridanek (approved):

PR #1170, opened 2023-07-21
Fixes #1169: use proper proton install directory in compile.sh

### LINKED ISSUE #1169: compile.sh uses conflicting values for the proton installation directory

kgiusti:


.github/scripts/compile.sh
```diff
@@ -185,7 +185,7 @@ export CXXFLAGS="${CXXFLAGS} ${common_sanitizer_flags}"
 do_build "_asan" asan
 do_build "_tsan" tsan
 
-tar -z -C "${PROTON_INSTALL_DIR}" -cf /qpid-proton-image.tar.gz usr
+tar -v -z -C "${PROTON_INSTALL_DIR}" -cf /qpid-proton-image.tar.gz usr
```
jiridanek, line 188: I might be missing something, but doesn't this in effect undo https://github.com/skupperproject/skupper-router/commit/60bbbf88b170e38a3e6c1c27c2227fd3554b32ca?

Let's see what the "security" checkers think of this in the Container build CI job. Also, whether logs say it is installing the proton certificates, thanks to your helpful `-v` here.

jiridanek: Ah, got it. We ship onwards only `${PROTON_INSTALL_DIR}/usr`, where only proton-python goes, as installed by pip. The rest of proton installed by cmake goes to `${PROTON_INSTALL_DIR}/lib64` and so on (what are usually the subdirs of `/usr`, but not in this case), therefore that does not get tar-gzed.

Hmm, do you really think this is any more sane than what was being done before?

ganeshmurthy: One thing I like in the change made by @kgiusti is that the install directory of proton will follow the same directory pattern as other software we are building like 

```
LWS_INSTALL_DIR="${LWS_DIR}/install"
LIBUNWIND_INSTALL_DIR="${LIBUNWIND_DIR}/install"
```

jirkadanek: Yes, but the other software ends up actually installed in the final image. Proton doesn't, or, only proton Python binding does.

ganeshmurthy: > Yes, but the other software ends up actually installed in the final image. Proton doesn't, or, only proton Python binding does.

Agreed. 
I want to understand what this change is trying to do. I looked at https://github.com/skupperproject/skupper-router/issues/1169 and it says "compile.sh uses conflicting values for the proton installation directory". 
I think we might need to add a bit more detail to the issue description.

kgiusti: @ganeshmurthy @jiridanek @jirkadanek 

TBH I'm not 100% sure this change is correct. The reason I proposed it was that some downstream rebuilders are hitting "file not found" errors when they run this script.

IIUC, the current version of compile.sh defines the following proton related paths:

PROTON_DIR="$WORKING_DIR/proton/app"
PROTON_INSTALL_DIR="${PROTON_DIR}/proton_install"
PROTON_BUILD_DIR="${PROTON_DIR}/build"

Note the value of PROTON_INSTALL_DIR. For example if WORKING_DIR is "/foo", then PROTON_INSTALL_DIR becomes "/foo/proton/app/proton_install"

No problem there, however the subsequent calls to cmake in do_build pass the following install dir value for installing proton:

-DCMAKE_INSTALL_PREFIX=$PROTON_BUILD_DIR${suffix}/install

Note that suffix is undefined for the non-asan/tsan build - that's fine. Continuing with our foo example, this should install proton in:

/foo/proton/app/build/install/

Correct?  So here's the problem:

Later in the script there is this line:

tar -z -C "${PROTON_INSTALL_DIR}" -cf /qpid-proton-image.tar.gz usr

Using the foo example, the directory that tar is attempting to read is

/foo/proton/app/proton_install/ 

but it wasn't installed there, right?  since CMAKE used  $PROTON_BUILD_DIR${suffix}/install/ which is  /foo/proton/app/build/install/

That results in the "file not found" error.

Now the proposed fix may not be correct.  As @ganeshmurthy points out, I've taken the liberty to rename the PROTON paths a bit to match the LWS paths.  It doesn't have to be that way, we can use whatever paths we wish.

Its just that we need to create the tar file using the correct install path, which is what is passed to cmake, right?

Again, this patch may not be correct - I really  don't know.

jiridanek: > TBH I'm not 100% sure this change is correct. The reason I proposed it was that some downstream rebuilders are hitting "file not found" errors when they run this script.

And this change fixes the errors? In that case it has to be taken seriously. I'll go look again.

ganeshmurthy: > I guess that other script is not doing this pip install

I believe the downstream compile.sh does not do that pip install. Instead it just installs python-proton via rpm

kgiusti: Ok it sounds like the issue is with the downstream builder.  Let's not merge this patch.

kgiusti:
Not the proper fix, abandoned.

PR #1174, opened 2023-07-25
Fixes #1173 - Added a terminus capability for inbound targets to indi…

### LINKED ISSUE #1173: Add a terminus capability for targets to indicate that deliveries will be streaming.
Setting the "qd.streaming-deliveries" capability to the target of an inbound link to the router will cause the router to treat all receive-incomplete messages on that terminus as streaming deliveries.

This is useful for testing but also for external adaptors that may use streaming deliveries.

ted-ross:
…cate deliveries to the terminus are streaming.

### DIFF
--- a/tests/system_tests_streaming_deliveries.py
+++ b/tests/system_tests_streaming_deliveries.py
@@ -1,0 +1,356 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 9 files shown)


kgiusti (approved):

tests/system_tests_streaming_deliveries.py
```diff
@@ -0,0 +1,356 @@
...
+    def timeout(self):
+        self.error = "Timeout Expired - n_ready_to_query=%d, n_ready_to_stream=%d, n_sent_fragments=%d, n_received_streams=%d" % \
+            (self.n_ready_to_query, self.n_ready_to_stream, self.n_sent_fragments, self.n_received_streams)
+        self.receiver_conn.close()
+        for ss in self.stream_senders:
+            ss.close()
+        if self.poll_timer:
+            self.poll_timer.cancel()
+
+    def fail(self, error):
```
ganeshmurthy, line 253: Can we call this function bail() instead of fail() ? fail() with called with an error=None means the test has succeeded which sounds kinda weird.

ganeshmurthy (approved):
LGTM

PR #1175, opened 2023-07-26
Fixes #1172: Emit flow records on high severity log messages

### LINKED ISSUE #1172: New feature: generate a router-level van flow record for warning+ log events
When the router detects an operational failure (e.g. TLS misconfiguration) it generates a log message with severity warning+. These log messages get written to the logfile or syslog.

The downside to this is that typically no one looks at the router logs until _after_ the router has failed in a service-affecting manner. Additionally logs are often truncated/rotated out on low memory/resource deployments (like containers) and older log messages can be lost.

A better approach would involve having the router proactively alert the skupper control plane when error events occur. This would allow the skupper control plane to immediately report a router issue via user-facing Van Flow monitoring system. 

The goal of this proactive event generation is to detect potentially service-affecting issues early in order to address them as soon as possible.

kgiusti:


src/log.c
```diff
@@ -337,19 +342,58 @@ static void write_log(qd_log_source_t *log_source, qd_log_entry_t *entry)
...
     }
-    if (sink->syslog) {
-        int syslog_level = level->syslog;
-        if (syslog_level != -1)
-            syslog(syslog_level, "%s", log_str);
+
+    // Emit a vanflow record for "serious" log events. Ignore vanflow logs to prevent infinite loops.
+    // Do this while holding the log_event_lock. This will prevent the vanflow thread from shutting
+    // down and freeing resources while we generate the record
+    if (level->bit >= QD_LOG_WARNING && strcmp("FLOW_LOG", entry->module) != 0) {
```
ganeshmurthy, line 365: I was wondering if instead of doing `strcmp("FLOW_LOG", entry->module)`, should we store a `qd_log_module_t`
in `qd_log_entry_t`. Will that be more efficient than `strcmp` ?

ganeshmurthy (approved):

kgiusti:
Need to make some changes based on (offline) review comments.  Hold off merging/reviewing until I update this patch, thanks.

src/log.c
kgiusti: In this case my perf testing shows the strcmp overhead to be basically none. The bit check ends up short circuiting and avoiding the strcmp entirely.

However in general storing the log module in the log entry may be a general improvement for the logging module, but that would be way beyond the scope of this patch.

ganeshmurthy: > However in general storing the log module in the log entry may be a general improvement for the logging

Yes, we could do this as a different patch. Since warnings (and CRITICAL and ERROR) are not that frequent, it is ok the way it is now. 

I don't have any more questions about this PR, it is ready to ship.

Issue #874, opened 2022-12-08
Error when compiling with python 3.11, library copy_file moved to distutils python package.


rafaelvzago:
When manually compiling skupper-router using  Python 3.11.

#### Versions:
qpid_proton = 0.38.0 (compiled from its source)
skupper_router = 2.2.0 (TAG)
python= 3.11
os: Fedora
command: bin/rebuild.sh

#### Error:
```
-- Installing: /home/rzago/Git/skupper-router/install/lib/skupper-router/python/skupper_router_internal/__pycache__/__init__.cpython-311.pyc
-- Installing: /home/rzago/Git/skupper-router/install/lib/skupper-router/python/skupper_router_internal/__pycache__/dispatch.cpython-311.pyc
Traceback (most recent call last):
  File "/home/rzago/Git/skupper-router/build/python/setup.py", line 23, in <module>
    from setuptools.command.build_ext import copy_file
ImportError: cannot import name 'copy_file' from 'setuptools.command.build_ext' (/home/rzago/.local/lib/python3.11/site-packages/setuptools/command/build_ext.py)
-- Installing: /home/rzago/Git/skupper-router/install/share/doc/skupper-router/skrouter.json
-- Installing: /home/rzago/Git/skupper-router/install/share/doc/skupper-router/skrouter.json.readme.txt
-- Installing: /home/rzago/Git/skupper-router/install/sbin/skrouterd
```

#### Workaround to compile
```
diff --git a/python/setup.py.in b/python/setup.py.in
index cc5d9453..4f279ee5 100755
--- a/python/setup.py.in
+++ b/python/setup.py.in
@@ -20,7 +20,7 @@
 
 from setuptools import setup
 from setuptools.command.build_py import build_py
-from setuptools.command.build_ext import copy_file
+from distutils.file_util import copy_file
 from os.path import join
 from os import environ
```

ganeshmurthy:
@rafaelvzago thank you very much for spotting and fixing this issue.

PR #1179, opened 2023-08-04
Issue #440: remove the last remaining distutils import

jiridanek:


jiridanek:
I think that this is the correct fix for @rafaelvzago 's issue

* https://github.com/skupperproject/skupper-router/issues/874

ganeshmurthy (approved):

PR #1195, opened 2023-09-12
Fixes #1194: Remove tox and run the flake8, pylint and mypy directly

### LINKED ISSUE #1194: Remove tox from skupper-router
Currently tox is being used to run the python linters but tox pulls in different python interpreters and screws up the dev environment. Remove tox and run the python linters directly.

ganeshmurthy:


kgiusti (approved):
Thanks for cleaning up those outstanding python errors.

PR #1200, opened 2023-09-14
Fixes #1182: Refactor Prometheus metrics reporting (ISSUE-1182 part 2)

### LINKED ISSUE #1182: Update/improve prometheus metrics
The router provides a set of metrics that can be scraped by Prometheus via the HTTP protocol. There's a few metrics that need to be added and updated, specifically involving the memory pool cache.

kgiusti:
o Implement dynamic registration of each alloc_pool metric
 o Refactor http-libwebsockets.c /metrics implementation
 o Update unit tests


src/http-libwebsockets.c
```diff
@@ -96,18 +98,28 @@ typedef struct connection_t {
...
+static void free_stats_request_state(stats_request_state_t *);
 
+// Context passed to metrics and healthz protocol callbacks. Instantiated by the LWS thread, represents a single HTTP
+// request transaction.
+//
 typedef struct stats_t {
-    size_t current;
-    bool headers_sent;
-    stats_request_state_t *context;
+    stats_request_state_t *state;
```
ganeshmurthy, line 121: Why not simply move the response_complete flag into `stats_request_state_t` and get rid of `stats_t` ?

src/http-libwebsockets.c
```diff
@@ -673,77 +732,106 @@ static int add_header_by_name(struct lws *wsi, const char* name, const char* val
...
-            } else {
-                qd_log(LOG_HTTP, QD_LOG_WARNING, "insufficient space in buffer");
-                break;
-            }
+        uint8_t *start = &stats->state->lws_prefix[LWS_PRE];
+        uint8_t *end = start + stats->state->buffer_size;  // first byte past buffer
+
+        // encode stats into buffer
+
+        if (lws_add_http_header_status(wsi, HTTP_STATUS_OK, &start, end)
```
ganeshmurthy, line 784: Can we please add a comment here.
Looking at the code, I am assuming that it is trying to first set the response header status to 200, sets the content type and text/plain and adds a connection: header and  finally finishes off the http header and if *any* of these operations fail, the whole request fails.

ganeshmurthy: > Can we please add a comment here.

Don't worry about this - the log message `Metrics request failed: cannot send headers` is good enough

tests/system_tests_http.py
```diff
@@ -202,26 +194,62 @@ def run(self):
...
+                      "qdr_links_blocked_total",
+                      "qdr_deliveries_redirected_to_fallback_total"]
+        for stat in r.management.query(type="io.skupper.router.allocator").get_dicts():
+            stat_names.append(stat['typeName'])
+
+        def _test(stat_names, port):
+            # sanity check that all expected stats are reported
+            resp = urlopen(f"http://localhost:{port}/metrics", cafile=self.ssl_file('ca-certificate.pem'))
+            self.assertEqual(200, resp.getcode())
+            metrics = [x for x in resp.read().decode('utf-8').splitlines() if not x.startswith("#")]
```
ganeshmurthy, line 235: I just got Conway-ed !!

src/http-libwebsockets.c
kgiusti: Most excellent question. It's because I'm paranoid.
It might be possible to get a spurious WRITEABLE event after the request is complete and we've freed the state. Can't tell for sure from the LWS code. Thought best to keep the flag out of the free-able state in case the spurious WRITEABLE occurred after the response was completed (and state freed).

PR #1208, opened 2023-09-15
Fixes #1207: add counters for total active service connections

### LINKED ISSUE #1207: Feature: add per-protocol counters for current service network connections
Count the currently active service connections on a router.  Count these per protocol (http1/2, tcp).

kgiusti:


tests/system_tests_http1_adaptor.py
```diff
@@ -61,6 +61,11 @@
 SERVER_KEY_NO_PASS = RouterTestSslBase.ssl_file('server-private-key-no-pass.pem')
 CHAINED_PEM = RouterTestSslBase.ssl_file('chained.pem')
 
+# Management Entity IDs
+LISTENER_TYPE = 'io.skupper.router.httpListener'
```
ganeshmurthy, line 65: These constants are already defined in system_test.py which can be imported here.

src/router_core/connections.c
```diff
@@ -1682,6 +1682,22 @@ static void qdr_connection_opened_CT(qdr_core_t *core, qdr_action_t *action, boo
         }
     } while (false);
 
+    const char *pname = conn->protocol_adaptor->name;
```
ganeshmurthy, line 1685: I was wondering what is the impetus to track adaptor specific connections. A customer that is using skupper/skupper-router is ever only going to need to know how many connections were opened. They don't need to be grouped like this into separate connection counters. If you using the TCP adaptor, all you need to know how many connections were opened. Could we not get away with just adding a core->num_connections and just add and subtract from that when a qdr_connection is opened and closed ?

Also this could be confusing to users if we there were http connections being opened over tcp adaptor and the http1.x counter remains zero ?

tests/system_tests_http1_adaptor.py
kgiusti: Ok

src/router_core/connections.c
kgiusti: We already have a connectionCount attribute in the router management entity that counts all open connections (basically all active qdr_connection_t's). That count includes every type of connection: inter-router, management client, edge, tcp adaptor client, http/* adaptor client.

Note well that the proposed names are "<protocol>ServiceConnections"...

What we - or at least I - want is to be able to track the number of connections currently open for skupper *services*. That is, connections associated with provisioned "skupper service". And since a skupper service is provisioned for a particular protocol (via the --protocol parameter) we can further classify the connection to better understand the usage, which would be useful for deployments that combine different service protocol types.

Imagine being able to look into a prometheus log and see that an unexpected spike in HTTP/2 client connections happens every 24 hours a 2:30 pm. Or that the number of HTTP/1 clients stays flat but the TCP client connections keep accumulating. Or that the large number of allocated qdr_connection_t's corresponds to an unexpected peak load of HTTP/1 client request (not a leak).

That level of detail would be very useful.

> Also this could be confusing to users if we there were http connections being opened over tcp adaptor and the http1.x counter remains zero ?

They'd see tcpServiceConnections non-zero, which would make sense since they configured a TCP skupper service and not an HTTP skupper service. Why would they expect to see Http1ServiceConnections when they never configured an HTTP/1 service?

ganeshmurthy: > better understand the usage, which would be useful for deployments that combine different service protocol types.

My question was based on my understanding that customers (so far) have not mixed and match protocol adaptors. They have stayed with one protocol adaptor across the board. If we see that changing, this PR makes sense. I was just concerned that we are introducing something that might potentially confuse people although the `ServiceConnections` does make it clearer.

kgiusti:
I'm going to refactor this to move the counters into their own entity.  Marked as draft for now.

src/router_core/connections.c
kgiusti: It's my understanding that there's nothing preventing provisioning multiple services with different protocols. IIUC that's a fully supported feature of Skupper.

Is that not the case? If that is indeed true won't customers be confused if they provision TCP, HTTP/2 and HTTP/1 and only those client connections using one particular service is counted? That seems like a bug to me.

ganeshmurthy (approved):

python/skupper_router/management/skrouter.json
```diff
@@ -594,6 +594,10 @@
                     "graph": true,
                     "description": "The current amount of memory in use by the router process in bytes.  This includes memory provisioned for stack, data, and code (VmSize).  This value is set to Null if the platform does not provide access to the process memory size."
                 },
+                "connectionCounters": {
+                    "type": "map",
```
kgiusti, line 598: document index values

src/protocols.c
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#include "qpid/dispatch/protocols.h"
+
+#include <stddef.h>
+
+// names must match the order of the enum entries in qd_protocol_t
+//
+static const char *_names[QD_PROTOCOL_TOTAL] = {
+    "tcp", "amqp", "http/1", "http/2"
```
kgiusti, line 27: remove slashes

PR #1226, opened 2023-09-22
Fixes #1225: prevent management from deleting egress-dispatcher conne…

### LINKED ISSUE #1225: Setting the egress-dispatch connection's adminStatus to deleted causes the router to shuffle off its mortal coil

kgiusti:
…ctions

### DIFF
--- a/src/router_core/agent_connection.c
+++ b/src/router_core/agent_connection.c
@@ -524,9 +524,12 @@ static void qdra_connection_update_set_status(qdr_core_t *core, qdr_query_t *que
         qd_iterator_t *admin_status_iter = qd_parse_raw(admin_state);
 
         if (qd_iterator_equal(admin_status_iter, (unsigned char*) QDR_CONNECTION_ADMIN_STATUS_DELETED)) {
-            // This connection has been force-closed.
-            // Inter-router and edge connections may not be force-closed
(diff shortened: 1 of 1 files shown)


kgiusti:
I don't like this solution.

It buries connection policy in the guts of the management agent. This is brittle since it's hard to find when/if it needs updating.

I think a better solution is to provide a "disallow delete vi management" flag that is provided to the connection constructor. That solution would hard to overlook and remove the hacky if statement filter in the delete code.

I'm just not certain where that flag belongs - in the qdr_connection_opened() call or in the associated qdr_connection_info_t structure.

Opinions?

ganeshmurthy (approved):

kgiusti:
I like the updated patch better. It adds a control flag to the qdr_connection_t that indicates whether or not the connection can be deleted by management.

Also folds in a few other booleans into the flags. I think this is clearer and safer than continually adding more boolean parameters to the constructor call.

ganeshmurthy:
This patch looks good to me but the CI seems to be failing on `system_tests_edge_router`

kgiusti:
Yeah there's problems here that I haven't yet figured out.
I'm closing this PR w/o merging it. I need to root cause these issues.

PR #1238, opened 2023-09-27
NO-ISSUE - Fixed python-checker failures on Fedora 36

ted-ross:


jiridanek:
@ganeshmurthy, maybe, the bigger issue is that whilst tox used to pin the version of pylint, etc. now the tests use whatever version of the tools is available on the system. Meaning that tools can be updated and new versions report new issues...

I guess if you are fine fixing what crops up, then this is fine, but if the old behavior is desired, then virtualenv would have to be used

ganeshmurthy:
> I guess if you are fine fixing what crops up

Good point Jiri. We are indeed resorting to fixing whatever crops up. 
This specific issue that Ted fixed is because of the python-checker issues he was seeing on Fedora 36 (probably using older version of the linters) which we do not see in Fedora 37.

On a related note, for whatever reason, people are happier fixing python linter issues (as and when they pop up) as opposed to clang-format.

PR #1256, opened 2023-10-12
Fixes #1254: add function name to panic call stack output

### LINKED ISSUE #1254: Feature: add function names to panic handler call stack output
We should be able to provide the function names when outputting the call stack via the panic handler. There is this bug that currently prevents us from using this libunwind feature:

https://github.com/libunwind/libunwind/issues/123

This fix should be available in the next release of libunwind (1.8?)

kgiusti:


kgiusti:
Wow - static analysis, I'm impressed!
Working on suppression/fixes...

ganeshmurthy (approved):

PR #1261, opened 2023-10-15
DO NOT MERGE; practical example of how gcc -fanalyzer works

jiridanek:


jiridanek:
Effectiveness of -fanalyzer has been proven.

My chosen approach for running clang-analyzer unfortunately does not seem to provide SARIF outputs and they cannot be obtained easily.

PR #1271, opened 2023-10-23
Fixes #1267: Removed  gdb, asan, tsan libraries and asan/tsan router …

### LINKED ISSUE #1267: Remove gdb, asan, tsan libraries and asan/tsan router executable from skupper-router container image

ganeshmurthy:
…executable from skupper-router container image


jiridanek (approved):
This also removes gdb, which maybe we want to keep around?

There is a follow-up change that needs to be made in skupper,

* https://github.com/skupperproject/skupper/issues/1252

Overall, it looks good to me.

PR #1290, opened 2023-11-03
remove trace option

pwright:
@ganeshmurthy  does this make sense?


ganeshmurthy (approved):
Ah, I missed this. Looks good to me

PR #1297, opened 2023-11-10
Fixes #1296: limit the frequency of legacy TCP window updates

### LINKED ISSUE #1296: Legacy TCP adaptor: avoid sending PN_RECIEVED window updates too frequently

kgiusti:


ganeshmurthy:
@kgiusti I am guessing you tested this with flimflam with many tcp senders and saw somewhat slight improvement in performance, or at least no drop in performance ?

kgiusti:
> @kgiusti I am guessing you tested this with flimflam with many tcp senders and saw somewhat slight improvement in performance, or at least no drop in performance ? Other this PR LGTM.

@ganeshmurthy Yes I used flimflam with iperf3 and two flows. In that case there was an increase in throughput and decrease in CPU utilization on the egress node: on my laptop throughput baseline was 16Gb/sec (best case) and increased to 19Gb/sec (best case). On the egress router total CPU utilization decreased from 191% to 171% (with --cpu-limit=3).

The decrease in window updates did not appear to affect the window close counts - they remained roughly the same between baseline and this patch.

PR #1321, opened 2023-12-18
Fixes #1305: Don't copy public skupper-router python modules into sit…

### LINKED ISSUE #1305: Remove the use of pip install to install the skupper-router python modules.
Instead just copy these modules into the install folder and modify PYTHONPATH to point to the appropriate folders

ganeshmurthy:
…e-packages. Copy them into the install folder.

Also remove use of pip install for installing the python modules. Instead just copy the python files


kgiusti (approved):
Tried a build/install/test cycle - works for me!

jiridanek (approved):

PR #1323, opened 2023-12-18
TCP Lite window flow control implementation

kgiusti:


src/adaptors/tcp_lite/tcp_lite.c
```diff
@@ -84,6 +84,31 @@ static uint64_t buffer_threshold_50;
...
+// received at the ingress adaptor it will update the window size by subtracting the section_offset from its total
+// received bytes counter. If the result of the subtraction is less than TCP_MAX_CAPACITY then backpressure is relieved
+// and empty receive buffers are given to the raw connection.
+//
+// TCP_MAX_CAPACITY: this is set to 2x the maximum number of bytes a cut through message can buffer. This makes the
+// window large enough to max out a message at each end of the TCP flow.
+//
+#define TCP_FULL_MSG      (QD_BUFFER_DEFAULT_SIZE * UCT_SLOT_COUNT * UCT_SLOT_BUF_LIMIT)
+#define TCP_MAX_CAPACITY  (TCP_FULL_MSG * UINT64_C(2))
+#define TCP_ACK_THRESHOLD TCP_FULL_MSG
```
ganeshmurthy, line 103: Can we call this TCP_ACK_THRESHOLD_BYTES since this is a byte counter ? Seems really clear that it is a BYTE counter

ganeshmurthy: What is UCT in UCT_SLOT_BUF_LIMIT ? I know UCT was already there but I just cant make out what it is by reading the name of the constant.

ganeshmurthy (approved):

mgoulish (approved):
Good comments!
I think I mostly understand what is going on, and it seems clear.

I'd like to understand sometime why window flow control would ever be disabled...

src/adaptors/tcp_lite/tcp_lite.c
kgiusti: "UCT" is short for "unicast cut through". It's used in all the cut through related stuff - it's better described in message.h where those symbols are defined.

kgiusti:
> Good comments! I think I mostly understand what is going on, and it seems clear.
> 
> I'd like to understand sometime why window flow control would ever be disabled...

The window gets disabled once the transfer has reached a point where the message is complete and needs to be drained. This specifically happens when the remote sets a _final_ outcome (like PN_ACCEPTED, etc). This is necessary since once a final outcome is set the remote cannot send any more PN_RECEIVED updates to open the window (that would be illegal according to the AMQP spec).  Another reason is when the remote settles the message, which also prevents further PN_RECEIVED updates.

I'll add a comment for this in the code.

Issue #704, opened 2022-08-31
system_tests_http2 and system_tests_http2_tls failing after running the first 2 tests


ganeshmurthy:
https://github.com/skupperproject/skupper-router/runs/8115954519?check_suite_focus=true#step:29:13422

```
56: self = <system_tests_http2.Http2TestTwoRouter testMethod=test_500>, address = 'http://127.0.0.1:22943/test/500', args = None, input = None, timeout = 300.0
56: http2_prior_knowledge = True, no_alpn = False, assert_status = True
56: 
56:     def run_curl(self, address, args=None, input=None, timeout=TIMEOUT,
56:                  http2_prior_knowledge=True,
56:                  no_alpn=False,
56:                  assert_status=True):
56:         """
56:         Run the curl command using the HTTP/2 protocol.  Returns (return code,
56:         stdout, stderr)
56:         """
56:         local_args = [str(address)]
56:         if http2_prior_knowledge:
56:             local_args += ["--http2-prior-knowledge"]
56:         if no_alpn:
56:             local_args += ["--no-alpn"]
56:     
56:         if args:
56:             local_args += args
56:     
56:         status, out, err = system_test.run_curl(local_args, input=input, timeout=timeout)
56:         if assert_status:
56: >           assert status == 0, f"CURL ERROR {status}: {out} {err}"
56: E           AssertionError: CURL ERROR 7:    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
56: E                                              Dload  Upload   Total   Spent    Left  Speed
56: E             
56: E               0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
56: E               0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
56: E             curl: (7) Failed to connect to 127.0.0.1 port 22943 after 9 ms: Connection refused
56: E             
56: E           assert 7 == 0
```

PR #1349, opened 2023-12-29
NO-ISSUE: upgrade hypercorn package in requirements-dev.txt

jiridanek:
We need to avoid buggy 0.14.x versions.

* https://github.com/skupperproject/skupper-router/issues/704


ganeshmurthy:
The HTTP2 adaptor tries to open a server side connection from the router to the http2 server/service as soon as the router starts. It is over this connection that the requests are sent from the router to the server. hypercorn>0.14

Issue #1364, opened 2024-01-02
geninfo: ERROR: Unexpected negative count '-11' for /home/runner/work/skupper-router/skupper-router/skupper-router/tests/alloc_test.c:130.


jiridanek:
This often happens on fedora 39

```
Processing /__w/skupper-router/skupper-router/skupper-router/build/tests/CMakeFiles/unit_tests.dir/alloc_test.c.gcda
geninfo: ERROR: Unexpected negative count '-11' for /home/runner/work/skupper-router/skupper-router/skupper-router/tests/alloc_test.c:130.
	Perhaps you need to compile with '-fprofile-update=atomic
	(use "geninfo --ignore-errors negative ..." to bypass this error)
Combining tracefiles.
.. found 2 files to aggregate.
Merging dispatch-base.info..1 remaining
Merging dispatch-ctest.info..0 remaining
lcov: ERROR: trace file 'dispatch-ctest.info' is empty
	(use "lcov --ignore-errors empty ..." to bypass this error)
lcov: ERROR: no valid records found in tracefile dispatch-total-raw.info
	(use "lcov --ignore-errors empty ..." to bypass this error)
genhtml: ERROR: no valid records found in tracefile dispatch-total.info
	(use "genhtml --ignore-errors empty ..." to bypass this error)
gmake[3]: *** [CMakeFiles/coverage.dir/build.make:70: CMakeFiles/coverage] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:992: CMakeFiles/coverage.dir/all] Error 2
gmake[3]: Leaving directory '/__w/skupper-router/skupper-router/skupper-router/build'
gmake[2]: Leaving directory '/__w/skupper-router/skupper-router/skupper-router/build'
gmake[1]: *** [CMakeFiles/Makefile2:999: CMakeFiles/coverage.dir/rule] Error 2
gmake[1]: Leaving directory '/__w/skupper-router/skupper-router/skupper-router/build'
gmake: *** [Makefile:543: coverage] Error 2
```

https://github.com/skupperproject/skupper-router/actions/runs/7340178454/job/19985745701#step:35:1417

PR #1374, opened 2024-01-08
F39

ganeshmurthy:


jiridanek:
Let's maybe merge the counters PR separately?

* https://github.com/skupperproject/skupper-router/pull/1368

ganeshmurthy:
> Let's maybe merge the counters PR separately?
> 
>     * [Fixes #1364: fight negative coverage counters with `-fprofile-update=atomic` #1368](https://github.com/skupperproject/skupper-router/pull/1368)

Sure, I will remove the counters commit from this PR.

tests/system_test.py
```diff
@@ -431,16 +432,19 @@ def error(msg):
...
             # running. This is cool if expected is Process.RUNNING - we'll
             # check that below
-            self.terminate()  # send SIGTERM
-            try:
-                self.wait(TIMEOUT)
-            except TimeoutExpired:
-                self.send_signal(signal.SIGABRT)  # so that we may get core dump
-                error("did not terminate properly, required SIGABORT")
+            if self.abort_process:
+                self.send_signal(signal.SIGABRT)
```
kgiusti, line 436: If I understand correctly the goal here is that there are some processes - like test servers, etc - that we really don't care if they are shutdown cleanly.

If that is the goal then I would recommend using SIGKILL rather than SIGABORT. Couple of reasons:

1. a program can mask sigabort. If the program is mis-behaving sending it a SIGABORT may not result in killing the program.
2. sigabort can generate a core dump. since the program being aborted is not something we're trying to debug the core dump just consumes resources

SIGKILL cannot be masked and it will force the process to exit immediately regardless of its state, which I think is what we want here.

kgiusti (changes requested):

kgiusti (approved):

Issue #1367, opened 2024-01-05
Build and release arm64/v2 (aarch64) skupper-router docker image


jiridanek:
This is to support

* https://github.com/skupperproject/skupper/pull/1064

## Plan

There are some implementation suggestions in the original PR

* https://github.com/skupperproject/skupper-router/pull/1100

and also @ganeshmurthy has a write-up discussing options.

In summary, we need to eventually produce multiarch manifest, but it's ok to not have it initially (and give the aarch64 image a different name, and then manually specify that when initializing skupper). We want the build to be reasonably fast (so qemu-user is out, or at least running everything in qemu-user is out). There should be two solutions, both crosscompilation and a native build. Crosscompilation is useful on GitHub Actions, and native build can be done on Travis CI or in general by those who have physical arm64 machines, such as inside Red Hat.

ganeshmurthy:
This has been resolved via https://github.com/skupperproject/skupper-router/issues/1404

PR #1396, opened 2024-02-05
Revert "Issue #1367: Build arm64/v2 (aarch64) skupper-router docker image"

jiridanek:
Since we've decided we don't want to cross-compile with zig for arm64, let's again remove the code added to support it.

Zig compilation fails with libunwind 1.8.0 as whitnessed by @ganeshmurthy .

* Reverts skupperproject/skupper-router#1100


ganeshmurthy (approved):

jiridanek:
There is a crash in system tests,

* https://github.com/skupperproject/skupper-router/issues/1397

PR #1403, opened 2024-02-07
Changes to allow multiarch skupper-router image build

ganeshmurthy:


.github/scripts/compile.sh
```diff
@@ -82,7 +82,6 @@ cmake -S "${LWS_DIR}" -B "${LWS_BUILD_DIR}" \
   -DLWS_LINK_TESTAPPS_DYNAMIC=ON \
   -DLWS_WITH_LIBUV=OFF \
   -DLWS_WITHOUT_BUILTIN_GETIFADDRS=ON \
-  -DLWS_USE_BUNDLED_ZLIB=OFF \
```
jiridanek, line 85: Why? I'm not disagreeing, just asking.

.github/workflows/build.yaml
```diff
@@ -888,16 +1033,14 @@ jobs:
...
+  container_amd64:
+    name: amd64 container image
     runs-on: ubuntu-latest
     env:
       ImageName: local/skupper-router:local
 
     steps:
-
       - uses: actions/checkout@v4
-
```
jiridanek, line 900: I like this newline. There should be a newline between steps, IMO

.github/workflows/image-main.yml
```diff
@@ -25,35 +25,68 @@ on:
...
-            -H "Authorization: Bearer ${GITHUB_TOKEN}" \
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            ${{ github.event.workflow_run.jobs_url }} > jobs.json
-
-          conclusion=$(jq < jobs.json --raw-output '.jobs[] |  select( .name == "Container image") | .conclusion')
-          echo ${conclusion}
-
-          [[ ${conclusion} == "success" ]]
+      - uses: actions/checkout@v4
+        if: ${{ !cancelled() }}
```
jiridanek, line 33:
```suggestion
```

this diff can go away too

.github/workflows/image-main.yml
```diff
@@ -25,35 +25,68 @@ on:
...
 
+      - name: Build and publish the main amd image
+        if: ${{ !cancelled() }}
```
jiridanek, line 36:
```suggestion
```

this one can be removed as well

jiridanek (commented):
There is a lot of YAML duplication in this... If I wanted to help to reduce this a bit, I'd have to study GitHub Actions features and possibly try to move some of the repeated steps into a Makefile or a helper Python scripts. I'm not sure I'll get to it any time soon, because I was intending to do this a year ago already.

I guess it is not too bad yet, it can be lived with.

.github/workflows/build.yaml
```diff
@@ -957,6 +1100,18 @@ jobs:
           podman run --rm --volume ${PODMAN_SOCK}:/var/run/docker.sock docker.io/anchore/grype:latest \
             --only-fixed --fail-on low ${{ env.ImageName }}
 
+  container_arm64:
+    name: arm64 container image
+    runs-on: self-hosted
```
fgiorgetti, line 1105: What do you think of using a more specific name, like: `self-hosted-arm64` (just in case we add
new self hosted runners for other platforms)?

fgiorgetti (approved):
Minor suggestion only, but overall it looks good.
Great work @ganeshmurthy!

ganeshmurthy:
> There is a lot of YAML duplication in this... If I wanted to help to reduce this a bit, I'd have to study GitHub Actions features and possibly try to move some of the repeated steps into a Makefile or a helper Python scripts. I'm not sure I'll get to it any time soon, because I was intending to do this a year ago already.
> 
> I guess it is not too bad yet, it can be lived with.

I learned while working on the self-hosted that environment variables cannot be shared between hosted and self-hosted machines. So, some code needs duplication. It would be nice if github offers something like a #define in C where we could simple define some code somewhere and just pull in that macro into the code.

.github/workflows/build.yaml
ganeshmurthy: > What do you think of using a more specific name, like: `self-hosted-arm64` (just in case we add new self hosted runners for other platforms)?

I have added a new label called self-hosted-arm64  to the Rasperrry Pi 5 hosted runner - 
https://github.com/skupperproject/skupper-router/settings/actions/runners/25

I will change one single line in build.yml from self-hosted to self-hosted-arm64 and if it works, I will change the rest to self-hosted-arm. Stay tuned !

.github/workflows/image-main.yml
ganeshmurthy: Done !

.github/workflows/image-main.yml
ganeshmurthy: Done !

.github/workflows/build.yaml
ganeshmurthy: > It worked !! I will change all self-hosted to self-hosted-arm64

Done !

jiridanek (approved):

PR #1425, opened 2024-02-22
Fixes #1388: implement TLS support in the tcp_lite adaptor

### LINKED ISSUE #1388: Implement TLS endpoint termination for TCP lite adaptor

kgiusti:
Closes #1388

### DIFF
--- a/src/adaptors/adaptor_listener.c
+++ b/src/adaptors/adaptor_listener.c
@@ -140,22 +140,31 @@ static void _listener_event_handler(pn_event_t *e, qd_server_t *qd_server, void 
                 break;
             }
 
-        case PN_LISTENER_ACCEPT:
-            // Log incoming client connections at DEBUG level since these can flood the router logs.
-            qd_log(log_module, QD_LOG_DEBUG, "Listener %s: new incoming client connection to %s", li->name,
-                   li->host_port);
+         case PN_LISTENER_ACCEPT: {
+             bool denied = false;
 
-            // block qd_adapter_listener_close() from returning during the accept call:
-            sys_mutex_lock(&li->lock);
-            if (li->on_accept)
-                li->on_accept(li, pn_event_listener(e), li->user_context);
-            else {
-                // The adaptor_listener is closing but a connection attempt arrived before the
-                // cleanup is complete. Deny the connection attempt.
-                qd_adaptor_listener_deny_conn(li, pn_event_listener(e));
-            }
-            sys_mutex_unlock(&li->lock);
-            break;
+             // Log incoming client connections at DEBUG level since these can flood the router logs.
+             qd_log(log_module, QD_LOG_DEBUG, "Listener %s: new incoming client connection to %s", li->name,
+                    li->host_port);
+
+             // block qd_adapter_listener_close() from returning during the accept call:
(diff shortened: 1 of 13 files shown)


src/adaptors/adaptor_listener.c
```diff
@@ -152,12 +150,9 @@ static void _listener_event_handler(pn_event_t *e, qd_server_t *qd_server, void
...
             else {
-                // the adaptor_listener is closing but a connection attempt arrived before the
-                // cleanup is complete.  Need to accept it then close it when the connection
-                // completes
-                pn_raw_connection_t *close_me = pn_raw_connection();
-                pn_raw_connection_set_context(close_me, &_conn_event_context);
-                pn_listener_raw_accept(pn_event_listener(e), close_me);
+                // The adaptor_listener is closing but a connection attempt arrived before the
+                // cleanup is complete. Deny the connection attempt.
+                qd_adaptor_listener_deny_conn(li, pn_event_listener(e));
```
ganeshmurthy, line 155: Above there is a debug log message that says new incoming client connetion. When we explicitly add another log message when we deny the connection in the _deny_conn_handler ?

src/adaptors/adaptor_listener.c
```diff
@@ -152,12 +150,9 @@ static void _listener_event_handler(pn_event_t *e, qd_server_t *qd_server, void
             if (li->on_accept)
                 li->on_accept(li, pn_event_listener(e), li->user_context);
             else {
-                // the adaptor_listener is closing but a connection attempt arrived before the
-                // cleanup is complete.  Need to accept it then close it when the connection
-                // completes
-                pn_raw_connection_t *close_me = pn_raw_connection();
-                pn_raw_connection_set_context(close_me, &_conn_event_context);
-                pn_listener_raw_accept(pn_event_listener(e), close_me);
+                // The adaptor_listener is closing but a connection attempt arrived before the
```
ganeshmurthy, line 153: We don't deny connection in the handler of the PN_LISTENER_ACCEPT event in server.c
Should we ?

src/adaptors/adaptor_listener.c
kgiusti: Sure... coming up.

src/adaptors/adaptor_listener.c
kgiusti: Hrmmm... that server listener has a different lifecycle than the adaptor listener: it stays active until the listener config is explicitly deleted.  That's unlike an adaptor listner which needs to disable/enable the socket frequently depending on the availability of a consumer...

So my feeling is no - this won't happen with server listeners.

src/adaptors/adaptor_listener.c
kgiusti: Added with latest commit.

kgiusti:
The last outstanding test failure - in the Http/1-over-tcp test - appears to be unrelated to the TLS implementation. I've opened a separate issue for tracking that:

https://github.com/skupperproject/skupper-router/issues/1432

I think the reason we're seeing that failure is because this patch enables the Http/1-over-tcp test which triggered the issue.

I'm going to merge this patch and turn my attention to #1432 next.

PR #1443, opened 2024-03-12
Fixes #1440: move TLS domain and session setup to I/O thread

### LINKED ISSUE #1440: Router timeouts due to long TLS Domain setup under load (tcp lite TLS)
This is a *good* one!

Occasionally I've been seeing random failures of the TcpTlsAdaptorLite.test_80_balancing test in CI.

On investigation it turns out the interior router is timing out the edge router (edge heartbeat loss).

The edge router is still functional, but under load from incoming client connections.

Further investigation revealed that the edge router's periodic 1 second system tick was being delayed by over 10 seconds (!), which caused the edge router to fail to produce heartbeats in a timely manner (every 2 seconds).  This was causing the lost of the connection to the interior.

Turns out the delayed tick was being caused by the timer handler that invokes the general work handler. One of the work events was taking over a 500msec to complete. As the entire work queue is drained at once via a single timer handler call this was the source of the timer blockage.

Yet more investigation narrowed the issue down to the adaptor listener lock being blocked for a long time. There are two threads that compete for each listener - the watched address updater (via general work) and the proactor thread event handler. It turns out that the address updater was blocked waiting for the listener lock for hundreds of milliseconds due to a large number of simultaneous connection attempts holding the lock way to long running the TCP lite listener's "on_accept" callback.

This callback accepts the incoming client request and fires off a new proton raw connection for it. All further connection related work happens on that new connection thread, well away from the listener lock.

*Except* for the TLS case. In that case, the on_accept handler attempts to instantiate a new TLS domain.  And *that* takes a long time because - of course - we're reading the credential files! This operation is expensive (timewise) since it involves filesystem access.

Q.E.D.

So... we need to move the creation of the TLS Domain object on to the newly spawned raw connection thread instead of running it on the listener event thread.

And we should probably implement debug code in the timer to assert() if any handler callbacks take "too long", but that's for another day.

kgiusti:
Previously TLS domain and session setup for a new connection occurred either in the (shared) listener thread or the (shared) egress-dispatcher connection thread. TLS domain/session setup is expensive timewise (> 50msec). This caused the listener/connector thread to slow down significantly when under load since each connection setup was serialized.

This patch moves the TLS domain/session setup to the new connections I/O thread, allowing the listener/connectors to service connection requests much faster and providing some parallelism via threading.

Closes #1440

### DIFF
--- a/src/adaptors/tcp_lite/tcp_lite.c
+++ b/src/adaptors/tcp_lite/tcp_lite.c
@@ -131,7 +131,7 @@ static uint64_t validate_outbound_message(const qdr_delivery_t *out_dlv);
 static void on_accept(qd_adaptor_listener_t *listener, pn_listener_t *pn_listener, void *context);
 static void on_tls_connection_secured(qd_tls_t *tls, void *user_context);
 static char *get_tls_negotiated_alpn(qd_message_t *msg);  // caller must free() returned string!
-
+static int setup_tls_session(tcplite_connection_t *conn, const qd_tls_domain_t *parent_domain, const char *alpn_protocol);
 
 //=================================================================================
 // Thread assertions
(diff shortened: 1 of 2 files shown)


ganeshmurthy (approved):
Nice work !!

PR #1449, opened 2024-03-21
Fixes #1432: prevent use-after-free race in cutthrough activation.

### LINKED ISSUE #1432: Use-after-free in cut-though activation
Very hard to reproduce, caught in CI:

```
=================================================================
==5689==ERROR: AddressSanitizer: use-after-poison on address 0x61600017ad20 at pc 0x5613b96524df bp 0x7fad9b6d5030 sp 0x7fad9b6d5020
READ of size 8 at 0x61600017ad20 thread T6
    #0 0x5613b96524de in qdr_delivery_get_context ../src/router_core/delivery.c:46
    #1 0x5613b9710d9f in qdr_node_delivery_pn_from_qdr ../src/router_node.c:118
    #2 0x5613b97157de in AMQP_conn_wake_handler ../src/router_node.c:446
    #3 0x5613b9529b4e in writable_handler ../src/container.c:387
    #4 0x5613b952bc09 in qd_conn_event_batch_complete ../src/container.c:477
    #5 0x5613b97370eb in thread_run ../src/server.c:1186
    #6 0x5613b95bd9c2 in _thread_init ../src/posix/threading.c:207
    #7 0x7fada5e94ac2 in start_thread nptl/pthread_create.c:442
    #8 0x7fada5f2684f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f) (BuildId: c289da5071a3399de893d2af81d6a30c62646e1e)

0x61600017ad20 is located 160 bytes inside of 576-byte region [0x61600017ac80,0x61600017aec0)
allocated by thread T6 here:
    #0 0x7fada74de57d in __interceptor_posix_memalign ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x5613b94f8342 in qd_alloc ../src/alloc_pool.c:348
    #2 0x5613b965230a in new_qdr_delivery_t ../src/router_core/delivery.c:24
    #3 0x5613b96c2ad2 in qdr_link_deliver ../src/router_core/transfer.c:45
    #4 0x5613b97181d5 in AMQP_rx_handler ../src/router_node.c:943
    #5 0x5613b9528385 in do_receive ../src/container.c:219
    #6 0x5613b952e8b5 in qd_container_handle_event ../src/container.c:730
    #7 0x5613b9736cb9 in handle ../src/server.c:1132
    #8 0x5613b9736f5d in thread_run ../src/server.c:1170
    #9 0x5613b95bd9c2 in _thread_init ../src/posix/threading.c:207
    #10 0x7fada5e94ac2 in start_thread nptl/pthread_create.c:442
```

The root cause that there is a stale message activation record sitting on the connection's inbound_cutthrough_worklist. When the connection processes the stale activation it drops the reference count to the delivery early. This results in the delivery being freed and the above error.

We can move the decref later in the activation handling code, but that merely closes the stale activation race window.  We need to make the clearing of the activation in the message synchronous with removing it from the connection's worklist (if it is on it).

Looking a bit at the logs it also appears as if the delivery's pointer to the corresponding proton delivery has been cleared.  This is obvious since the proton delivery increments the delivery reference count and if the proton delivery was present the delivery would not be freed by the activation handler.  So we need to be careful - add a check for a null pointer or better yet clear the activation before the proton delivery is settled.

kgiusti:
The patch adds locking to the cutthrough activation logic that prevents the activation handler from being released while another thread attempts using the activation.

Closes #1432

### DIFF
--- a/include/qpid/dispatch/cutthrough_utils.h
+++ b/include/qpid/dispatch/cutthrough_utils.h
@@ -24,10 +24,10 @@
 /**
  * Perform any needed cut-through notification upon the production or consumption of stream buffers
  * either inbound (from outside of the router) or outbound (to outside of the router).
- * 
- * @param msg - Pointer to a stream
+ *
+ * @param activation - activation record taken from the message (see message.c)
  */
-void cutthrough_notify_buffers_produced_inbound(qd_message_t *msg);
-void cutthrough_notify_buffers_consumed_outbound(qd_message_t *msg);
+void cutthrough_notify_buffers_produced_inbound(const qd_message_activation_t *activation);
+void cutthrough_notify_buffers_consumed_outbound(const qd_message_activation_t *activation);
 
 #endif
--- a/include/qpid/dispatch/message.h
+++ b/include/qpid/dispatch/message.h
@@ -754,20 +754,6 @@ void qd_message_produce_buffers(qd_message_t *stream, qd_buffer_list_t *buffers)
 int qd_message_consume_buffers(qd_message_t *stream, qd_buffer_list_t *buffers, int limit);
 
 
-/**
- * Indicate whether this stream should be resumed from a stalled state.  This will be the case
- * if (a) the stream was stalled due to being full, and (b) the payload has shrunk down below
- * the resume threshold.
- *
- * If the result is true, there is a side effect of clearing the 'stalled' state.
- *
- * @param stream Pointer to the message
- * @return true Yes, the stream was stalled and buffer production may continue
- * @return false No, the stream was not stalled or it was stalled and is not yet ready to resume
- */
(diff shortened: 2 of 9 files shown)


include/qpid/dispatch/cutthrough_utils.h
```diff
@@ -27,7 +27,7 @@
  * 
  * @param msg - Pointer to a stream
```
ganeshmurthy, line 28: The @param is not an activation object, not a msg

kgiusti: Good catch - fixed!

ganeshmurthy (approved):

PR #1464, opened 2024-04-15
Amqp adaptor

kgiusti:


ganeshmurthy (approved):

src/adaptors/amqp/qd_connection.c
ssorj, line 1: Why does this file (and similar) have the qd_ prefix?  That's not the established pattern in the codebase.

kgiusti: I'm not convinced that pattern actually helps make this code any clearer, or even if that's a pattern we've consciously accepted as part of our coding style.

The pattern we do try to use is the typical file per class (struct) implementation: The .c file holds the implementation for a particular class and there is a corresponding .h file for the exported API.

In this patch qd_connection.c is the home for the qd_connection_t implementation code (accessors, constructors, etc). It has a corresponding header named qd_connection.h for the exported API.

For this pattern the prefix is significant - especially in the case of the term "connection" for which we have several connection-type classes throughout the codebase.

For example if we strip the prefix then we've got "connection.c" and "connection.h". And if we extend the strip-the-prefix pattern to qdr_connection_t then we get "connection.c/h" as well.  When I find a source file that includes "connection.h" it's ambiguous as to which class is being included.

That's why I find qd_connection.c/h (and "qdr_connection.c/h") much clearer in terms of what those files are intended for.

Issue #1494, opened 2024-05-02
ncat TCP TLS test fails due to inbound data stall on connector-side inbound flow


kgiusti:
Very rare, but IMHO a serious issue:

Manifests as a timeout failure in CI running the TcpTlsAdaptor.test_76_half_closed_EA1_EC2_large_message test.

See [this test failure](https://github.com/skupperproject/skupper-router/actions/runs/8911242481/job/24472141917#step:36:5561)

The stall is occurring on the TLS stream coming from the echo server into the connector side TLS. While transferring the large message the TCP adaptor stops taking encrypted inbound buffers off of the raw connection. From the trace the raw connection is posting PN_RAW_CONNECTION_READ events for the incoming data, but the adaptor/tls layer is not taking them from the raw conn and moving them into the TLS decryption layer.

Here's the log trace of the incident as given on the connector-side router:

```

# The last proper transfer of encrypted data from the raw conn to the TLS layer:

2024-05-01 15:25:19.352879 +0000 TCP_ADAPTOR (debug) [C31] on_connection_event_CSIDE_IO: PN_RAW_CONNECTION_READ (/home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/tcp/tcp_adaptor.c:1965)
2024-05-01 15:25:19.352902 +0000 TCP_ADAPTOR (debug) [C31] 166 encrypted bytes read from the raw connection passed to TLS for decryption (1 buffers) (/home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/adaptor_tls.c:1301)
2024-05-01 15:25:19.352923 +0000 TCP_ADAPTOR (debug) [C31] 144 decrypted bytes taken from TLS for adaptor input (1 buffers) (/home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/adaptor_tls.c:1388)
2024-05-01 15:25:19.352945 +0000 TCP_ADAPTOR (debug) [C31] CSIDE TLS read: Produced 144 octets into stream (/home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/tcp/tcp_adaptor.c:1545)
2024-05-01 15:25:19.352988 +0000 TCP_ADAPTOR (debug) [C31] grant_read_buffers_XSIDE_IO - 1 (/home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/tcp/tcp_adaptor.c:736)

[...]

PR #1500, opened 2024-05-15
Issue #1494: add verbose TLS logging to diagnose stall

kgiusti:
This does not solve the issue. Additional logging will help us root cause the issue. This commit may be reverted after #1494 is resolved for performance improvement.


ganeshmurthy:
Since TLS is meant to go only over TCP and this adaptor tls module will not be used by any other adaptor, can the ```tls->log_module``` be replaced with `LOG_TCP_ADAPTOR` or perhaps introduce something like a LOG_TCP_TLS log module ?

This above comment is unrelated to this PR but I just had that thought. This PR is ready to go.

ganeshmurthy (approved):

PR #1512, opened 2024-05-29
Fixes #1418: fix and reenable gcc -fanalyzer

jiridanek:


.github/workflows/gcc-fanalyzer.yml
```diff
@@ -0,0 +1,114 @@
...
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 360
+    permissions:
+      actions: read
+      contents: read
+      # https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security
+      security-events: write
+
+    container:
+      image: 'quay.io/fedora/fedora:rawhide'
```
kgiusti, line 44: Why rawhide and not a stable fedora release?

jiridanek: no particular reason now; back then gcc -fanalyzer was rawhide -only; will switch it to 40

kgiusti (approved):
Looks good to me.
One question: is there a tool for viewing the resulting sarif output?

jiridanek:
> Looks good to me. One question: is there a tool for viewing the resulting sarif output?

Yes, the Security tab in GitHub. The whole point of me dealing with SARIF is that then GitHub can visualize that. Otherwise I'd go for either text output or some html like clang-analyzer (if gcc supports html; it used to not be implemented in the past).

image

jiridanek:
Fwiw, clang results are at https://skupperproject.github.io/skupper-router/#sort=file-path&asc=1

PR #1513, opened 2024-05-29
Issue #1498: Initial HTTP/1.x observer implementation

kgiusti:


include/qpid/dispatch/log.h
```diff
@@ -56,6 +56,7 @@ typedef enum {
     LOG_HTTP_ADAPTOR,
     LOG_FLOW_LOG,
     LOG_ADDRESS_WATCH,
+    LOG_HTTP_OBSERVER,
```
ganeshmurthy, line 59: Should this be LOG_HTTP1_OBSERVER so I can have a LOG_HTTP2_OBSERVER ?

kgiusti: Sure I'll make that change - update to follow...

tests/system_tests_http1_observer.py
```diff
@@ -0,0 +1,240 @@
...
+                          'port': cls.tester.get_port()}),
+
+            ('tcpListener', {'host': "0.0.0.0",
+                             'port': listener_port,
+                             'address': 'Http1ObserverTest'}),
+            ('tcpConnector', {'host': "localhost",
+                              'port': server_port,
+                              'address': 'Http1ObserverTest'}),
+
+            ('log', {'module': 'DEFAULT',
```
ganeshmurthy, line 83: `Qdrouterd` class in system_test.py already enabled `debug+` for all log modules. Is this necessary ?

ganeshmurthy (approved):

tests/vanflow_snooper.py
```diff
@@ -0,0 +1,405 @@
...
+    6: "FLOW",
+    7: "PROCESS",
+    8: "IMAGE",
+    9: "INGRESS",
+    10: "EGRESS",
+    11: "COLLECTOR",
+    12: "PROCESS_GROUP",
+    13: "HOST",
+    14: "LOG",
+}
```
astitcher, line 52: Consider using an Enum for this (`from enum import Enum`). It is nice idiomatic python that I think will work here.

tests/vanflow_snooper.py
```diff
@@ -0,0 +1,405 @@
...
+    51: "SOURCE_LINE",
+}
+
+log_severity = {
+    0: "DEBUG",
+    1: "INFO",
+    2: "WARNING",
+    3: "ERROR",
+    4: "CRITICAL"
+}
```
astitcher, line 118: Simiilarly here: Consider using an Enum for these as well.

astitcher (commented):
Using an IntEnum might be a solution here too.

tests/system_tests_http1_observer.py
kgiusti: Ah - no it's not.  Originally I had the test parse the log output looking for van flow log events to validate the observer. I changed that to use the external vanflow_snooper instead. I'll remove that log config since it's not necessary.

tests/vanflow_snooper.py
kgiusti: Nice!

tests/vanflow_snooper.py
kgiusti: Done.

include/qpid/dispatch/log.h
kgiusti: Done

tests/system_tests_http1_observer.py
kgiusti: Done

astitcher:
I was perhaps a bit too terse earlier.: The usual python recommendation is to use Enum rather than IntEnum if you can unless you have back compatibility constraints. I see you used IntEnum, but I can't tell if you could have equally easily used a plain Enum whiich recommended for new code.
Sorry to be a pain!

kgiusti:
> I was perhaps a bit too terse earlier.: The usual python recommendation is to use Enum rather than IntEnum if you can unless you have back compatibility constraints. I see you used IntEnum, but I can't tell if you could have equally easily used a plain Enum whiich recommended for new code. Sorry to be a pain!

Yeah I'd rather keep it as IntEnum.  Changing it to Enum requires explicit use of the .value attribute:

```
>>> from enum import Enum
>>> class AttributeTypes(Enum):
...     RECORD_TYPE = 0
...     IDENTITY = 1
...     PARENT = 2
... 
>>> AttributeTypes.IDENTITY
<AttributeTypes.IDENTITY: 1>
>>> AttributeTypes.IDENTITY.value
1
>>> dd = {1: 'a', 3: 'b', 0: 'c'}
>>> dd[1]
'a'
>>> dd[AttributeTypes.IDENTITY]
KeyError: <AttributeTypes.IDENTITY: 1>
>>> dd[AttributeTypes.IDENTITY.value]
'a'
```

Whereas I find 
```
dd[AttributeTypes.IDENTITY]
```
a bit clearer.

PR #1537, opened 2024-06-24
Fixes #1536: Removed unnecessary check for n >= 0

ganeshmurthy:

PR #1539, opened 2024-06-25
Fixes #1536: Replace unnecessary check for `n >= 0` with assertion that `n` is `unsigned long`

### LINKED ISSUE #1536: Coverity issue: Checking if unsigned number is >= 0
amqp.c
```
123    if (*endptr == '\0') {
     
CID 379188: (#1 of 1): Unsigned compared against 0 (NO_EFFECT)
unsigned_compare: This greater-than-or-equal-to-zero comparison of an unsigned value is always true. n >= 0UL.
124        if (!errno && n >= 0 && n <= 0xFFFF)
125            return n;
126        else
127            return -1;
128    }
```

jiridanek:
Alternate fix to

* https://github.com/skupperproject/skupper-router/pull/1537

I saw that Coverity warning a long time ago already, but I always felt unsure about depending on a variable being of a certain type. Here's a solution that should pass Coverity and it still ensures that `n` is not negative.

Do with it as you will.

Fixes #1536

### DIFF
--- a/tests/cpp/cpp_unit/CMakeLists.txt
+++ b/tests/cpp/cpp_unit/CMakeLists.txt
@@ -40,7 +40,6 @@ add_executable(cpp_unit
         test_amqp.cpp
         test_server.cpp
         test_terminus.cpp
-        test_static_assert.c
 )
 target_link_libraries(cpp_unit cpp-stub pthread skupper-router ${bfd_lib})
 # http_common.h includes "delivery.h"
--- a/tests/cpp/cpp_unit/CMakeLists.txt
+++ b/tests/cpp/cpp_unit/CMakeLists.txt
@@ -40,6 +40,7 @@ add_executable(cpp_unit
         test_amqp.cpp
         test_server.cpp
         test_terminus.cpp
+        test_static_assert.c
 )
 target_link_libraries(cpp_unit cpp-stub pthread skupper-router ${bfd_lib})
 # http_common.h includes "delivery.h"
--- a/tests/static_assert_test.c
+++ b/tests/static_assert_test.c
@@ -17,7 +17,7 @@
  * under the License.
  */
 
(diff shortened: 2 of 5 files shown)


tests/cpp/cpp_unit/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ add_executable(cpp_unit
         test_amqp.cpp
         test_server.cpp
         test_terminus.cpp
+        test_static_assert.c
```
ganeshmurthy, line 43: Can we please add this test as part of unit_test_SOURCES - https://github.com/skupperproject/skupper-router/blob/main/tests/CMakeLists.txt#L28

tests/static_assert_test.c
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "../include/qpid/dispatch/static_assert.h"
```
ganeshmurthy, line 20: This can simply be - 
`#include "qpid/dispatch/static_assert.h"`  ?

tests/cpp/cpp_unit/CMakeLists.txt
jiridanek: yes

tests/static_assert_test.c
jiridanek: done, :facepalm:

jiridanek:
over all, maybe I do like the fix of just removing the check better... anyways, yeah, I won't be overthinking this any more

ganeshmurthy (approved):
LGTM !! Nice work !

PR #1540, opened 2024-07-02
Fixes #1456 - Implemented the administrative ACCESS_POINT and LINK

### LINKED ISSUE #1456: Change to VanFlow LINKs - Making them administrative
Currently, the LINK record in VanFlow is _operational_.  The records only exist when there is an open inter-router link in operation.  There is no positive indication of the desire to have a link that may or may not be currently connected.

This change changes the semantics of LINK and adds a new record type ACCESS_POINT.  These track link connectors and listeners respectively.  LINK has an attribute called OPER_STATUS that is either "down" or "up" depending on whether the link is connected to a peer router or not.  ACCESS_POINT has an attribute called LINK_COUNT that indicates the number of link connections that are connected inbound from peer routers.

Improvements:
- These entities are administrative and follow the lifecycle of the configured listeners and connectors.
- They handle _edge_ links as well as _inter-router_ links (currently, only _inter-router_ links are shown).

kgiusti:
This also separates the closure-of-all-connections from the freeing of the server module so that connections can be closed at an earlier point of the shutdown sequence.

This also adds accepted-connections to the listener reference counts for a more orderly freeing of resources.

Original work by Ted Ross. Ported to amqp adaptor and obfuscated by kgiusti

### DIFF
--- a/src/adaptors/amqp/connection_manager.c
+++ b/src/adaptors/amqp/connection_manager.c
@@ -30,6 +30,7 @@
 #include "qpid/dispatch/ctools.h"
 #include "qpid/dispatch/failoverlist.h"
 #include "qpid/dispatch/threading.h"
+#include "qpid/dispatch/vanflow.h"
 
 #include <proton/listener.h>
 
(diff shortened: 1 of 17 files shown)


src/adaptors/amqp/connection_manager.c
```diff
@@ -414,8 +428,17 @@ QD_EXPORT qd_connector_t *qd_dispatch_configure_connector(qd_dispatch_t *qd, qd_
         item->host_port = malloc(hplen);
         snprintf(item->host_port, hplen, "%s:%s", item->host , item->port);
 
-        DEQ_INSERT_TAIL(ct->conn_info_list, item);
+        //
```
ted-ross, line 431: Should this also be filtered by role (i.e. don't create vanFlow records for 'normal' connectors)?

ted-ross (approved):
Just one minor comment inline.

src/adaptors/amqp/connection_manager.c
kgiusti: Yes it should. I think we're getting lucky here - the CI test checks that tcpConnectors don't add LINK records but that's because the tcpConnector implementation is in the tcp adaptor and this code doesn't apply.

But yeah, better to be explicit here - good catch.

PR #1545, opened 2024-07-08
Fixes #1543 - Added an optional configuration entity the cause the ro…

### LINKED ISSUE #1543: Add capability for the router to emit a vFlow SITE record
In situations where a Skupper site is deployed without a controller (e.g. a static Podman site), there is no controller to emit a SITE record for vFlow.

This feature allows the router to be configured to emit a SITE record based on content in its configuration.

ted-ross:
…uter to emit a SITE record into VanFlow.

### DIFF
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -618,6 +618,36 @@
                 }
             }
         },
+        "site": {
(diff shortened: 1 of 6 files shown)


ganeshmurthy (approved):
LGTM !!

PR #1566, opened 2024-07-12
CI: enhance vflow test client verification API

kgiusti:

PR #1570, opened 2024-07-16
#1563 - Delete connections optionally on deletion of tcpListener/tcpConnector

### LINKED ISSUE #1563: Delete corresponding connections on deletion of tcpListener and tcpConnector
When a tcpListener or tcpConnector is deleted, delete the connections that correspond to these listeners and connectors.

gabordozsa:
Fixes #1563: Delete corresponding connections optionally on deletion of tcpListener and tcpConnector

New config parameters added to tcpListener and tcpConnector entities in order to make deletion of connections optional.

### DIFF
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -1321,13 +1321,6 @@
                     "type": ["up", "down"],
                     "description": "The operational status of TCP socket listener: up - the service is active and incoming connections are permitted; down - the service is not active and incoming connection attempts will be refused.",
                     "create": false
-                },
-                "closeConnectionsOnDelete": {
-                    "type": "boolean",
-                    "default": false,
-                    "description": "Terminate existing connections when the tcpListener is deleted.",
-                    "required": false,
-                    "create": true
                 }
             }
         },
@@ -1394,13 +1387,6 @@
                     "type": "boolean",
                     "default": true,
                     "description": "yes: Ensures that when initiating a connection (as a client) the host name in the URL to which this connector connects to matches the host name in the digital certificate that the peer sends back as part of the TLS connection; no: Does not perform host name verification",
-                    "create": true
-                },
-                "closeConnectionsOnDelete": {
-                    "type": "boolean",
-                    "default": false,
-                    "description": "Terminate existing connections when the tcpConnector is deleted.",
-                    "required": false,
                     "create": true
                 }
             }
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -1321,6 +1321,13 @@
                     "type": ["up", "down"],
                     "description": "The operational status of TCP socket listener: up - the service is active and incoming connections are permitted; down - the service is not active and incoming connection attempts will be refused.",
                     "create": false
+                },
+                "closeConnectionsOnDelete": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Terminate existing connections when the tcpListener is deleted.",
+                    "required": false,
+                    "create": true
                 }
             }
         },
@@ -1387,6 +1394,13 @@
                     "type": "boolean",
                     "default": true,
                     "description": "yes: Ensures that when initiating a connection (as a client) the host name in the URL to which this connector connects to matches the host name in the digital certificate that the peer sends back as part of the TLS connection; no: Does not perform host name verification",
+                    "create": true
+                },
+                "closeConnectionsOnDelete": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Terminate existing connections when the tcpConnector is deleted.",
+                    "required": false,
                     "create": true
                 }
             }
--- a/src/adaptors/adaptor_common.c
+++ b/src/adaptors/adaptor_common.c
@@ -66,8 +66,7 @@ qd_error_t qd_load_adaptor_config(qdr_core_t *core, qd_adaptor_config_t *config,
     config->site_id = qd_entity_opt_string(entity, "siteId", 0);               CHECK();
     config->ssl_profile_name  = qd_entity_opt_string(entity, "sslProfile", 0); CHECK();
     config->authenticate_peer = qd_entity_opt_bool(entity, "authenticatePeer", false); CHECK();
-    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);   CHECK();
-    config->terminate_conns   = qd_entity_opt_bool(entity, "closeConnectionsOnDelete", false);
+    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);
     CHECK();
 
     config->backlog = qd_entity_opt_long(entity, "backlog", 0);
--- a/src/adaptors/adaptor_common.c
+++ b/src/adaptors/adaptor_common.c
@@ -67,6 +67,8 @@ qd_error_t qd_load_adaptor_config(qdr_core_t *core, qd_adaptor_config_t *config,
     config->ssl_profile_name  = qd_entity_opt_string(entity, "sslProfile", 0); CHECK();
     config->authenticate_peer = qd_entity_opt_bool(entity, "authenticatePeer", false); CHECK();
     config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);   CHECK();
+    config->terminate_conns   = qd_entity_opt_bool(entity, "closeConnectionsOnDelete", false);
+    CHECK();
 
     config->backlog = qd_entity_opt_long(entity, "backlog", 0);
     CHECK();
--- a/src/adaptors/tcp/tcp_adaptor.c
+++ b/src/adaptors/tcp/tcp_adaptor.c
@@ -2389,7 +2389,7 @@ QD_EXPORT void qd_dispatch_delete_tcp_listener(qd_dispatch_t *qd, void *impl)
         //
         // Initiate termination of existing connections
         //
-        if (listener->adaptor_config->terminate_conns) {
+        if (tcp_context->qd->terminate_tcp_conns) {
             sys_mutex_lock(&listener->lock);
             qd_tcp_connection_t *conn = DEQ_HEAD(listener->connections);
             while (conn) {
@@ -2428,7 +2428,7 @@ QD_EXPORT void qd_dispatch_delete_tcp_connector(qd_dispatch_t *qd, void *impl)
         //
         // Initiate termination of existing connections
         //
-        if (connector->adaptor_config->terminate_conns) {
+        if (tcp_context->qd->terminate_tcp_conns) {
             sys_mutex_lock(&connector->lock);
             qd_tcp_connection_t *conn = DEQ_HEAD(connector->connections);
             while (conn) {
--- a/src/adaptors/tcp/tcp_adaptor.c
+++ b/src/adaptors/tcp/tcp_adaptor.c
@@ -455,7 +455,7 @@ static void terminate_connections(qd_tcp_connection_list_t *connections)
         // will trigger the closing at link setup in the PN_RAW_CONNECTION_CONNECTED
         // event handler.
         sys_mutex_lock(&conn->activation_lock);
-        if (!!conn->core_conn && IS_ATOMIC_FLAG_SET(&conn->raw_opened))
+        if (IS_ATOMIC_FLAG_SET(&conn->raw_opened) && !!conn->core_conn)
             qdr_core_close_connection(conn->core_conn);
         else
             conn->pending_close = true;
--- a/src/adaptors/tcp/tcp_adaptor.c
+++ b/src/adaptors/tcp/tcp_adaptor.c
@@ -499,17 +499,8 @@ static void free_connection_IO(void *context)
         if (conn->common.parent->context_type == TL_LISTENER) {
             qd_tcp_listener_t *listener = (qd_tcp_listener_t*) conn->common.parent;
             sys_mutex_lock(&listener->lock);
+            DEQ_REMOVE(listener->connections, conn);
             listener->connections_closed++;
-            if (!!conn->core_conn) {
-                // conn is not in the connections list if conn->core_conn is zero
-                DEQ_REMOVE(listener->connections, conn);
-                // Note: core_conn can be accessed when management thread is deleting the listener
-                // and initiates closing the TCP connections. We need to hold the listener->lock when
-                // removing core_conn.
-                qdr_connection_closed(conn->core_conn);
-                conn->core_conn = 0;
-                qd_connection_counter_dec(QD_PROTOCOL_TCP);
-            }
             sys_mutex_unlock(&listener->lock);
             //
             // Call listener decref when a connection associated with the listener is removed (DEQ_REMOVE(listener->connections, conn))
@@ -520,16 +511,7 @@ static void free_connection_IO(void *context)
             qd_tcp_connector_t *connector = (qd_tcp_connector_t*) conn->common.parent;
             sys_mutex_lock(&connector->lock);
             connector->connections_closed++;
-            if (!!conn->core_conn) {
-                // conn is not in the connections list if conn->core_conn is zero
-                DEQ_REMOVE(connector->connections, conn);
-                // Note: core_conn can be accessed when management thread is deleting the connector
-                // and initiates closing the TCP connections. We need to hold the connector->lock when
-                // removing core_conn.
-                qdr_connection_closed(conn->core_conn);
-                conn->core_conn = 0;
-                qd_connection_counter_dec(QD_PROTOCOL_TCP);
-            }
+            DEQ_REMOVE(connector->connections, conn);
             sys_mutex_unlock(&connector->lock);
             //
             // Call connector decref when a connection associated with the connector is removed (DEQ_REMOVE(connector->connections, conn))
@@ -596,8 +578,6 @@ static void close_connection_XSIDE_IO(qd_tcp_connection_t *conn)
         sys_mutex_unlock(&conn->activation_lock);
     }
 
-    free(conn->reply_to);
-
     if (!!conn->inbound_stream) {
         qd_log(LOG_TCP_ADAPTOR, QD_LOG_DEBUG, DLV_FMT " TCP cancel producer activation", DLV_ARGS(conn->inbound_delivery));
         qd_message_cancel_producer_activation(conn->inbound_stream);
@@ -644,10 +624,17 @@ static void close_connection_XSIDE_IO(qd_tcp_connection_t *conn)
         vflow_end_record(conn->common.vflow);
     }
 
+    if (!!conn->core_conn) {
+        qdr_connection_closed(conn->core_conn);
+        conn->core_conn = 0;
+        qd_connection_counter_dec(QD_PROTOCOL_TCP);
+    }
+
     qd_tls_free2(conn->tls);
     qd_tls_domain_decref(conn->tls_domain);
     free(conn->alpn_protocol);
 
+    free(conn->reply_to);
     conn->reply_to          = 0;
     conn->inbound_link      = 0;
     conn->inbound_stream    = 0;
@@ -926,17 +913,8 @@ static void link_setup_LSIDE_IO(qd_tcp_connection_t *conn)
     qdr_terminus_set_dynamic(source);
 
     qd_raw_conn_get_address_buf(conn->raw_conn, host, sizeof(host));
-
-    sys_mutex_lock(&li->lock);
     conn->core_conn = TL_open_core_connection(conn->conn_id, true, host);
     qdr_connection_set_context(conn->core_conn, conn);
-    // Add the tcp conn to the listener list only when core_conn is setup
-    DEQ_INSERT_TAIL(li->connections, conn);
-    if (li->closing)
-        // listener is being deleted and all connections must be closed
-        qdr_core_close_connection(conn->core_conn);
-    sys_mutex_unlock(&li->lock);
-
     conn->inbound_link = qdr_link_first_attach(conn->core_conn, QD_INCOMING, qdr_terminus(0), target, "tcp.lside.in", 0, false, 0, &conn->inbound_link_id);
     qdr_link_set_context(conn->inbound_link, conn);
     conn->outbound_link = qdr_link_first_attach(conn->core_conn, QD_OUTGOING, source, qdr_terminus(0), "tcp.lside.out", 0, false, 0, &conn->outbound_link_id);
@@ -951,18 +929,9 @@ static void link_setup_CSIDE_IO(qd_tcp_connection_t *conn, qdr_delivery_t *deliv
     ASSERT_RAW_IO;
 
     assert(conn->common.parent->context_type == TL_CONNECTOR);
-    qd_tcp_connector_t *connector = (qd_tcp_connector_t *) conn->common.parent;
-    const char         *host      = connector->adaptor_config->host_port;
-
-    sys_mutex_lock(&connector->lock);
-    conn->core_conn = TL_open_core_connection(conn->conn_id, false, host);
+    const char *host = ((qd_tcp_connector_t *) conn->common.parent)->adaptor_config->host_port;
+    conn->core_conn  = TL_open_core_connection(conn->conn_id, false, host);
     qdr_connection_set_context(conn->core_conn, conn);
-    // Add the tcp conn to the connector list only when core_conn is setup
-    DEQ_INSERT_TAIL(connector->connections, conn);
-    if (connector->closing)
-        // connector is being deleted and all connections must be closed
-        qdr_core_close_connection(conn->core_conn);
-    sys_mutex_unlock(&connector->lock);
 
     // use an anonymous inbound link in order to ensure credit arrives otherwise if the client has dropped the state machine will stall waiting for credit
     conn->inbound_link = qdr_link_first_attach(conn->core_conn, QD_INCOMING, qdr_terminus(0), qdr_terminus(0), "tcp.cside.in", 0, false, 0, &conn->inbound_link_id);
@@ -1254,6 +1223,7 @@ static uint64_t handle_first_outbound_delivery_CSIDE(qd_tcp_connector_t *connect
     conn->context.handler = on_connection_event_CSIDE_IO;
 
     sys_mutex_lock(&connector->lock);
+    DEQ_INSERT_TAIL(connector->connections, conn);
     connector->connections_opened++;
     vflow_set_uint64(connector->common.vflow, VFLOW_ATTRIBUTE_FLOW_COUNT_L4, connector->connections_opened);
     vflow_set_ref_from_record(conn->common.vflow, VFLOW_ATTRIBUTE_CONNECTOR, connector->common.vflow);
@@ -1750,6 +1720,7 @@ static void connection_run_LSIDE_IO(qd_tcp_connection_t *conn)
             //
             // Don't do anything
             //
+            return;
             break;
 
         default:
@@ -1757,6 +1728,18 @@ static void connection_run_LSIDE_IO(qd_tcp_connection_t *conn)
             break;
         }
     } while (repeat);
+
+    bool closing = false;
+    if (conn->common.parent)
+        closing = IS_ATOMIC_FLAG_SET(&((qd_tcp_listener_t *) conn->common.parent)->closing);
+    if (closing) {
+        if (!!conn->core_conn) {
+            qdr_core_close_connection(conn->core_conn);
+        } else {
+            close_raw_connection(conn, "Parent-deleted", "Forced closed");
+            set_state_XSIDE_IO(conn, XSIDE_CLOSING);  // prevent further connection I/O
+        }
+    }
 }
 
 
@@ -1818,13 +1801,26 @@ static void connection_run_CSIDE_IO(qd_tcp_connection_t *conn)
             //
             // Don't do anything
             //
+            return;
             break;
 
         default:
             assert(false);
             break;
-    }
+        }
     } while(repeat);
+
+    bool closing = false;
+    if (conn->common.parent)
+        closing = IS_ATOMIC_FLAG_SET(&((qd_tcp_connector_t *) conn->common.parent)->closing);
+    if (closing) {
+        if (!!conn->core_conn) {
+            qdr_core_close_connection(conn->core_conn);
+        } else {
+            close_raw_connection(conn, "Parent-deleted", "Forced closed");
+            set_state_XSIDE_IO(conn, XSIDE_CLOSING);  // prevent further connection I/O
+        }
+    }
 }
 
 
@@ -2052,6 +2048,7 @@ static void on_accept(qd_adaptor_listener_t *adaptor_listener, pn_listener_t *pn
     conn->context.handler = on_connection_event_LSIDE_IO;
 
     sys_mutex_lock(&listener->lock);
+    DEQ_INSERT_TAIL(listener->connections, conn);
     listener->connections_opened++;
     vflow_set_uint64(listener->common.vflow, VFLOW_ATTRIBUTE_FLOW_COUNT_L4, listener->connections_opened);
     sys_mutex_unlock(&listener->lock);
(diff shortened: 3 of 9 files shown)


src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -1936,6 +1936,28 @@ static int setup_tls_session(qd_tcp_connection_t *conn, const qd_tls_domain_t *p
     return 0;
 }
 
+/**
+ * Initiate the termination of a group of connections.
+ *
+ * @param connections Reference to the list of connections
+ */
+static void terminate_connections(qd_tcp_connection_list_t *connections)
```
ganeshmurthy, line 1944: There is a function called `close_connection_XSIDE_IO()` which can be used to terminate tcp connections. It take care of freeing `qd_tcp_connection_t` objects and well as its corresponding `qdr_connection_t` objects
Like so - 
```
static void terminate_listener_connections(qd_tcp_listener_t *listener) {
        //
        // Deliberately call qd_tcp_listener_incref() to make sure that freeing the connections, don't
        // free the listener. Then we call qd_tcp_listener_free() to forcefully free the listener without checking the listener->ref_count
        //
        qd_tcp_listener_incref(listener);
        qd_tcp_connection_t *conn = DEQ_HEAD(listener->connections);
        while (conn) {
            qd_tcp_connection_t *next_conn = DEQ_NEXT(conn);
            close_connection_XSIDE_IO(conn);
            conn = next_conn;
        }
        qd_tcp_listener_free(listener);
}
```

ganeshmurthy:
Also, this PR needs to include a system test using vanflow -
The test must start two routers, create a  bunch of connections from a tcpListener on one router and then delete the tcpListener and validate that the associated connections are gone, same with tcpConnector (on the other router). The checking for the presence and deletion of a tcpConnection can be done using the vanflow API.

src/adaptors/tcp/tcp_adaptor.c
gabordozsa: I thought that `close_connection_XSIDE_IO()` cannot be called directly in a loop because it makes proton calls on the connection. Another proactor thread may process the same connection at the same time which is prohibited. I maybe misunderstood this paragraph from `threading-info.txt`: https://github.com/skupperproject/skupper-router/blob/35d5a0fae00af782663e95f4ebea69299bb7edd6/docs/notes/threading-info.txt#L50

ganeshmurthy: No, you did not misunderstand the paragraph. What I suggested - calling close_connection_XSIDE_IO() - from a management thread - is wrong. I will get back to you shortly

ganeshmurthy: You might try calling `qdr_core_close_connection(conn->core_conn)`
like here - https://github.com/skupperproject/skupper-router/blob/main/src/adaptors/http2/http2_adaptor.c#L3403

gabordozsa: Okay, thanks for the pointer! If I call that function then I do not need to change the type of the `closed` flag from `bool` to `sys_atomic_t` in `qdr_connection_t`. Is there any other reason why the current PR is not a proper solution? I'm just asking because I would like to understand the router code better :-).

ganeshmurthy: Yes, if you call `qdr_core_close_connection(conn->core_conn)`, you will not need to change the type on the closed flag to `sys_atomic_t` in `qdr_connection_t`. 

Notice that the closed flag is accessed in two places, inside the core thread callbacks and in the `qdr_connection_process()` under a `conn->work_lock`.
The core thread is single threaded and so most of the core thread data structures don't need locking when accessed by the core thread except during connection activation which happens on an I/O thread. 

We usually tend to have I/O thread functions which then employ the core thread to modify these data structures via core thread callbacks. So, we generally access `conn->closed` from a core thread callback and `qdr_core_close_connection(conn->core_conn)` does that for you.

gabordozsa:
> Also, this PR needs to include a system test using vanflow - The test must start two routers, create a bunch of connections from a tcpListener on one router and then delete the tcpListener and validate that the associated connections are gone, same with tcpConnector (on the other router). The checking for the presence and deletion of a tcpConnection can be done using the vanflow API.

I'm working on this system test now (so I converted the PR to draft for now). Could you point me to an existing system test which could be a good starting point for me for this new test?

src/adaptors/tcp/tcp_adaptor.c
ganeshmurthy: Two things
1. You don't need to call `pn_raw_connection_wake()` since it is already called as part of the call to `qdr_core_close_connection()`
2. You also don't need to lock on the activaton_lock since you are not going to call `pn_raw_connection_wake()` anymore

ganeshmurthy:
> > Also, this PR needs to include a system test using vanflow - The test must start two routers, create a bunch of connections from a tcpListener on one router and then delete the tcpListener and validate that the associated connections are gone, same with tcpConnector (on the other router). The checking for the presence and deletion of a tcpConnection can be done using the vanflow API.
> 
> I'm working on this system test now (so I converted the PR to draft for now). Could you point me to an existing system test which could be a good starting point for me for this new test?

Please take a look at https://github.com/skupperproject/skupper-router/pull/1566 and see if you understand how the vanflow_snooper.py works and let me know if you have any questions

ganeshmurthy:
@gabordozsa Just wanted to mention that if you are working on a github issue, please make sure you assign the issue to yourself before you start working on it. That will let other developers know who is responsible for this issue, so they will not work on that issue. Also if the issue is already assigned to someone, and you want to work on it, please contact that person to see if they are not already working on it, if not, you can that it be assigned to you. Thanks.

src/adaptors/tcp/tcp_adaptor.c
gabordozsa: Just to make sure that I understand correctly the code, would the call sequence be roughly be as follows:

1.  `qdr_core_close_connection() ` will activate the core thread
2.  qdr_core_close_connection_CT() will set `conn->closed=true` and activate the connection
3. the TCP adaptor activation callback `CORE_activate()` will call `pn_raw_connection_wake()`

I think the check for `conn->raw_opened` can also be removed from  `terminate_connections()` because the same check is done in `CORE_active()`.

gabordozsa:
> @gabordozsa Just wanted to mention that if you are working on a github issue, please make sure you assign the issue to yourself before you start working on it. That will let other developers know who is responsible for this issue, so they will not work on that issue. Also if the issue is already assigned to someone, and you want to work on it, please contact that person to see if they are not already working on it, if not, you can that it be assigned to you. Thanks.

Sure, I will do! Thanks for the reminder and for assigning 1563 to me!

src/adaptors/tcp/tcp_adaptor.c
ganeshmurthy: 1. `qdr_core_close_connection(qdr_connection_t *conn)` will add a core thread action with callback `qdr_core_close_connection_CT()` and wake up the core thread 
2. The `qdr_core_close_connection_CT()` is executed by the core thread and will set the  `conn->closed=true` and activate the connection to be deleted via a call to `qdr_connection_activate_CT(core,  conn);` 
3. `qdr_connection_process()' in `connections.c` is executed on an I/O thread and an adaptor callback function is called (line 359 of connections.c) - `conn->protocol_adaptor->conn_close_handler(conn->protocol_adaptor->user_context, conn, conn->error);`
4. The `conn_close_handler` is set here - https://github.com/skupperproject/skupper-router/blob/main/src/adaptors/tcp/tcp_adaptor.c#L2548 which is `static void CORE_connection_close(void *context, qdr_connection_t *conn, qdr_error_t *error)` where there is a call to `pn_raw_connection_close(tcp_conn->raw_conn);` which will generate a PN_RAW_CONNECTION_DISCONNECTED proton event and the handler for that event ultimately calls `close_connection_XSIDE_IO(conn)` where the `qd_tcp_connection_t` object and its contents are ultimately freed.

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -2420,6 +2431,17 @@ QD_EXPORT void qd_dispatch_delete_tcp_connector(qd_dispatch_t *qd, void *impl)
         connector->core_conn = 0;
         qd_connection_counter_dec(QD_PROTOCOL_TCP);
         //
+        // Initiate termination of existing connections
+        //
+        if (connector->adaptor_config->terminate_conns) {
+            qd_tcp_connection_t *conn;
+            conn = DEQ_HEAD(listener->connections);
```
ganeshmurthy, line 2438: This should be `conn = DEQ_HEAD(connector->connections)'  not `conn = DEQ_HEAD(listener->connections)`?

src/adaptors/tcp/tcp_adaptor.c
gabordozsa: Thanks!

src/adaptors/tcp/tcp_adaptor.c
gabordozsa: Ops, copy-paste issue.

gabordozsa:
@ganeshmurthy  I have added a system test. I think the PR is ready for more reviews.

tests/system_tests_tcp_conns_terminate.py
```diff
@@ -0,0 +1,387 @@
+#
```
ganeshmurthy, line 1: @gabordozsa can you please add this test to the list of tests in tests/CMakeLists.txt ? This test is not currently executing in CI

tests/system_tests_tcp_conns_terminate.py
```diff
@@ -0,0 +1,387 @@
...
+        """
+        Setup tcp flows via the tcpListeners and the tcpConnector which have
+        the particular VAN address. Two flows are created for each tcpListener.
+        """
+        router_1_id = self.router_1.config.router_id
+        router_2_id = self.router_2.config.router_id
+
+        # vflow Ids are used to associate flows with tcpListeners and tcpConnectors
+        vflow_ids = {}
+        self.assertTrue(self.router_1_vflow_id is not None)
```
ganeshmurthy, line 239: You can use `self.assertIsNotNone` instead of `self.assertTrue`

tests/system_tests_tcp_conns_terminate.py
```diff
@@ -0,0 +1,387 @@
...
+        self.assertTrue(self.router_2_vflow_id is not None)
+        vflow_ids['listener_1'] = self.get_tcp_entity_vflow_id(self.router_1_vflow_id,
+                                                               'LISTENER', address)
+        vflow_ids['listener_2'] = self.get_tcp_entity_vflow_id(self.router_2_vflow_id,
+                                                               'LISTENER', address)
+        self.assertTrue(vflow_ids['listener_1'] is not None)
```
ganeshmurthy, line 245: Same as above, use  `self.assertIsNotNone` instead of `self.assertTrue`

ganeshmurthy:
This test must also include another scenario where a TLS enabled `tcpListener` and TLS enabled `tcpConnector` are deleted.  You can attach an `sslProfile` to a `tcpListener` or a `tcpConnector` to make them do TLS. Examples of this can be found in `system_tests_tcp_adaptor_tls`

tests/system_tests_tcp_conns_terminate.py
gabordozsa: I added it to the list.

ganeshmurthy:
To test the TLS case, you can use the `EchoClientRunner` to start the echo client which can be found in `system_tests_tcp_adaptor`. It has the `test_ssl` flag which can be set to True

tests/system_tests_tcp_conns_terminate.py
```diff
@@ -0,0 +1,387 @@
...
+        res = self.snooper_thread.get_results()
+        try:
+            for rec in res[router_vflow_id]:
+                if rec['RECORD_TYPE'] == record_type and rec['PROTOCOL'] == 'tcp' and rec['VAN_ADDRESS'] == address:
+                    return rec['IDENTITY']
+        except KeyError:
+            pass
+        return None
+
+    def run_skmanage(self, cmd, router):
```
ganeshmurthy, line 178: There is a class in `system_test.py` called `QdManager` that runs the `skmanage` command for you. You can use that class instead of directly running `skmanage`

ganeshmurthy:
@gabordozsa I introduced a commit - https://github.com/skupperproject/skupper-router/commit/f1f61ee2f5cbdc41e16d45f755eec26f5125000c - where I renamed the `QdManager` class to `SkManager`  so it is more in line with the actual name of the tool `skmanage`
Please rebase with the latest main branch and use the SkManager class in your PR

tests/system_tests_tcp_conns_terminate.py
gabordozsa: Done

tests/system_tests_tcp_conns_terminate.py
gabordozsa: Done

tests/system_tests_tcp_conns_terminate.py
gabordozsa: I rebased and rebased and removed the run_skmanage() method.

ganeshmurthy:
@gabordozsa This looks good to me, great work ! 
I want @kgiusti to look at this and once he approves it, we can push this PR to main branch. He is back on Monday !

gabordozsa:
> @gabordozsa This looks good to me, great work ! I want @kgiusti to look at this and once he approves it, we can push this PR to main branch. He is back on Monday !

@ganeshmurthy Thanks! I pushed another commit to take the listener/connector lock when traversing the connections list. I think this is necessary because the list might be modified by another thread if a new connection is created or an existing one is closed. Does that make sense?

ganeshmurthy:
> Does that make sense?

Yes. Excellent catch.

grs:
Re-asking a question here from https://github.com/skupperproject/skupper/pull/1558#issuecomment-2260242047:

Is there a reason for having this defined per connector/listener rather than either being simply the default behaviour or globally configured?

gabordozsa:
> Re-asking a question here from [skupperproject/skupper#1558 (comment)](https://github.com/skupperproject/skupper/pull/1558#issuecomment-2260242047):
> 
> Is there a reason for having this defined per connector/listener rather than either being simply the default behaviour or globally configured?

My understanding is that we should keep the current default behaviour for v1. As for using a global config flag, the reasoning  was that it would not simplify the implementation.

ganeshmurthy:
> Is there a reason for having this defined per connector/listener rather than either being simply the default behaviour or globally configured?

@grs I agree. I think we can move the closeConnectionsOnDelete flag to the global `router` entity instead of having it in each `tcpConnector/tcpListener`. Is that what you had in mind ?

grs:
> @grs I agree. I think we can move the closeConnectionsOnDelete flag to the global `router` entity instead of having it in each `tcpConnector/tcpListener`. Is that what you had in mind ?

That seems simpler for users who want to alter the default behaviour.

ganeshmurthy:
> > @grs I agree. I think we can move the closeConnectionsOnDelete flag to the global `router` entity instead of having it in each `tcpConnector/tcpListener`. Is that what you had in mind ?
> 
> That seems simpler for users who want to alter the default behaviour.

Agree !  I think we should modify this PR to move the flag to the `router` entity.
@gabordozsa @kgiusti What do  you think ?

gabordozsa:
> > > @grs I agree. I think we can move the closeConnectionsOnDelete flag to the global `router` entity instead of having it in each `tcpConnector/tcpListener`. Is that what you had in mind ?
> > 
> > 
> > That seems simpler for users who want to alter the default behaviour.
> 
> Agree ! I think we should modify this PR to move the flag to the `router` entity. @gabordozsa @kgiusti What do you think ?

I'm fine with that. The flag may look at bit out of place in the 'router' entity because it is only relevant to tcpListeners and tcpConnectors. But I agree that would make easier to change the default behaviour.

ganeshmurthy:
> > > > @grs I agree. I think we can move the closeConnectionsOnDelete flag to the global `router` entity instead of having it in each `tcpConnector/tcpListener`. Is that what you had in mind ?
> > > 
> > > 
> > > That seems simpler for users who want to alter the default behaviour.
> > 
> > 
> > Agree ! I think we should modify this PR to move the flag to the `router` entity. @gabordozsa @kgiusti What do you think ?
> 
> I'm fine with that. The flag may look at bit out of place in the 'router' entity because it is only relevant to tcpListeners and tcpConnectors. But I agree that would make easier to change the default behaviour.

@gabordozsa There are already some fields in the `router` entity namely `defaultDistribution`, `dataConnectionCount` etc that are not directly pertinent to the `router` itself. We have used the `router` entity in the past to add attributes that apply to the router globally.

gabordozsa:
@ganeshmurthy I moved the new config flag into the 'router' entity and adjusted the system tests.

src/adaptors/adaptor_common.c
```diff
@@ -66,7 +66,9 @@ qd_error_t qd_load_adaptor_config(qdr_core_t *core, qd_adaptor_config_t *config,
     config->site_id = qd_entity_opt_string(entity, "siteId", 0);               CHECK();
     config->ssl_profile_name  = qd_entity_opt_string(entity, "sslProfile", 0); CHECK();
     config->authenticate_peer = qd_entity_opt_bool(entity, "authenticatePeer", false); CHECK();
-    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);   CHECK();
+    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);
+    CHECK();
+    CHECK();
```
ganeshmurthy, line 71: There are two calls to `CHECK();`, should only be one. This file `src/adaptors/adaptor_common.c` should not appear in the list of changed files because the the change to obtain `terminate_tcp_conns` flag is in `dispatch.c`

gabordozsa: I removed the extra CHECK(). The file still is in the changeset because `git clang-format` moves the `CHECK()` to the next line.

gabordozsa:
I renamed the new config flag to `dropTcpConnections` to be aligned with https://github.com/skupperproject/skupper/pull/1564

python/skupper_router/management/skrouter.json
```diff
@@ -623,6 +623,13 @@
                     "required": false,
                     "default": "auto",
                     "create": true
+                },
+                "dropTcpConnections": {
+                    "type": "boolean",
+                    "default": false,
```
ganeshmurthy, line 629: Since this PR is going to land on main which is going to be released as Skupper 2.0 (skupper-router 3.0), @grs suggested that the default for this flag can be true. 

Since 2.0 is a major version, we are not going to guarantee backward compatibility, so defaulting the flag to true is ok to do.

ganeshmurthy: When this PR is back ported to 1.8 version of Skupper, we will set the default on this flag to false, since that will ensure backward compatibility.

1.8 is the upcoming upstream release of Skupper

gabordozsa: I changed the default value to true.

ganeshmurthy:
I see this CI error - 
```
 /home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:1288: RuntimeError
---------------------- generated xml file: /__w/skupper-router/skupper-router/skupper-router/build/tests/junitxmls/system_tests_grpc.xml -----------------------
=================================================================== short test summary info ====================================================================
ERROR ::GrpcServiceMethodsTestOverTcpTls::test_grpc_04_client_stream - RuntimeError: Errors during teardown: 

Process 3933 (name=grpc-test-router) error: returned error code -11
skrouterd -c grpc-test-router.conf -I /home/runner/work/skupper-router/skupper-router/skupper-router/python
/__w/skupper-router/skupper-router/skupper-router/build/tests/system_test.dir/tests/system_tests_grpc/GrpcServiceMethodsTestOverTcpTls/setUpClass/grpc-test-router-5.cmd
>>>>
/home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/connections.c:338:5: runtime error: member access within null pointer of type 'struct qdr_connection_t'
    #0 0x66813f in qdr_core_close_connection /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/connections.c:338
    #1 0x5146f6 in qd_dispatch_delete_tcp_listener /home/runner/work/skupper-router/skupper-router/skupper-router/src/adaptors/tcp/tcp_adaptor.c:2397
    #2 0x7f54d3082055 in ffi_call_unix64 (/lib64/libffi.so.8+0x9055) (BuildId: a190bf03e644181cadab122962ab83ae96271696)
    #3 0x7f54d307e69f in ffi_call_int.lto_priv.0 (/lib64/libffi.so.8+0x569f) (BuildId: a190bf03e644181cadab122962ab83ae96271696)
    #4 0x7f54d30814ed in ffi_call (/lib64/libffi.so.8+0x84ed) (BuildId: a190bf03e644181cadab122962ab83ae96271696)
    #5 0x7f54d30992a7 in _ctypes_callproc (/usr/lib64/python3.11/lib-dynload/_ctypes.cpython-311-x86_64-linux-gnu.so+0x102a7) (BuildId: e105d1b113f81c5a55ea7e4b1def4168d56667da)
    #6 0x7f54d3098c80 in PyCFuncPtr_call (/usr/lib64/python3.11/lib-dynload/_ctypes.cpython-311-x86_64-linux-gnu.so+0xfc80) (BuildId: e105d1b113f81c5a55ea7e4b1def4168d56667da)
    #7 0x7f54d555a972 in _PyObject_MakeTpCall (/lib64/libpython3.11.so.1.0+0x131972) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #8 0x7f54d55630fe in _PyEval_EvalFrameDefault (/lib64/libpython3.11.so.1.0+0x13a0fe) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #9 0x7f54d555f44e in _PyEval_Vector (/lib64/libpython3.11.so.1.0+0x13644e) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #10 0x7f54d5597a8a in method_vectorcall (/lib64/libpython3.11.so.1.0+0x16ea8a) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #11 0x7f54d5557ba6 in _PyObject_CallFunctionVa (/lib64/libpython3.11.so.1.0+0x12eba6) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #12 0x7f54d5557adf in PyObject_CallFunction (/lib64/libpython3.11.so.1.0+0x12eadf) (BuildId: dc5b8f1564a7d29af984bfb632f884ae4c52988d)
    #13 0x6165c5 in qd_io_rx_handler /home/runner/work/skupper-router/skupper-router/skupper-router/src/python_embedded.c:658
    #14 0x6bfbd2 in qdr_forward_on_message /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/forwarder.c:375
    #15 0x6eeac0 in qdr_general_handler /home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/router_core.c:1031
    #16 0x7572e2 in qd_timer_visit /home/runner/work/skupper-router/skupper-router/skupper-router/src/timer.c:317
    #17 0x752288 in handle_proactor_other_event /home/runner/work/skupper-router/skupper-router/skupper-router/src/server.c:142
    #18 0x7524bf in proactor_thread /home/runner/work/skupper-router/skupper-router/skupper-router/src/server.c:194
    #19 0x6124b5 in _thread_init /home/runner/work/skupper-router/skupper-router/skupper-router/src/posix/threading.c:207
    #20 0x7f54d595af95 in asan_thread_start(void*) (/lib64/libasan.so.8+0x5df95) (BuildId: 79824421bd82bb3ef4addf048e1265e2a93cfc64)
    #21 0x7f54d4494506 in start_thread (/lib64/libc.so.6+0x97506) (BuildId: 8f53abaad945a669f2bdcd25f471d80e077568ef)
    #22 0x7f54d451840b in clone3 (/lib64/libc.so.6+0x11b40b) (BuildId: 8f53abaad945a669f2bdcd25f471d80e077568ef)
```

https://github.com/skupperproject/skupper-router/actions/runs/10254149016/job/28368308564?pr=1570#step:34:5924

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -2387,6 +2387,19 @@ QD_EXPORT void qd_dispatch_delete_tcp_listener(qd_dispatch_t *qd, void *impl)
             listener->common.vflow = 0;
         }
         //
+        // Initiate termination of existing connections
+        //
+        if (tcp_context->qd->terminate_tcp_conns) {
+            sys_mutex_lock(&listener->lock);
+            qd_tcp_connection_t *conn = DEQ_HEAD(listener->connections);
```
ganeshmurthy, line 2394: Each member of `listener->connections` must be removed so that at the end of this loop `listener->connections` will be empty

Something like this - 

```
sys_mutex_lock(&listener->lock);
 qd_tcp_connection_t *conn = DEQ_HEAD(listener->connections);

 while (conn) {
     qd_tcp_connection_t *next_conn = DEQ_NEXT(conn);
     DEQ_REMOVE_HEAD(listener->connections);
     qdr_core_close_connection(conn->core_conn);
     conn = next_conn;
 }
 sys_mutex_unlock(&listener->lock);
```

ganeshmurthy: Please ignore this comment

gabordozsa:
```
/home/runner/work/skupper-router/skupper-router/skupper-router/src/router_core/connections.c:338:5: runtime error: member access within null pointer of type 'struct qdr_connection_t'
```
The  error points to this line: https://github.com/gabordozsa/skupper-router/blob/0f07033f08a7a98d5ade343c13763271920dc5b5/src/router_core/connections.c#L338
It seems that `conn->core_conn` is null pointer when `qdr_core_close_connection(conn->core_conn)` is called in the connections traversing loop.   That may happen if` close_connection_XSIDE_IO()` has already called for this `conn`.  Does that make sense? If so, I do not see how removing the this `conn` from the `connections` list would fix this.

gabordozsa: @ganeshmurthy  I pushed two new commits as an attempt to fix this error.

1.  I take the` conn->activation_lock` and check if `conn->raw_opened` is set before calling `qdr_core_close_connection(conn->core_conn)`.  The reasoning here is that `close_connection_XSIDE_IO()` can zero  `conn->core_conn`  but it also clears  `conn->raw_opened` before. If the ` conn->activation_lock` is taken then `close_connection_XSIDE_IO()` cannot change the value of `conn->core_conn` while we are calling `qdr_core_close_connection(conn->core_conn)`.
2. I  also check if `conn->core_conn` is not zero before calling `qdr_core_close_connection(conn->core_conn)`. This is because `conn->core_conn` is initialised when the the link is created in ` link_setup_[L,C]SIDE_IO ()` but the `conn->raw_opened` is set before that at the PN_RAW_CONNECTION_CONNECTED event. 

Does that make sense? I think that still more work is necessary for this PR because the TCP connection is not terminated if case (2) happens (i.e. tcpListener is deleted which has a connection with 'raw_opened' set but 'core_conn' is still zero because the link is not setup yet).

ganeshmurthy: Yes, that makes sense @gabordozsa 

I would like to state the problem in simple terms - 
The `PN_RAW_CONNECTION_CONNECTED` event or the `PN_RAW_CONNECTION_DISCONNECTED` event could come upon any of the connections in `listener->connections` while the listener is being deleted.

Case 1: When a connection receives a `PN_RAW_CONNECTION_DISCONNECTED` event at the same time of the listener deletion (listener deletion call runs on a management thread and `PN_RAW_CONNECTION_DISCONNECTED` event comes in via an i/o thread),  your if statement - 
`if (!!conn->core_conn && IS_ATOMIC_FLAG_SET(&conn->raw_opened))` will make sure that we don't call `qdr_core_close_connection(conn->core_conn)` on an already closed connection. So, this case is covered.

Case 2: A connection receives a `PN_RAW_CONNECTION_CONNECTED` event at the same time of the listener deletion. The listener delete function might miss deleting the connection because `conn->core_conn` is not set yet. This will lead to a leaking `qd_tcp_connection_t` and `qdr_connection_t`.

gabordozsa: @ganeshmurthy  I added a new `pending_close` flag to `qd_tcp_connection_t` as per your suggestion. The `connection_run_[L,C]SIDE_IO()` functions closing the connection if this flag is set. I also set the existing `closing` flag in `tcp_listener` and `tcp_connector` in order to set the `pending_close` flag for connections which are not in the `connections` list yet (e.g. when `on_accept()` created the new `conn` but has not called `DEQ_INSERT_TAIL(conn)` yet). I think the `closing` flag was unused before.

ganeshmurthy:
@gabordozsa there seem to be some CI failures in this PR like this one - https://github.com/skupperproject/skupper-router/actions/runs/10354198908/job/28658907680?pr=1570

I have kicked off the main branch CI just to make sure that it is clean, let's see what happens on main branch CI.

ganeshmurthy:
@gabordozsa The main CI seems fine

src/adaptors/adaptor_common.c
```diff
@@ -66,7 +66,8 @@ qd_error_t qd_load_adaptor_config(qdr_core_t *core, qd_adaptor_config_t *config,
     config->site_id = qd_entity_opt_string(entity, "siteId", 0);               CHECK();
     config->ssl_profile_name  = qd_entity_opt_string(entity, "sslProfile", 0); CHECK();
     config->authenticate_peer = qd_entity_opt_bool(entity, "authenticatePeer", false); CHECK();
-    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);   CHECK();
+    config->verify_host_name  = qd_entity_opt_bool(entity, "verifyHostname", false);
+    CHECK();
```
ganeshmurthy, line 70: Is clang format causing this diff to appear ? Can you simply restore this file to its original version and not run the clang formatter ?

ganeshmurthy:
Leak of `qd_tls_domain_t` object
https://github.com/skupperproject/skupper-router/actions/runs/10355435503/job/28663084819?pr=1570#step:34:5092

gabordozsa:
> Leak of `qd_tls_domain_t` object https://github.com/skupperproject/skupper-router/actions/runs/10355435503/job/28663084819?pr=1570#step:34:5092

@ganeshmurthy  In the same error log I can  see (https://github.com/skupperproject/skupper-router/actions/runs/10355435503/job/28663084819?pr=1570#step:34:4777):
```
alloc.c: Items of type 'qd_tcp_connection_t' remain allocated at shutdown: 1
Leak: 2024-08-12 16:14:00.677874 +0000 type: qd_tcp_connection_t address: 0x5160001f9610
```
I'm wondering if the problem could be  in `ADAPTOR_final()` https://github.com/gabordozsa/skupper-router/blob/2c3629b9e17b5a9ce0bb7591d6d9e2dd9b715a52/src/adaptors/tcp/tcp_adaptor.c#L2648. `ADAPTOR_final()` iterates through the connections list and trigger closing them. In the main branch every accepted connection is added to the connections list. However, we delay adding it to the list  now (until the link is setup).  So if a new connection is 
 accepted but `ADAPTOR_final()` is called before the link is setup for it then this conn can leak. Does that make sense?

ganeshmurthy:
> However, we delay adding it to the list now (until the link is setup). So if a new connection is
> accepted but `ADAPTOR_final()` is called before the link is setup for it then this conn can leak. Does that make sense?

Good catch @gabordozsa . It looks like the on_accept() was called and the connection was created there but before we could get a PN_RAW_CONNECTION_CONNECTED event from proton, the router seems to be shutting down which leaves the newly created connection (which is not yet in the list) to leak.

If you ask me, this should not really be a problem. The router is anyway shutting down, who cares about the leak ? But unfortunately, this is how we do things in the router.

gabordozsa:
Closing this as code got merged via https://github.com/skupperproject/skupper-router/pull/1601

PR #1575, opened 2024-07-22
Fixes #1205: Spilt the router entity into Router and RouterMetrics. The…

### LINKED ISSUE #1205: Top down management QUERY retrieves invalid data for Router management entity
The io.skupper.router.router management object used to be implemented in the python code. It has been moved to C (see src/router_core/agent_router.c).  However the old python RouterEntity is still registered. Turns out if you issue a management query against the short form "router" object name (rather than the proper 'io.skupper.router.router' name), the query is handled by the old python RouterEntity. The response body is incorrect and reports invalid values. No management error is generated.

Remove RouterEntity from python/skupper_router_internal/management/agent.py and nuke the src/router_agent.c file from orbit.

It's the only way to be sure...

ganeshmurthy:
… RouterStats entity is handled by the C management agent.


python/skupper_router/management/skrouter.json
```diff
@@ -368,6 +368,143 @@
...
+                    "type": "integer",
+                    "description": "Number of links attached to the router node.",
+                    "graph": true
+                },
+                "nodeCount": {
+                    "type": "integer",
+                    "description": "Number of known peer router nodes.",
+                    "graph": true
+                },
+                "linkRouteCount": {
```
kgiusti, line 400: Can we drop linkRouteCount since this is a new record?

python/skupper_router/management/skrouter.json
```diff
@@ -368,6 +368,143 @@
...
+                    "type": "integer",
+                    "description": "Number of link routes attached to the router node.",
+                    "graph": true
+                },
+                "autoLinkCount": {
```
kgiusti, line 405: Is this also a deprecated feature - I honestly do not recall...

python/skupper_router/management/skrouter.json
```diff
@@ -368,6 +368,143 @@
             "attributes": {}
         },
 
+        "routerStats": {
```
kgiusti, line 371: Instead of "routerStats" what do you think about "routerMetrics"?  I'd run this past The Grand Poobah of All Things Naming  (@ssorj).  Just an idea - not a request.

kgiusti (approved):

python/skupper_router/management/skrouter.json
ganeshmurthy: Renamed routerStats to routerMetrics - done

python/skupper_router/management/skrouter.json
ganeshmurthy: Removed linkRouteCount

python/skupper_router/management/skrouter.json
ganeshmurthy: We are keeping autoLinkCount as per @ted-ross

PR #1598, opened 2024-08-14
Issue 1563 1

ganeshmurthy:


src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -2051,17 +2046,19 @@ static void on_accept(qd_adaptor_listener_t *adaptor_listener, pn_listener_t *pn
...
     conn->context.handler = on_connection_event_LSIDE_IO;
 
+    conn->raw_conn = pn_raw_connection();
+    pn_raw_connection_set_context(conn->raw_conn, &conn->context);
+    if (listener->protocol_observer) {
+        conn->observer_handle = qdpo_begin(listener->protocol_observer, conn->common.vflow, conn, conn->conn_id);
+    }
+
     sys_mutex_lock(&listener->lock);
+    DEQ_INSERT_TAIL(listener->connections, conn);
```
gabordozsa, line 2056: Can the tcpListener delete loop be executed by the mgmt thread just before the `listener->lock` was taken here?   This new` conn` will not have the `management_close` flag set then so it will not be terminated.

src/adaptors/tcp/tcp_adaptor.c
```diff
@@ -2017,6 +2000,18 @@ static void on_connection_event_CSIDE_IO(pn_event_t *e, qd_server_t *qd_server,
     }
 
     connection_run_CSIDE_IO(conn);
+
+    qd_tcp_listener_t *connector = (qd_tcp_listener_t*) conn->common.parent;
+    sys_mutex_lock(&connector->lock);
+    if(conn->management_close) {
```
gabordozsa, line 2006: Could we make the `conn->management_close` flag atomic? Then we could avoid taking the `connector->lock` here (for every event of every connection).

PR #1611, opened 2024-09-20
 Fixes #1610: echo client wait() blocks until completion 

### LINKED ISSUE #1610: TCP_echo_client tool is flakey and may not actually send data
It appears that calling the wait() method on a TCP_echo_client instance causes the client to immediately exit even if data has not yet transferred.  In other words:

```
client = TcpEchoClient(count=50, size=1000, ...)
client.wait()
```

The call to wait() can return before 50 messages have been sent (including no data).

What's really surprising is that there appears to be a number of work-arounds in the test code to ensure that the client actually sends all the data before calling wait() - including the "delay_close" parameter and the whole EchoClientRunner() class.

Does anyone know why this behavior is desired?  AFAIKT none of the tests need that behavior. And as far as I'm concerned this behavior makes false positive the default for tests using the echo client.

I propose we modify the client so calling wait() actually blocks until either all data is transfered or an error occurs (on which wait() should raise an error).

kgiusti:
Removed dead test and removed extra polling of the echo clients by the tests.
Also modified tests to throw exceptions on error rather than rely on us checking return codes. You can miss a return code, but an exception is harder to ignore.

### DIFF
--- a/tests/TCP_echo_client.py
+++ b/tests/TCP_echo_client.py
@@ -63,22 +63,22 @@ class TcpEchoClient:
                  timeout=TIMEOUT,
                  logger=None,
                  ssl_info=None,
-                 delay_close=False,
-                 wait_connected=True):
+                 delay_close=False):
         """
+        Spawn an echo client in a background thread. On return the echo client
+        is running in the background. Raises an exception if the client fails
+        to connect.
+
         :param host: connect to this host
         :param port: connect to this port
         :param size: size of individual payload chunks in bytes
         :param count: number of payload chunks
-        :param strategy: "1" Send one payload;  # TODO more strategies
-                             Recv one payload
+        :param timeout: return after transferring for timeout seconds
         :param logger: Logger() object
         :param ssl_info: optional SSL information for the connection
         :param delay_close: by default the connection is closed after all data
         has been passed. If delay_close=True, hold the connection open after
         passing all data until the call to 'wait'
-        :param wait_connected: block the caller until the socket has connected (or failed)
-        :return:
         """
         # Start up
         self.sock = None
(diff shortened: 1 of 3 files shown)


tests/TCP_echo_client.py
```diff
@@ -63,7 +63,8 @@ def __init__(self, prefix, host, port, size, count,
                  timeout=TIMEOUT,
                  logger=None,
                  ssl_info=None,
-                 delay_close=False):
+                 delay_close=False,
+                 wait_connected=True):
```
ganeshmurthy, line 67: Just curious as to what is the use case when an echo client will be instantiated with wait_connected=False ?

tests/TCP_echo_client.py
```diff
@@ -286,33 +298,39 @@ def run(self):
...
-            while self.delay_close:
+            # If delaying the close of the connection block until wait is
+            # called
+            if self.delay_close:
                 self.logger.log("%s Delayed closing wait..." % self.prefix)
-                self.is_waiting_to_close = True
-                self._lock.acquire()
-                self._cond.wait(self.timeout)
-                self._lock.release()
+                while self._delay_waiting.wait(timeout=TIMEOUT) is False:
```
ganeshmurthy, line 305: Can this be a if statement instead of a while statement ?
```  
if self._delay_waiting.wait(timeout=TIMEOUT) is False:
    pass
```

ganeshmurthy (approved):

tests/TCP_echo_client.py
kgiusti: That's the original behavior - it never waited for the client to connect. Left that possible in case any of the existing tests broke but none have.  I could simply remove it and make the default wait. In the future if any test was written that expected the client to not connect (like test removing a listener) then we'd need to add it back.

tests/TCP_echo_client.py
kgiusti: I could... but if it throws a timeout is that an error?  It's basically done (successfully) and just waiting for the test to reclaim it.  Removing the loop means we cannot use the standalone client to simply hold TCP connections open in the background while we execute long running tests.

But I suppose the other argument is that it can possibly hang forever if the main test hangs...  dunno.

tests/TCP_echo_client.py
ganeshmurthy: > I could simply remove it and make the default wait. In the future if any test was written that expected the client to not connect (like test removing a listener) then we'd need to add it back.

Yes. that is most excellent. Let's remove the flag and make the default wait.

PR #1618, opened 2024-10-02
Fixes #1613: Added fuzz testing for the http2 decoder. Fixes for two …

### LINKED ISSUE #1613: Add fuzz testing for http2 decoder
Add fuzz testing infrastructure that both the http2 and http1 decoders can use. Also introduce fuzz testing for http2 decoder

ganeshmurthy:
…issues that the fuzz tester found.


tests/fuzz/fuzz_http2_decoder.c
```diff
@@ -0,0 +1,66 @@
...
+void set_alloc_pool_initialized(bool initialized)
+{
+    alloc_pool_initialized = initialized;
+}
+
+bool is_alloc_pool_initialized(void)
+{
+    return alloc_pool_initialized;
+}
+
```
astitcher, line 44: Not a comment on this work: Having to define these functions here - I assume to avoid a linkage error (otherwise why are they defined in a header file) - implies to me that the skupper router library linked might have something a bit wrong with the factoring.

tests/fuzz/fuzz_http2_decoder.c
```diff
@@ -0,0 +1,66 @@
...
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    if (!is_alloc_pool_initialized()) {
+        qd_alloc_initialize();
+        qd_log_initialize();
+        qd_error_initialize();
+        set_alloc_pool_initialized(true);
+    }
```
astitcher, line 51: Why not use `__attribute__((constructor))`? (You use the symmetric `destructor` below)

astitcher (approved):
On the whole lgtm.
Mostly minor quibbles.

tests/fuzz/StandaloneFuzzTargetInit.c
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.
+ *
+ */
+#include <qpid/dispatch/alloc_pool.h>
+#include <qpid/dispatch/log.h>
+#include "libFuzzingEngine.h"
+
+
+
+
```
astitcher, line 27: This file should still have the empty implementation of ```LLVMFuzzerInitialize(argc, argv)```. It is meant to be a (possibly) modified version of the file supplied by either libfuzzer or AFL (I don't remember which one!).

This is just for consistency in building the static library used for regression testing - in theory this file isn't needed any longer for this project.

Ultimately the two file used to build this library should be identical to ones you copied from qpid proton.

astitcher (changes requested):
This looks like it's in the correct form now except for a small niggle with StandaloneFuzzTargetInit,c.

At this point I expect that the files that go to make the static library StandaloneFuzzTargetMain are identical to the ones you got from Qpid Proton. I'm pretty sure you should be able to use these unmodified now.

tests/fuzz/StandaloneFuzzTargetInit.c
ganeshmurthy: Thanks @astitcher I have now pushed up a commit to this PR which makes the two files, `tests/fuzz/StandaloneFuzzTargetInit.c` and `tests/fuzz/StandaloneFuzzTargetMain.c` identical to the files in proton. Please take a look.

astitcher (approved):
lgtm

ganeshmurthy:
> lgtm

Woo hoo !! Thanks for all your help !!

PR #1634, opened 2024-10-14
Added --break-system-packages to pip install

ganeshmurthy:


ganeshmurthy:
This is a test. DO NOT MERGE

.github/workflows/clang-analyzer.yml
```diff
@@ -74,7 +74,7 @@ jobs:
     # https://clang-analyzer.llvm.org/command-line.html
     - name: Create compilation database
       run: |
-        sudo pip3 install codechecker
+        sudo pip3 install --break-system-packages codechecker
```
jiridanek, line 77: Yesterday I decided to use `pipx` for this (https://pipx.pypa.io/stable/)

* https://github.com/opendatahub-io/notebooks/pull/744/files

but this should also do the trick.

I tried looking for what actually changed on github that this is needed (I am familiar with this from macOS where python is installed this way, to disallow pip install, but I'm not sure why it's suddenly on linux gha)

PR #1682, opened 2024-11-25
Fixes #1681: remove unsupported AMQP link detach handling

### LINKED ISSUE #1681: Disallow attempts to detach/reattach an AMQP link
AMQP link detach/reattach is not supported by the skupper-router. The code will interpret a detach as a close, which is incorrect.

The router should treat any attempt to do link detach/reattach as an error and close the link with a proper error message.

kgiusti:


kgiusti:
Verified that the router currently responds to a received detach(close=False) with detatch(close=True) unconditionally. If the detaching client then attempts to re-attach via pn_link_open() - as per [this section of the AMQP 1.0 Spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp339696) then the link simply hangs.

ganeshmurthy (approved):

PR #1685, opened 2024-11-26
Fixes #1365 multicast duplication issue

### LINKED ISSUE #1365: Vanflow: Duplicate message delivery per distinct router path
As discussed in the Skupper Team g chat, I am observing repeated delivery of vanflow messages. This was uncovered by an unrelated skupper collector bug. Thanks for following along @ted-ross!  I suspect this issue is in the router, but I didn't know where to start trying to reproduce it without the control plane.

**To reproduce:**
* Set up Skupper version 1.5.1 running in three namespaces a, b and c and link all three.
```
skupper token create --namespace a --uses 2 token-a.yaml

skupper link create --namespace b token-a.yaml
skupper token create --namespace b token-b.yaml

skupper link create --namespace c token-a.yaml
skupper link create --namespace c token-b.yaml
```

* Deploy this rough draft vanflow logging utility to each namespace to tail the logs. [source](https://github.com/c-kruse/skupper/tree/vanflow-tool/cmd/vanflow-tool).
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: vanflow-logging
  labels:
    app: vanflow-logging
spec:
  replicas: 1
  selector:
    matchLabels:
      app: vanflow-logging
  template:
    metadata:
      labels:
        app: vanflow-logging
    spec:
      containers:
      - image: quay.io/ckruse/vanflow-tool:1.0-alpha3
        imagePullPolicy: Always
        name: main
        securityContext:
          runAsNonRoot: true
        volumeMounts:
        - mountPath: /etc/messaging/
          name: skupper-local-client
        args:
        - '-messaging-config=/etc/messaging/connect.json'
        - '-enable-flow-records=false'
        - 'log'
      volumes:
      - name: skupper-local-client
        secret:
          defaultMode: 420
          secretName: skupper-local-client
```

* Tail the logs of the vanflow logging utilities as you interact with the cluster. Messages originating in the A namespace are logged once from the A namespace but twice from the B and C namespaces.

Karen-Schoener:
Fix qdr_forward_multicast_CT to check valid_origins.

Fixes #1365

### DIFF
--- a/src/router_core/forwarder.c
+++ b/src/router_core/forwarder.c
@@ -631,7 +631,9 @@ int qdr_forward_multicast_CT(qdr_core_t      *core,
 
             // get the inter-router connection associated with path to rnode:
             int conn_bit = (rnode->next_hop) ? rnode->next_hop->conn_mask_bit : rnode->conn_mask_bit;
-            if (conn_bit >= 0 && (!link_exclusion || qd_bitmask_value(link_exclusion, conn_bit) == 0)) {
+
+            if (conn_bit >= 0 && (!link_exclusion || qd_bitmask_value(link_exclusion, conn_bit) == 0) &&
+		(bypass_valid_origins || qd_bitmask_value(rnode->valid_origins, origin))) {
                 qd_bitmask_set_bit(conn_set, conn_bit);
             }
         }


tests/system_tests_multicast.py
```diff
@@ -429,6 +449,245 @@ def test_999_check_for_leaks(self):
...
+
+        cls.INT_A.wait_router_connected('INT.B')
+        cls.INT_A.wait_router_connected('INT.C')
+        cls.INT_B.wait_router_connected('INT.A')
+        cls.INT_B.wait_router_connected('INT.C')
+        cls.INT_C.wait_router_connected('INT.A')
+        cls.INT_C.wait_router_connected('INT.B')
+        cls.EA1.wait_connectors()
+        cls.EB1.wait_connectors()
+        cls.EC1.wait_connectors()
```
ganeshmurthy, line 584: @Karen-Schoener you might also want to use the `is_edge_routers_connected` on each interior router as seen here - 

https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_tcp_adaptor.py#L609

tests/system_tests_multicast.py
```diff
@@ -429,6 +449,245 @@ def test_999_check_for_leaks(self):
...
+
+        d = dict()
+        mgmt = router.management
+        q = mgmt.query(type=ALLOCATOR_TYPE).get_dicts()
+        for name in stats:
+            d[name] = next(a for a in q if a['typeName'] == name)
+
+        return d
+
+    def _check_for_leaks(self):
```
ganeshmurthy, line 657: This `_check_for_leaks(self)` can be removed since there is already code in alloc_pool.c that checks for leaks as seen here - 

https://github.com/skupperproject/skupper-router/blob/main/src/alloc_pool.c#L605

tests/system_tests_multicast.py
Karen-Schoener: Thanks for mentioning.

The new test seems to run more reliably with this additional check.

In an earlier rev of the test (without this check), sometimes 1 of the receivers reported 0 multicast message received...

tests/system_tests_multicast.py
Karen-Schoener: Removed `_check_for_leaks(self)`

tests/system_tests_multicast.py
ganeshmurthy: > The new test seems to run more reliably with this additional check.

Yay !!

Karen-Schoener:
Rebased.

Following @ted-ross's recommendation to remove the system_tests_multicast.py changes from this PR.

I'll open a separate PR with the system_tests_multicast.py changes.

ted-ross (approved):

PR #1691, opened 2024-12-06
Fixes #1690 add mesh multicast test

### LINKED ISSUE #1690: Add a mesh multicast test
Add a mesh multicast test to `system_tests_multicast.py`.

Test goal is to verify the fix for multicast duplicate packet issue: https://github.com/skupperproject/skupper-router/issues/1365.

Karen-Schoener:
Fixes #1690

### DIFF
--- a/tests/system_tests_multicast.py
+++ b/tests/system_tests_multicast.py
@@ -33,7 +33,8 @@ from proton import Link
 from proton import Message
 from proton import Delivery
 from system_test import AsyncTestSender, AsyncTestReceiver, TestCase
-from system_test import Qdrouterd, main_module, TIMEOUT, TestTimeout, unittest
+from system_test import Qdrouterd, main_module, TIMEOUT, TestTimeout, \
+    PollTimeout, unittest
 from system_test import ALLOCATOR_TYPE, ROUTER_ADDRESS_TYPE
 
 
@@ -131,6 +132,12 @@ class MulticastLinearTest(TestCase):
 
         cls.INT_A.wait_router_connected('INT.B')
         cls.INT_B.wait_router_connected('INT.A')
+
+        # Check to make sure that the routers are connected to their
+        # respective edge router.
+        cls.INT_A.is_edge_routers_connected(num_edges=1)
+        cls.INT_B.is_edge_routers_connected(num_edges=1)
+
         cls.EA1.wait_connectors()
         cls.EB1.wait_connectors()
 
@@ -269,6 +276,7 @@ class MulticastLinearTest(TestCase):
         body = " MCAST PRESETTLED "
         test = MulticastPresettled(self.config, 10, body, SendPresettled())
         test.run()
+        self.assertIsNone(test.error)
 
     def test_51_presettled_mixed_large_msg(self):
         # Same as above, but large message bodies (mixed sender settle mode)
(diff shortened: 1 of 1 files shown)


tests/system_tests_multicast.py
```diff
@@ -429,6 +438,137 @@ def test_999_check_for_leaks(self):
...
+    def setUpClass(cls):
+        """Start a router"""
+        super(MulticastMeshTest, cls).setUpClass()
+
+        def router(name, mode, extra):
+            config = [
+                ('router', {'mode': mode,
+                            'id': name,
+                            'workerThreads': W_THREADS}),
+                ('listener', {'role': 'normal',
```
ganeshmurthy, line 456: This listener is not necessary. It is already being added here - https://github.com/skupperproject/skupper-router/blob/main/tests/system_test.py#L701

tests/system_tests_multicast.py
```diff
@@ -429,6 +438,137 @@ def test_999_check_for_leaks(self):
...
+             'senders':     [],
+             'receivers':   ['R-INT_C-1', 'R-INT_C-2'],
+             'subscribers': 2,
+             'remotes':     2,
+             }
+        ]
+
+    #
+    # run any negative tests first so that if we screw up the internal
+    # state of the brokers the positive tests will likely fail
```
ganeshmurthy, line 555: Should this comment be removed ? What are the negative tests ?

Karen-Schoener: Removed comment.

tests/system_tests_multicast.py
```diff
@@ -429,6 +438,128 @@ def test_999_check_for_leaks(self):
...
+                              'maxFrameSize': MAX_FRAME,
+                              'linkCapacity': LINK_CAPACITY}),
+                ('address', {'prefix': 'multicast', 'distribution': 'multicast'}),
+            ]
+
+            if extra:
+                config.extend(extra)
+            config = Qdrouterd.Config(config)
+            cls.routers.append(cls.tester.qdrouterd(name, config, wait=True))
+            router_instance = cls.routers[-1]
```
ganeshmurthy, line 467: This `router_instance` is unused and hence can be removed. The test passed for me without it !

tests/system_tests_multicast.py
Karen-Schoener: Discussed this issue a bit further with @ganeshmurthy.

Updated with Ganesh's advice: keep the 'listener' line here, since the test creates routers with role=inter-router; update test to not specify: maxFrameSize, linkCapacity.

tests/system_tests_multicast.py
Karen-Schoener: Removed `router_instance`.

tests/system_tests_multicast.py
```diff
@@ -506,26 +634,30 @@ def __init__(self, config, count, body, topic=None, **handler_kwargs):
...
         self.n_outcomes = {}
 
         self.error = None
         self.timers = []
         self.reactor = None
+        self.done_timer = None
 
     def done(self):
         # stop the reactor and clean up the test
         for t in self.timers:
```
ganeshmurthy, line 649: Was wondering why there needs to be a list of timers in `self.timers`
Can it just be self.timer ? I noticed the the `len(self.timers)` is always one

tests/system_tests_multicast.py
```diff
@@ -673,7 +820,13 @@ def check_if_done(self):
         # sender
         to_rcv = self.n_senders * self.msg_count * self.n_receivers
         if to_rcv == self.n_received and not self.unsettled_deliveries:
-            self.done()
+            if self.done_wait_interval:
+                # start a timer to wait before stopping test in order to
+                # wait for any any duplicate messages to be received
+                self.done_timer = threading.Timer(self.done_wait_interval, self.done)
```
ganeshmurthy, line 826: You could use a `poll_timer` like here - https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_delivery_counts.py#L954 - instead of `threading.Timer`

tests/system_tests_multicast.py
Karen-Schoener: I had on open question about this myself.

I notice that `MulticastBase.on_start()` has code to append to the timers list.
https://github.com/skupperproject/skupper-router/blob/main/tests/system_tests_multicast.py#L541

I was not sure if the result of `self.reactor.schedule(TIMEOUT, TestTimeout(self))` can/should be mixed with a threading.Timer.

If you prefer that I remove self.done_timer and append the 'done' timer to self.timers, just let me know.  I'll be curious to hear your preference...

tests/system_tests_multicast.py
Karen-Schoener: Removed done_timer.
Removed `import threading`.
Added poll_timer.

tests/system_tests_multicast.py
Karen-Schoener: Removed done_timer.
Removed `import threading`.
Added poll_timer.

ganeshmurthy (approved):

PR #1695, opened 2024-12-11
Fixes #1694: Populated the SOURCE_HOST vanflow attribute in the http2…

### LINKED ISSUE #1694: Capture the source port in the http2 adaptor
The source port can be found in the `x-forwarded-for` header. Capture the value of this header in the vanflow attribute `VFLOW_ATTRIBUTE_SOURCE_HOST`

ganeshmurthy:
… adaptor

### DIFF
--- a/src/observers/http2/http2_observer.c
+++ b/src/observers/http2/http2_observer.c
@@ -34,7 +34,6 @@ ALLOC_DEFINE(qd_http2_stream_info_t);
  */
 static qd_error_t insert_stream_info_into_hashtable(qdpo_transport_handle_t *transport_handle, qd_http2_stream_info_t *stream_info, uint32_t stream_id)
 {
-    assert(stream_id != 0);
     qd_hash_t  *stream_id_hash = transport_handle->http2.stream_id_hash;
     // Convert stream_id to string
     char stream_id_str[11];
@@ -47,7 +46,6 @@ static qd_error_t insert_stream_info_into_hashtable(qdpo_transport_handle_t *tra
  */
 static qd_error_t get_stream_info_from_hashtable(qdpo_transport_handle_t *transport_handle, qd_http2_stream_info_t **stream_info, uint32_t stream_id)
 {
-    assert(stream_id != 0);
     qd_hash_t  *stream_id_hash = transport_handle->http2.stream_id_hash;
     // Convert stream_id to string
     char stream_id_str[11];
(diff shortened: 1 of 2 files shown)


src/observers/http2/http2_observer.c
```diff
@@ -194,6 +195,19 @@ static int on_data_recv_callback(qd_http2_decoder_connection_t *conn_state,
     return 0;
 }
 
+static void set_stream_vflow_attribute(qdpo_transport_handle_t *transport_handle, uint32_t stream_id, vflow_attribute_t attribute_type, const char *string_value)
+{
+    qd_http2_stream_info_t *stream_info = 0;
+    qd_error_t error = get_stream_info_from_hashtable(transport_handle, &stream_info, stream_id);
+    assert(stream_info != 0);
+    if(error == QD_ERROR_NOT_FOUND) {
```
kgiusti, line 203: Question that has nothing to do with this patch: should failure to find the given stream cause the observer to disable itself?  Looks like we assert if this happens in a debug build so why continue decoding in a production build?

kgiusti (approved):

src/observers/http2/http2_observer.c
ted-ross: Why is there an assert statement here?  Does null stream_info indicate a software failure or could it be some kind of protocol anomaly.  If it's the latter, crashing (even in a debug build) is unacceptable.

ganeshmurthy: > should failure to find the given stream cause the observer to disable itself?

I am thinking aloud here. Say there are 100 streams in a http2 request and the first stream id is not found, would it be ok to continue to report on the other 99 streams instead of calling `qdpo_http2_final()` ?

```suggestion
    if (!stream_info)
        return;
    if(error == QD_ERROR_NOT_FOUND) {
```

ganeshmurthy: > kind of protocol anomaly

I would like to keep the assert in the debug build just to so we can get a chance to see why the issue is happening. But in production, we should just return if the stream_info is not found.

ted-ross: Also, 'if' is not a function call.  Please put a space after it.

ted-ross: > > kind of protocol anomaly
> 
> I would like to keep the assert in the debug build just to so we can get a chance to see why the issue is happening. But in production, we should just return if the stream_info is not found.

If you're using the assert to look for a bug that you're seeing, convert this to a draft PR and remove the assert after you solve the problem.

I would expect that fuzz testing might cause that assert to fire.  I don't think we want the router crashing in test due to known protocol anomalies.

ted-ross (approved):

PR #1699, opened 2024-12-13
Fixes #1698: prevent use-after-free of qd_link_t on session close

### LINKED ISSUE #1698: Teardown of an AMQP session with active links and deliveries causes a router crash
```
=================================================================
==258021==ERROR: AddressSanitizer: use-after-poison on address 0x5130000ed4a0 at pc 0x000000478562 bp 0x7ff47adabe60 sp 0x7ff47adabe58
WRITE of size 8 at 0x5130000ed4a0 thread T5
    #0 0x478561 in qd_container_handle_event /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/container.c:564
    #1 0x49a426 in qd_connection_handle_event /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/qd_connection.c:828
    #2 0x6a8cf2 in proactor_thread /home/kgiusti/tmp/skupper-router/src/server.c:172
    #3 0x551579 in _thread_init /home/kgiusti/tmp/skupper-router/src/posix/threading.c:207
    #4 0x7ff49185df95 in asan_thread_start(void*) (/lib64/libasan.so.8+0x5df95) (BuildId: a4ad7eb954b390cf00f07fa10952988a41d9fc7a)
    #5 0x7ff4908a66d6 in start_thread (/lib64/libc.so.6+0x976d6) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)
    #6 0x7ff49092a60b in __clone3 (/lib64/libc.so.6+0x11b60b) (BuildId: 77c77fee058b19c6f001cf2cb0371ce3b8341211)

0x5130000ed4a0 is located 160 bytes inside of 320-byte region [0x5130000ed400,0x5130000ed540)
allocated by thread T5 here:
    #0 0x7ff4918f6f8d in posix_memalign (/lib64/libasan.so.8+0xf6f8d) (BuildId: a4ad7eb954b390cf00f07fa10952988a41d9fc7a)
    #1 0x4c57fc in qd_alloc /home/kgiusti/tmp/skupper-router/src/alloc_pool.c:343
    #2 0x471f8a in new_qd_link_t /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/container.c:63
    #3 0x472e4f in setup_incoming_link /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/container.c:182
    #4 0x47950d in qd_container_handle_event /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/container.c:604
    #5 0x49a426 in qd_connection_handle_event /home/kgiusti/tmp/skupper-router/src/adaptors/amqp/qd_connection.c:828
    #6 0x6a8cf2 in proactor_thread /home/kgiusti/tmp/skupper-router/src/server.c:172
    #7 0x551579 in _thread_init /home/kgiusti/tmp/skupper-router/src/posix/threading.c:207
    #8 0x7ff49185df95 in asan_thread_start(void*) (/lib64/libasan.so.8+0x5df95) (BuildId: a4ad7eb954b390cf00f07fa10952988a41d9fc7a)
```

kgiusti:
Previously the detach handler might free the qd_link_t but it was not consistent. This patch moves the qd_link_t cleanup to the container.

### DIFF
--- a/src/adaptors/amqp/container.c
+++ b/src/adaptors/amqp/container.c
@@ -54,10 +54,11 @@ struct qd_link_t {
     qd_alloc_safe_ptr_t         incoming_msg;  // DISPATCH-1690: for cleanup
     pn_snd_settle_mode_t        remote_snd_settle_mode;
     qd_link_ref_list_t          ref_list;
-    bool                        q2_limit_unbounded;
-    bool                        q3_blocked;
     DEQ_LINKS_N(Q3, qd_link_t); ///< Q3 blocked links
     uint64_t                    link_id;
+    bool                        q2_limit_unbounded;
+    bool                        q3_blocked;
+    bool                        policy_counted;  // has this been counted by policy?
 };
 
 ALLOC_DEFINE_SAFE(qd_link_t);
@@ -83,6 +84,9 @@ struct qd_session_t {
     // See qd_session_get_outgoing_capacity() for details.
     size_t outgoing_bytes_high_threshold;
     size_t outgoing_bytes_low_threshold;
+
+    // Has this session been counted by policy? Only remotely initiated sessions can be counted by policy
+    bool policy_counted;
 };
 
 
@@ -148,7 +152,7 @@ static inline qd_session_t *qd_session_from_pn(pn_session_t *pn_ssn)
     return (qd_session_t *)pn_session_get_context(pn_ssn);
 }
 
-static void setup_outgoing_link(qd_container_t *container, pn_link_t *pn_link)
+static qd_link_t *setup_outgoing_link(qd_container_t *container, pn_link_t *pn_link)
 {
     qd_link_t *link = new_qd_link_t();
     if (!link) {
@@ -156,7 +160,7 @@ static void setup_outgoing_link(qd_container_t *container, pn_link_t *pn_link)
         pn_condition_set_name(cond, QD_AMQP_COND_INTERNAL_ERROR);
         pn_condition_set_description(cond, "Insufficient memory");
         pn_link_close(pn_link);
-        return;
+        return 0;
     }
 
     ZERO(link);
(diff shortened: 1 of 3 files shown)


src/adaptors/amqp/container.c
```diff
@@ -645,6 +627,7 @@ void qd_container_handle_event(qd_container_t *container, pn_event_t *event,
         pn_link = pn_event_link(event);
         if (pn_link_state(pn_link) == (PN_LOCAL_CLOSED | PN_REMOTE_CLOSED)) {
             add_link_to_free_list(&qd_conn->free_link_list, pn_link);
+            qd_link_free((qd_link_t *) pn_link_get_context(pn_link));
```
ganeshmurthy, line 630: Instead of calling `qd_link_free()` once in the handler to `PN_LINK_REMOTE_CLOSE` and once in `PN_LINK_LOCAL_CLOSE`, can we call it just once in the `qd_conn_event_batch_complete()` function where we loop thru the `qd_conn->free_link_list` ?

In fact, I was wondering if we could have the `qd_conn_event_batch_complete()` function as the only place we call 'qd_link_free()' and 'pn_link_free()'

src/adaptors/amqp/container.c
```diff
@@ -529,52 +508,43 @@ void qd_container_handle_event(qd_container_t *container, pn_event_t *event,
...
-                if (qd_conn->policy_settings) {
-                    qd_conn->n_sessions--;
-                }
-
                 pn_session_close(ssn);
             }
             else if (pn_session_state(ssn) == (PN_LOCAL_CLOSED | PN_REMOTE_CLOSED)) {
                 qd_session_t *qd_ssn = qd_session_from_pn(ssn);
                 if (qd_ssn) {
+                    if (qd_ssn->policy_counted) {
```
ganeshmurthy, line 543: Can we please move this `policy_counted` related code to a new commit so this commit only contains the code that fixes the router crash ?

kgiusti: I had to add that hacky policy flag because otherwise this patch would break the policy code.  The problem is that not all sessions or links are counted by policy: only those that are 1) remotely initiated and 2) policy is in use.

That means the code cannot _unconditionally_ decrement the relevant policy counters since the counters may not have been incremented for that particular session/link (for reasons 1 and 2 above). So I have to add that ugly "hey I'm counted by policy so remember to decrement when I'm done" hack.

Unconditionally incrementing/decrementing the counters will break the policy logic: there's even asserts in the (existing) code to guard against underflow: https://github.com/skupperproject/skupper-router/blob/2249924bc5304593a71ef36213def662177665cd/src/adaptors/amqp/container.c#L551

ganeshmurthy: > That means the code cannot _unconditionally_ decrement the relevant policy counters since the counters may not have been incremented for that particular session/link

Makes sense to me.

ganeshmurthy (approved):

PR #1701, opened 2024-12-17
Fixes #1700: refactor the AMQP link lifecycle

### LINKED ISSUE #1700: Refactor AMQP link detach/close handling to clarify link lifecycle
The code that handles the creation and deletion of AMQP links is implemented to support a deprecated feature called AMQP link attach routing. Since this feature has long been removed from the router it would be a good idea to see if we can simplify the AMQP link lifecycle.

Currently there is a lot of logic involved in tracking the attach/detach performative handshakes that can be simplified now that link attach routing is no longer supported.

Specifically the point where the link lifecycle terminates is a little ambiguous especially where the qd_link_t/qdr_link_t state is shared between the core and I/O threads.

kgiusti:
This change removes some of the old link attach routing logic and attempts to clean up the link API.

The logic that used to track the exchange of Attach/Detach performatives has been simplified. The various counters and booleans maintained by the qdr_link_t structure for tracking this exchange has been reduced to a mask/flag implementation similar to Protons implementation of endpoint state.

This patch refactors the link detach adaptor API to be more like the existing AMQP connection API: there is now an explict API call to release the link instance at the end of its lifecycle.

The adaptor API is modified by separating the AMQP detach handling from the release of the link instance. The old qdr_link_detach() adaptor function has been refactored into two functions: qdr_link_detach_received() and qdr_link_close().

The qdr_link_detach_received() call is made by the AMQP adaptor when a Detach Peformative has been received by the peer. It is only used by the AMQP adaptor.

The new qdr_link_closed() API call is made by all adaptors when the link instance is destroyed. This is similar to the existing qdr_connection_closed() call but for links. It is used by all adaptors to indicate to the core that the link is no longer in use and can be cleaned up. In the case of the AMQP adaptor this call will be made after the link detach handshake has completed.

Test coverage by the test-sender AMQP client has been increased by adding a clean connection close function.

### DIFF
--- a/include/qpid/dispatch/protocol_adaptor.h
+++ b/include/qpid/dispatch/protocol_adaptor.h
@@ -643,12 +643,6 @@ qd_iterator_t *qdr_terminus_dnp_address(qdr_terminus_t *term);
  ******************************************************************************
  */
 
-typedef enum {
-    QD_DETACHED,  // Protocol detach
-    QD_CLOSED,    // Protocol close
-    QD_LOST       // Connection or session closed
-} qd_detach_type_t;
-
 /**
  * qdr_link_set_context
  *
@@ -800,15 +794,32 @@ qdr_link_t *qdr_link_first_attach(qdr_connection_t *conn,
 void qdr_link_second_attach(qdr_link_t *link, qdr_terminus_t *source, qdr_terminus_t *target);
 
 /**
- * qdr_link_detach
+ * qdr_link_detach_received
  *
- * This function is invoked when a link detach arrives.
+ * This function is invoked when a link detach performative arrives from the remote peer. This may the first detach
+ * (peer-initiated link detach) or in response to a detach sent by the router (second detach).
  *
  * @param link The link pointer returned by qdr_link_first_attach or in a FIRST_ATTACH event.
- * @param dt The type of detach that occurred.
  * @param error The link error from the detach frame or 0 if none.
  */
-void qdr_link_detach(qdr_link_t *link, qd_detach_type_t dt, qdr_error_t *error);
+void qdr_link_detach_received(qdr_link_t *link, qdr_error_t *error);
+
+
+/**
+ * qdr_link_closed
+ *
+ * This function is invoked by the adaptor when the link has fully closed. This will be the last call made by the
+ * adaptor for this link. This may be called as a result of a successful detach handshake or due to link loss. This will
+ * also be called during adaptor shutdown on any outstanding links.
+ *
+ * The core may free the qdr_link_t by this call. The adaptor MUST NOT reference the qdr_link_t on return from this
+ * call.
+ *
+ * @param link The link pointer returned by qdr_link_first_attach or in a FIRST_ATTACH event.
+ * @param forced True if the link was closed due to failure or shutdown. False if closed by clean detach handshake.
+ */
+void qdr_link_closed(qdr_link_t *link, bool forced);
+
 
 /**
  * qdr_link_deliver
--- a/include/qpid/dispatch/protocol_adaptor.h
+++ b/include/qpid/dispatch/protocol_adaptor.h
@@ -386,15 +386,18 @@ qdr_connection_t *qdr_connection_opened(qdr_core_t                    *core,
                                         void                          *bind_token);
 
 /**
- * qdr_connection_closed
+ * qdr_connection_notify_closed
  *
- * This function must be called when a connection is closed, either cleanly by protocol
- * or uncleanly by lost connectivity.  Once this function is called, the caller must never
- * again refer to or use the connection pointer.
+ * This function is invoked by the adaptor to notify the core that the given connection has been closed. This must be
+ * called when a connection is closed, either cleanly by protocol or uncleanly by lost connectivity.
+ *
+ * This must be the last core API call made by the adaptor for this connection. The core thread will free the
+ * qdr_connection_t as a result of this call therefore the adaptor MUST NOT reference the qdr_connection_t on return
+ * from this call.
  *
  * @param conn The pointer returned by qdr_connection_opened
  */
-void qdr_connection_closed(qdr_connection_t *conn);
+void qdr_connection_notify_closed(qdr_connection_t *conn);
 
 /**
  * qdr_connection_set_tracing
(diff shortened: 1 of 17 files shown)


src/router_core/connections.c
```diff
@@ -2433,24 +2405,78 @@ static void qdr_link_inbound_detach_CT(qdr_core_t *core, qdr_action_t *action, b
...
+            qd_log(LOG_ROUTER_CORE, QD_LOG_DEBUG,
+                   "[C%"PRIu64"][L%"PRIu64"] qdr_link_closed_CT(forced=%s) handle %s detach",
+                   link->conn->identity, link->identity, forced_close ? "YES" : "NO",
+                       (link->state & QDR_LINK_STATE_DETACH_SENT) == 0 ? "first" : "second");
+
+            qdr_link_process_detach(core, link, 0);
+        }
+    }
+
+    qdr_link_cleanup_protected_CT(core, link->conn, link,
```
ganeshmurthy, line 2477: Can the `qdr_link_cleanup_protected_CT()` function be removed and its contents copied into `qdr_link_closed_CT()`

ganeshmurthy:
Can the `qdr_link_cleanup_CT()` function be called `qdr_link_free_CT()`
IMO, it will make its intent very clear.

include/qpid/dispatch/protocol_adaptor.h
```diff
@@ -810,15 +804,32 @@ qdr_link_t *qdr_link_first_attach(qdr_connection_t *conn,
...
- * This function is invoked when a link detach arrives.
+ * This function is invoked when a link detach performative arrives from the remote peer. This may the first detach
+ * (peer-initiated link detach) or in response to a detach sent by the router (second detach).
  *
  * @param link The link pointer returned by qdr_link_first_attach or in a FIRST_ATTACH event.
- * @param dt The type of detach that occurred.
  * @param error The link error from the detach frame or 0 if none.
  */
-void qdr_link_detach(qdr_link_t *link, qd_detach_type_t dt, qdr_error_t *error);
+void qdr_link_detach_received(qdr_link_t *link, qdr_error_t *error);
```
ganeshmurthy, line 815: I like renaming the function `qdr_link_detach()` to `qdr_link_detach_received()`. The new function name makes it clear that a detach (first or second) has been received from the remote peer.

include/qpid/dispatch/protocol_adaptor.h
```diff
@@ -810,15 +804,32 @@ qdr_link_t *qdr_link_first_attach(qdr_connection_t *conn,
...
+ * adaptor for this link. This may be called as a result of a successful detach handshake or due to link loss. This will
+ * also be called during adaptor shutdown on any outstanding links.
+ *
+ * The core may free the qdr_link_t by this call. The adaptor MUST NOT reference the qdr_link_t on return from this
+ * call.
+ *
+ * @param link The link pointer returned by qdr_link_first_attach or in a FIRST_ATTACH event.
+ * @param forced True if the link was closed due to failure or shutdown. False if closed by clean detach handshake.
+ */
+void qdr_link_closed(qdr_link_t *link, bool forced);
```
ganeshmurthy, line 831: I looked for the usage of this function in the tcp adaptor, say in `close_connection_XSIDE_IO()` - 

```
    if (!!conn->outbound_link) {
        qdr_link_closed(conn->outbound_link, true);
    }
```

In the above usage, we are calling the `qdr_link_closed()` function to close a link. Should the function be called `qdr_link_close()` or `qd_close_link()` instead ? `closed` seems to mean that the link has already been closed

src/adaptors/amqp/amqp_adaptor.c
```diff
@@ -1257,29 +1238,59 @@ static int AMQP_link_detach_handler(qd_router_t *router, qd_link_t *link, qd_det
...
 }
 
+
+/**
+ * Link closed handler
+ *
+ * This is the last callback for the given link - the link will be freed on return from this call! Forced is true if the
+ * link has not properly closed (detach handshake completed).
+*/
+static void AMQP_link_closed_handler(qd_router_t *router, qd_link_t *qd_link, bool forced)
```
ganeshmurthy, line 1264: I was wondering if the `AMQP_` prefix should be used only for functions that are called directly as a response to proton AMQP events. This function can be called `qd_link_close()` ?

ganeshmurthy (approved):

src/router_core/connections.c
kgiusti: Good eye - not quite yet but soon.  My next patch will (hopefully) rework that "protected" stuff to remove the whole "Is the core finished? Is the I/O finished? Can it be freed?" flags handshake.  Hopefully that qdr_link_cleanup_protected_CT() goes away and we simply have a qdr_link_free_CT() call (like you recommend below).

Stay tuned.

include/qpid/dispatch/protocol_adaptor.h
kgiusti: Thanks!

include/qpid/dispatch/protocol_adaptor.h
kgiusti: > closed seems to mean that the link has already been closed

Indeed that's the idea: the adaptor has detected the link is closed and needs to notify the core (hence **qdr**_link_closed rather than **qd**_link_closed: it's a Core API call not an adaptor call) that the link no longer exists.

I chose **qdr_link_closed** because the function performs essentially the same thing as the exising core function **qdr_connection_closed** except the new one deals with qdr_link_t's not qdr_connection_t's).

You'll see that both adaptors call qdr_connection_close() right before they clean up their connection state objects (qd_tcp_connection_t or qd_connection_t - depending on the adaptor).  The TCP adaptor used to call qdr_link_detach() even though there's not such thing as a detach performative for TCP!  For TCP we basically simulate a network connection drop since there's no AMQP close handshake going on.

See [here](https://github.com/kgiusti/skupper-router/blob/ecf7305f3028c19288f739015dc333957d0cbb49/include/qpid/dispatch/protocol_adaptor.h#L389)  and [here](https://github.com/kgiusti/skupper-router/blob/ecf7305f3028c19288f739015dc333957d0cbb49/include/qpid/dispatch/protocol_adaptor.h#L819)

What do you think?

src/adaptors/amqp/amqp_adaptor.c
kgiusti: Could be... but I'm not sure if other devs would be ok with that new naming concept.  @ssorj @ted-ross and others - opinions?

PR #1704, opened 2024-12-19
Fixes #1694: Additional fix. Capture the first left-most x-forwarded-…

### LINKED ISSUE #1694: Capture the source port in the http2 adaptor

ganeshmurthy:
…for header value

### DIFF
--- a/src/observers/http2/http2_observer.c
+++ b/src/observers/http2/http2_observer.c
@@ -191,18 +191,6 @@ static int on_data_recv_callback(qd_http2_decoder_connection_t *conn_state,
     return 0;
 }
 
-static void set_stream_vflow_attribute(qdpo_transport_handle_t *transport_handle, uint32_t stream_id, vflow_attribute_t attribute_type, const char *string_value)
-{
-    qd_http2_stream_info_t *stream_info = 0;
-    qd_error_t error = get_stream_info_from_hashtable(transport_handle, &stream_info, stream_id);
-    if (error == QD_ERROR_NOT_FOUND) {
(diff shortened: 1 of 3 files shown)


src/observers/http2/http2_observer.c
```diff
@@ -246,7 +240,33 @@ static int on_header_recv_callback(qd_http2_decoder_connection_t *conn_state,
...
+                // X-Forwarded-For: 2001:db8:85a3:8d3:1319:8a2e:370:7348, 197.1.773.201
+                // X-Forwarded-For: 195.0.223.001
+                // X-Forwarded-For: 203.0.113.195, 2007:db5:85a3:8d3:1319:8a2e:370:3221
+                //
+                // We will obtain the value of the first X-Forwarded-For (2001:db8:85a3:8d3:1319:8a2e:370:7348, 197.1.773.201) and ignore the
+                // other two X-Forwarded-For headers.
+                // The first X-Forwarded-For header is comma separated list, we will obtain the leftmost (the first) value (2001:db8:85a3:8d3:1319:8a2e:370:7348) in the list
+
+                // const uint8_t *value passed into this function is guaranteed to be NULL-terminated
+                char value_copy[valuelen+1];
```
kgiusti, line 262: Is there a limit to valuelen?  Can it be large enough to overflow the stack?

ganeshmurthy: > so hardcoding a array size of 128 would be more than enough I would think.

Agreed. I made the change. Thanks @kgiusti

PR #1716, opened 2025-01-16
Fixes #1715: Removed libyaml libusbx and systemd-libs from the skuppe…

### LINKED ISSUE #1715: Remove libyaml libusbx systemd-libs from the skupper-router container
systemd-libs caused  CVE-2021-3997 (libusbx depends on systemd-libs, so we removed that as well)
libyaml caused CVE-2024-35325

ganeshmurthy:
…r-router container


kgiusti (approved):
LGTM

PR #1717, opened 2025-01-17
Fixes #1650: Added the version field to the site entity

### LINKED ISSUE #1650: [VanFlow]: Include a "version" configuration field for site records
The control plane and console presently rely on a version string set on the SITE vanflow records to communicate skupper controller versions of connected sites (distinct from router version.)

Could we add a configuration option so that the "controller" (in this case the non-kube bootstrap process) could expose its version?

This was likely missed on the first go-around as the vanflow spec was missing this attribute: https://github.com/skupperproject/vanflow-specification/commit/260504a7b547836ce3a0d9bbc7d3ffff4120c9d9

ganeshmurthy:


python/skupper_router/management/skrouter.json
```diff
@@ -650,6 +650,12 @@
                     "required": false,
                     "create": true
                 },
+                "version": {
+                    "description":"Build version of skupper",
```
ted-ross, line 654: Maybe make this a little more descriptive:  "Build version of the running control-plane software"

ted-ross (approved):
Approved with one suggested improvement.

python/skupper_router/management/skrouter.json
ganeshmurthy: > Maybe make this a little more descriptive: "Build version of the running control-plane software"

Done !

Issue #1718, opened 2025-01-21
Install rsyslog, util-linux packages as part of skupper-router image


ukumar009:
Hi Skupper team,

From skupper-router 2.7.3 version we noticed images is not working as expected. We install rsyslog package on our end for logging purposes. but i noticed from 2.7.3 microdnf is removed due to which we are not able to install additional packages.

Could you please consider installing the following packages as part of skupper-router?
`rsyslog, util-linux`

Thanks,

kgiusti:
Hi Uday. I think a better approach to solving this would be to simply add microdnf to the Containerfile.  That way you and others can easily install stuff you need on the fly.

I'll follow up in the PR you submitted, thanks.

ukumar009:
sure @kgiusti  thank you

ukumar009:
Hi @kgiusti i tried [christian kruise suggestion](https://github.com/skupperproject/skupper-router/pull/1719#issuecomment-2608116252) and i am able to install rsyslog and other packages into skupper-router image. So, I am happy to close this issue.

PR #1719, opened 2025-01-21
Add rsyslog & util-linux package Containerfile

ukumar009:
Added `rsyslog` and `util-linux` package for logging support.

Fix for: https://github.com/skupperproject/skupper-router/issues/1718

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -71,7 +71,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
 # Remove gnutls, libarchive and everything that depends on it. 
 # https://github.com/skupperproject/skupper-router/issues/1477
 # https://github.com/skupperproject/skupper-router/issues/1639
-RUN microdnf -y remove gnutls glib2 gobject-introspection libpeas microdnf gnupg2 gpgme libdnf json-glib libmodulemd librepo librhsm libsolv rpm rpm-libs libarchive libyaml libusbx systemd-libs
+RUN microdnf -y remove gnutls glib2 gobject-introspection libpeas microdnf gnupg2 gpgme libdnf json-glib libmodulemd librepo librhsm libsolv rpm rpm-libs libarchive libyaml libusbx
 
 RUN useradd --uid 10000 runner
 USER 10000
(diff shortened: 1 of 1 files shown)


kgiusti:
Hi Uday - can you try changing [this line](https://github.com/skupperproject/skupper-router/blob/2a0aaba75ccf8807843719d84d3d6f0697b2c5bc/Containerfile#L74) by removing where microdnf is removed instead?  Keeping microdnf in the container would be very useful, thanks.

ganeshmurthy:
We removed `microdnf` as part of removing `gnutls` (see e8ae8f172c0b4f97cbf3574274d76ccb4e78d478) 
`gnutls` was causing lots of CVEs, so we removed it and also ended up removing 'microdnf' because it depended on the former. 
Removing 'microdnf' is actually good security wise since no one can install additional packages to mess with the container. 

I don't know much about `rsyslog` but can the logs be copied out of the container and `rsyslog` be run on those logs ?

c-kruse:
An alternative would be to build a container image with these packages installed (rather than including a package manager.)

Maybe something like this?
```
# Containerfile
FROM registry.access.redhat.com/ubi9/ubi-minimal:latest as overlay
RUN mkdir overlay && \
  microdnf install \
        --noplugins \
        --installroot /overlay/ \
        --releasever 9.5 \
        --config=/etc/dnf/dnf.conf \
        --setopt=reposdir=/etc/yum.repos.d \
        --setopt=cachedir=/var/cache/microdnf \
        --setopt=varsdir=/etc/dnf \
        --setopt=install_weak_deps=0  \
        --nodocs \
        -y \
        rsyslog util-linux

FROM quay.io/skupper/skupper-router:2.7.3-1
COPY --from=overlay /overlay /
```

It seems to me that stripped down container images without a package manager are very popular, and there should be pretty well documented means to get packages installed in those environments. Not a rhel ecosystem pro, but this approach appears to work well for me. Maybe look at docs on how to use the ubi9 micro container images (does not have a package manager by default)?

ukumar009:
Thank you @c-kruse  I will try it out and let you know.

ukumar009:
Hi @c-kruse  I tried your approach and it is working as per our expectation. So, I am happy to close the issue and PR.

PR #1724, opened 2025-01-28
Fixes #1723: For inter-router connectors, force set the dataConnectio…

### LINKED ISSUE #1723: Force dataConnectionCount in the router entity to 1 if set to zero
With the new Certificate Rotation feature (https://github.com/skupperproject/skupper-router/issues/1722)
the dataConnectionCount attribute of the router entity cannot be set to 0 (zero). If accidentally set to zero, force set it to 1.

ganeshmurthy:
…nCount to 1 if it is set to 0

### DIFF
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -618,7 +618,7 @@
                     "required": false
                 },
                 "dataConnectionCount": {
-                    "description": "The number of parallel data connections to carry streaming data between routers. Applies only to interior routers",
+                    "description": "The number of parallel data connections to carry streaming data between routers. Applies only to interior routers. Defaults to 'auto', don't set to 0 (zero), if set to zero, will be forced to 1",
                     "type": "string",
                     "required": false,
                     "default": "auto",
--- a/src/adaptors/amqp/server_config.c
+++ b/src/adaptors/amqp/server_config.c
@@ -152,12 +152,14 @@ qd_error_t qd_server_config_load(qd_dispatch_t *qd, qd_server_config_t *config, 
     if (strcmp(config->role, "inter-router") == 0) {
         // For inter-router connections only, the dataConnectionCount defaults to "auto",
         // which means it will be determined as a function of the number of worker threads.
-        config->data_connection_count = strdup(qd->data_connection_count);
-        // If the user has *not* explicitly set the value "0",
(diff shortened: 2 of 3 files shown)


python/skupper_router/management/skrouter.json
```diff
@@ -618,7 +618,7 @@
                     "required": false
                 },
                 "dataConnectionCount": {
-                    "description": "The number of parallel data connections to carry streaming data between routers. Applies only to interior routers",
+                    "description": "The number of parallel data connections to carry streaming data between routers. Applies only to interior routers. Defaults to 'auto', don't set to 0 (zero), if set to zero, will be forced to 1",
```
kgiusti, line 621: Not something you changed but maybe this description should be more specific WRT the connection role.  Like "Applies only to connectors with a role of "inter-router"???  Or something like that...

kgiusti:
Stepping back a bit - I don't like how the data connection count is stored as a string then constantly re-computed every time a new connector is created.  I'm not being critical about your patch - reviewing your patch leads me to believe the original implementation of the data connection count leaves room for improvement.

For example, the data connection count is stored in the router configuration record. It's value is initially read in dispatch.c::qd_dispatch_configure_router.  That code simply stores it in a string. But then the actual computation of the data connection count integer is buried in the connection manager.  That computation must be repeated every time a new connector is created.

This seems complicated and the logic is needlessly spread across files.

Wouldn't it be simpler to:

1) Compute the actual value of data connection count in the qd_dispatch_configure_router routine where the configuration is read?
1a) By "compute" I mean translate the string "auto" to the actual integer value based on the number of worker threads. The number of worker threads is also read from the same configuration record as the data connection count so it's immediately accessible.
2) store the data connection count as an integer, not a string.  We need an integer, not a string.

Then you can remove all the extra logic in connection_manager.c::qd_dispatch_configure_connector() for computing the integer value of data connection count and simply copy it from the qd parameter.

Then those access functions that go through the core (via the management thread - shudder) can be removed.  Specifically connection_manager.c::auto_calc_connection_count and also qdr_core_get_worker_thread_count which is totally unnecessary since the thread count is available in the qd parameter.

I think that would greatly simplify the implementation of the data_connection_count configuration - what do you think?

ganeshmurthy:
> what do you think?

I will embark on this mission and will keep you posted.

PR #1726, opened 2025-01-28
Fixes #1725: Removed libxml2 and its dependencies to fix CVE-2022-49043

### LINKED ISSUE #1725: Remove libxml2 and its dependencies from the Containerfile
CVE-2022-49043 is caused by libxml2 

Description for CVE-2022-49043 
xmlXIncludeAddNode in xinclude.c in libxml2 before 2.11.0 has a use-after-free.

ganeshmurthy:


Containerfile
```diff
@@ -71,7 +71,7 @@ RUN microdnf -y --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install \
 # Remove gnutls, libarchive and everything that depends on it. 
 # https://github.com/skupperproject/skupper-router/issues/1477
 # https://github.com/skupperproject/skupper-router/issues/1639
-RUN microdnf -y remove gnutls glib2 gobject-introspection libpeas microdnf gnupg2 gpgme libdnf json-glib libmodulemd librepo librhsm libsolv rpm rpm-libs libarchive libyaml libusbx systemd-libs
+RUN microdnf -y remove gettext-libs gettext libxml2 gnutls glib2 gobject-introspection libpeas microdnf gnupg2 gpgme libdnf json-glib libmodulemd librepo librhsm libsolv rpm rpm-libs libarchive libyaml libusbx systemd-libs
```
skitt, line 74: If `gettext` isn’t needed, wouldn’t it be better to skip installing in line 67, instead of installing it and then removing it?

ganeshmurthy: > wouldn’t it be better to skip installing it in line 67

Yes, agree. Nice catch. I have fixed it.

PR #1728, opened 2025-01-29
Build the skupper-router image from scratch

skitt:
This uses a separate "packager" stage to install *only* the required packages, with nothing else. This produces an image:
* with no extraneous packages
* with no content in base layers
* with no package manager
* that can still be processed by security scanning tools

Image size is reduced from 226MiB to 129MiB.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -53,22 +53,21 @@ RUN if [ "$PLATFORM" = "amd64" ]; then tar zxpf /qpid-proton-image.tar.gz --one-
 RUN if [ "$PLATFORM" = "arm64" ]; then tar zxpf /qpid-proton-image.tar.gz --one-top-level=/image && tar zxpf /skupper-router-image.tar.gz --one-top-level=/image && tar zxpf /libwebsockets-image.tar.gz --one-top-level=/image; fi
 RUN mkdir /image/licenses && cp ./LICENSE /image/licenses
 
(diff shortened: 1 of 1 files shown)


skitt:
FYI @ganeshmurthy — note that this will fail if other packages in the base image are used without being declared. This also has no labels, if any of the base image labels are necessary they will have to be copied across.

jiridanek:
I don't have any experience with this, actually. I just saw that comment on

and thought that it sounds interesting and might just work.

Looks like the approach has been around a long time and semi-official Red Hat blogs mention it, so it should not cause too many unexpected problems.

* https://opensource.com/article/18/5/containers-buildah
* https://www.redhat.com/en/blog/tiny-containers

> FYI @ganeshmurthy — note that this will fail if other packages in the base image are used without being declared.

That's one problem, but there was the same problem already before, if something was installed into the builder image but it was forgotten to install the non-devel pkg version into the run image.

I'm tempted to actually check for myself if this helped with image size any, compared to the microdnf uninstall.

ganeshmurthy:
> I'm tempted to actually check for myself if this helped with image size any, compared to the microdnf uninstall.

Main image size is 230 MB which the image from this PR is 129MB

jiridanek (approved):
wow, so much smaller? if it works, ship it!

edit: just noticed the size info in the PR description : facepalm :

PR #1730, opened 2025-01-30
Set up the user in the packager stage

skitt:
This allows shadow-utils to be removed in the final image, which reduces the exposure to CVEs in the relevant packages (shadow-utils and its dependencies). It also reduces the final image size by 3MiB.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -63,13 +63,14 @@ RUN dnf -y --setopt=install_weak_deps=0 --nodocs \
     libnghttp2 \
     hostname iputils \
     shadow-utils \
+ && chroot /output useradd --uid 10000 runner \
+ && dnf -y --installroot /output remove shadow-utils \
  && dnf clean all --installroot /output
 
 FROM scratch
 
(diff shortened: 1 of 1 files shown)


jiridanek:
That's something I considered suggesting yesterday, but decided against it. My suggestion actually would've been to not install `shadow-utils` in the first place, and only do

```
USER 10000
```

in the output image, for Kubernetes purposes.

Openshift will set random UID and add entry into /etc/passwd on its own.

And I faintly remember that I saw before some non-redhat images that either did not bother with having the `USER` actually existing, or they set it to the UID of the `nobody` user that was present by default in their /etc/passwd.

But since I can't vouch for "works in Kubernetes without any issues", I decided not to speak.

skitt:
@jiridanek yes, it might be simplest to just rely on the `USER` statement without actually adding entries in `/etc/passwd` and `/etc/group` in the image itself — it’s unlikely that anything in the router itself relies on finding a name for the uid (but I haven’t checked).

PR #1737, opened 2025-02-04
Enable Clair security scanning on Quay

skitt:
Since the router image was simplified, Quay reports scanning results, but these are misleading. It turns out that while Clair (the scanning tool used in Quay) can identify the packages shipped in the minimised images, it needs extra information to identify vulnerabilities in those packages; specifically, in UBI-based images, /root/buildinfo. See
https://github.com/quay/claircore/blob/main/rhel/repositoryscanner.go for reference.

This fixes that by copying the appropriate information from the "packager" image. While we're at it, copy the repository information too; this simplifies building further images on top of this one. Some WORKDIR statements are superfluous so those are removed too.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -70,13 +70,13 @@ RUN dnf -y --setopt=install_weak_deps=0 --nodocs \
 FROM scratch
 
 COPY --from=packager /output /
+COPY --from=packager /etc/yum.repos.d /etc/yum.repos.d
+COPY --from=packager /root/buildinfo /root/buildinfo
 
 USER 10000
 
-WORKDIR /
 COPY --from=builder /image /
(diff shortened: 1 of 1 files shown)


skitt:
See https://quay.io/repository/skitt/skupper-router/manifest/sha256:e308c94fe1314de68204615e7c1897615c04c340d39c98acb3a57869cc3a69ff?tab=vulnerabilities for a scan of an image with this fix; compare with the packages in https://quay.io/repository/skupper/skupper-router/manifest/sha256:6630cfda0e51dade490ed6a4d936bc9fd6c95fff80e3d373684a9e73fdc4601a?tab=packages

PR #1747, opened 2025-02-20
Fixes #1743: manage connection lifecycle on TLS ordinal mgmt

### LINKED ISSUE #1743: Connector-side: Instantiate/delete connections based on TLS ordinal updates

kgiusti:


src/adaptors/amqp/qd_connector.c
```diff
@@ -47,6 +47,59 @@ ALLOC_DEFINE(qd_connector_config_t);
...
+
+
+/**
+ * Scan the connector_config connectors and delete any that are no longer active
+ */
+static void qd_connector_config_cleanup_conns(void *context)
+{
+    ASSERT_MGMT_THREAD;  // only the mgmt thread can modify the connector list!
+
+    qd_connector_config_t *ctor_config = (qd_connector_config_t *) context;
```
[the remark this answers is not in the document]
ganeshmurthy: > 

Oops. That was my bad. Please ignore this comment

src/adaptors/amqp/connection_manager.c
```diff
@@ -369,7 +363,7 @@ QD_EXPORT void qd_connection_manager_start(qd_dispatch_t *qd)
     }
 
     while (ctor_config) {
-        qd_connector_config_connect(ctor_config);
+        qd_connector_config_activate(ctor_config);
         ctor_config = DEQ_NEXT(ctor_config);
     }
```
ganeshmurthy, line 369: Can we please remove this `first_start` static variable as part of this change even though this PR has nothing to do with it? I don't understand why it is there

src/adaptors/amqp/qd_connector.c
```diff
@@ -205,21 +262,16 @@ const char *qd_connector_policy_vhost(const qd_connector_t* ct)
 }
 
 
-bool qd_connector_connect(qd_connector_t *ct)
+void qd_connector_activate(qd_connector_t *ct)
 {
     sys_mutex_lock(&ct->lock);
-    if (ct->state != CTOR_STATE_DELETED) {
```
ganeshmurthy, line 211: This if statement `if (ct->state != CTOR_STATE_DELETED)`  was added as part of this crash fix - https://github.com/skupperproject/skupper-router/commit/725dda6d12f800ebd15e8828779d4753fa48179f
In a production, when there is no assert, this if statement would prevent a crash ?

src/adaptors/amqp/qd_connector.h
```diff
@@ -56,7 +56,7 @@ typedef struct qd_connector_t {
 
     /* Referenced by parent qd_connector_config_t and child qd_connection_t */
     sys_atomic_t              ref_count;
-    qd_timer_t               *timer;
+    qd_timer_t               *reconnect_timer;
```
ganeshmurthy, line 59: Excellent change renaming `timer` to `reconnect_timer`

src/adaptors/amqp/qd_connector.c
```diff
@@ -439,27 +500,90 @@ void qd_connector_remove_connection(qd_connector_t *connector, bool final, const
...
+                ctor = DEQ_NEXT(ctor);
+            }
+        }
+
+        // Create a new set of connectors that will use the updated TLS credentials
+
+        ctor_config->tls_ordinal = new_ordinal;
+
+        for (int i = 0; i < ctor_config->data_connection_count; i++) {
+            ctor = qd_connector_create(ctor_config, true);
```
ganeshmurthy, line 568: Can we please create two new functions `qd_connector_create_data_conn()` and `qd_connector_create_control_conn()` and remove the boolean parameter ?

src/adaptors/amqp/qd_connector.c
```diff
@@ -439,27 +500,90 @@ void qd_connector_remove_connection(qd_connector_t *connector, bool final, const
...
+                qd_log(LOG_SERVER, QD_LOG_ERROR, "Failed to create data Connector %s: resource allocation failed", ctor_config->config.name);
+            } else {
+                DEQ_INSERT_HEAD(ctor_config->connectors, ctor);
+                qd_connector_activate(ctor);
+            }
+        }
+
+        ctor = qd_connector_create(ctor_config, false);
+        if (!ctor) {
+                qd_log(LOG_SERVER, QD_LOG_ERROR, "Failed to create data Connector %s: resource allocation failed", ctor_config->config.name);
```
ganeshmurthy, line 581:
```suggestion
                qd_log(LOG_SERVER, QD_LOG_ERROR, "Failed to create admin connector %s: resource allocation failed", ctor_config->config.name);
```

src/router_core/connections.c
```diff
@@ -1456,42 +1456,48 @@ void qdr_process_addr_attributes_CT(qdr_core_t *core, qdr_address_t *addr)
 static void qdr_connection_group_setup_CT(qdr_core_t *core, qdr_connection_t *conn)
 {
     const char *correlator = conn->connection_info->group_correlator;
+    uint64_t    ordinal    = conn->connection_info->group_ordinal;
 
     assert(conn->role == QDR_ROLE_INTER_ROUTER);
     qd_log(LOG_ROUTER_CORE, QD_LOG_DEBUG,
-           "CGROUP qdr_connection_group_setup_CT - %lx (%s)",
-           (ulong) conn, conn->connection_info->host);
+           "CGROUP setup group '%s' for parent [C%"PRIu64"] with ordinal=%"PRIu64" (host=%s)",
```
ganeshmurthy, line 1463:
```suggestion
           "Connection group setup group '%s' for parent [C%"PRIu64"] with ordinal=%"PRIu64" (host=%s)",
```

ganeshmurthy: People might confuse CGROUP with the Linux cgroup

src/adaptors/amqp/qd_connector.c
```diff
@@ -205,21 +262,16 @@ const char *qd_connector_policy_vhost(const qd_connector_t* ct)
 }
 
 
-bool qd_connector_connect(qd_connector_t *ct)
+void qd_connector_activate(qd_connector_t *ct)
```
ganeshmurthy, line 265:
```suggestion
void qd_connector_schedule_connection(qd_connector_t *ct)
```

src/router_core/connections.c
```diff
@@ -1456,42 +1456,48 @@ void qdr_process_addr_attributes_CT(qdr_core_t *core, qdr_address_t *addr)
...
+           correlator, conn->identity, ordinal, conn->connection_info->host);
 
     if (strnlen(correlator, QD_DISCRIMINATOR_SIZE) > 0) {
+#if 0   // KAG TODO FIX: allow dupes for now:
```
ganeshmurthy, line 1467: Should this `TODO` be resolved ?

src/router_core/connections.c
```diff
@@ -1456,42 +1456,48 @@ void qdr_process_addr_attributes_CT(qdr_core_t *core, qdr_address_t *addr)
...
+#endif
 
         //
         // Record the group's correlator in the core record.
         // Check the unallocated member list for matching correlators.  Import the
         // matches into this connection's group.
         //
         assert(core->group_correlator_by_maskbit[conn->mask_bit][0] == '\0');
-        qd_log(LOG_ROUTER_CORE, QD_LOG_DEBUG, "CGROUP     correlator[%d] = %s", conn->mask_bit, correlator);
+        qd_log(LOG_ROUTER_CORE, QD_LOG_DEBUG, "CGROUP [C%"PRIu64"] recording correlator[%d]='%s'", conn->identity, conn->mask_bit, correlator);
```
ganeshmurthy, line 1491: Replace all `CGROUP` with `Connection group`

src/adaptors/amqp/connection_manager.c
kgiusti: The first_start var is a bit weird, but it seems to be related to detecting listener failures (and setting the listener->exit_on_error flag).  I'm not sure of the ramifications of removing that flag.  Since I've got to deal with listener lifecycle later (see Issue #1753) I'd recommend we hold off until then.

src/adaptors/amqp/qd_connector.c
kgiusti: That doesn't make sense to me.  When I see "schedule" it implies that there's possibly a need to re-schedule in the future.  That isn't the case here.  The only thing that is re-scheduled is the connection re-connect attempt and that's handled auto-magically internal to the connector itself.

The intent of the name is convey that the connection manager is kicking off the internal connector state machine.  Once that's done the connection manager is totally out of the picture as far as scheduling the connection is concerned (IOW: the connection manager does not monitor the connection state nor does it periodically re-schedule the connector).

src/adaptors/amqp/connection_manager.c
ganeshmurthy: `first_start` is set to true at the beginning of the function and never modified to false until the last line of the function which seems weird. That can be dealt with later. Thanks.

src/adaptors/amqp/qd_connector.c
kgiusti: Yes the code will still prevent re-connection when the deleted state is present but this patch expands on when the re-connect is no longer legal to do beyond just the deleted state. Remember when the cert is rotated old conns are left up but the reconnect is now disabled - this is new functionality.  So the connector state is still Open [as reported by management] *but* reconnect is disabled (again this is new functionality introduced by this patch).

So this new functionality essentially expands the conditions for which reconnect should not be called.  Essentially:
```
if (state == DELETED or state==not deleted but reconnect_disabled flag is true) {
    // cannot reconnect
 else
    // can reconnect
}
```

Logically very clunky/error prone (need to check several vars to determine whether re-connect is allowed whenever the timer is used).

So instead this patch uses the value of connector->reconnect_timer itself as a flag to determine if reconnect is enable (or better said: can we schedule reconnect?):

```
if (connector->reconnect_timer == 0)
   // reconnect is disabled
 else
   // we can schedule reconnect
```

So you'll see in the code wherever reconnect is disabled (like entry to DELETE state or when the connection is rotated) the connector->reconnect_timer is freed and set to zero.  And wherever a re-connect is attempted the connector->reconnect_timer is checked to be non-zero before scheduling it.  This makes the reconnect_timer value a *flag* as well as a pointer to a timer. And more importantly it makes accidentally reconnecting impossible, since you'd be calling schedule on a freed (zero pointer) timer will crash in CI.

TL;DR: this preserves the DELETE behavior but uses a more general way to cancel reconnect.

src/router_core/connections.c
kgiusti: Good point - I should've thought of that!

src/router_core/connections.c
kgiusti: Actually that TODO gets fixed and removed in the next patch #1748 
For now we need to allow dups because the control links are not removed yet.

src/adaptors/amqp/qd_connector.c
ganeshmurthy: Makes sense to me. Should the `qd_connector_handle_transport_error()` function in `qd_connector.c` also be modified to use the same logic i.e. don't use the `CTOR_STATE_DELETED` but instead use the 
`connector->reconnect_timer` ?

tests/system_tests_cert_rotation.py
```diff
@@ -0,0 +1,199 @@
...
+        ]
+        config.extend(test_config)
+        return self.tester.qdrouterd(name, Qdrouterd.Config(config), **kwargs)
+
+    def get_inter_router_conns(self, router):
+        mgmt = router.management
+        conns = mgmt.query(type=CONNECTION_TYPE).get_dicts()
+        return [c for c in conns if 'inter-router' in c['role']]
+
+    def wait_inter_router_conns(self, router, count):
```
ganeshmurthy, line 54: Seems generic and useful enough to move this function to `system_test.py`

src/adaptors/amqp/qd_connector.c
kgiusti: Oh, I didn't realize that was happening.  Probably should - it would be bad if it tried to reschedule a connector that should not reconnect.

I'll take a look.

kgiusti: Follow up:  It appears that qd_connector_handle_transport_error does not attempt to reconnect.  IIUC all it does is interrogate the proton transport for any error condition associated with the conn failure and logs it.  I suspect the check for DELETED is merely an effort to stop logging about a connection that was deleted by management. I'd like to keep the logging in the case where the connector hasn't been deleted but will not reconnect so we can track when a connection that is carrying streaming data fails (since the app will see a connection reset).

src/adaptors/amqp/qd_connector.c
kgiusti: I like this idea but when I was looking into making that change it didn't make sense because qd_connector_create() can create other types of connectors as well (edge, inter-edge, route-container, normal).  All this is hidden in the qd_connector_create() constructor, which is good.

The flag is a hack that basically tells the constructor "ignore the configured role and use inter-router-data instead".

Given all this I'd prefer not to have a dedicated qd_connector_create_data_conn() because that would require duplicating all of qd_connector_create() and just leave out a few lines of data connection specific logic.

Or maybe instead of a true/false parameter we define an enum "ROLE_OVERRIDE_DATA_CONN" and pass that?

src/adaptors/amqp/qd_connector.c
kgiusti: We should leave out admin and just say "connector" becase the connector role may not be control (could be 'edge', etc).

src/adaptors/amqp/qd_connector.c
```diff
@@ -439,27 +500,90 @@ void qd_connector_remove_connection(qd_connector_t *connector, bool final, const
...
+                ctor = DEQ_NEXT(ctor);
+                DEQ_REMOVE(ctor_config->connectors, delete_me);
+                qd_connector_close(delete_me);
+                qd_connector_decref(delete_me);
+            } else {
+                ctor = DEQ_NEXT(ctor);
+            }
+        }
+
+        // Create a new set of connectors that will use the updated TLS credentials
```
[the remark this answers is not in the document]
kgiusti: See latest push

src/adaptors/amqp/qd_connector.c
kgiusti: Gah, I'm stumped.

I ended up moving all the connector creation logic to a local helper function and commented that. Puts it all in one place for now.  Let me know what you think.

src/adaptors/amqp/qd_connector.c
kgiusti: See latest patch

src/router_core/connections.c
kgiusti: Done.

src/router_core/connections.c
kgiusti: fixed

tests/system_tests_cert_rotation.py
kgiusti: moved

src/adaptors/amqp/qd_connector.c
ganeshmurthy: Makes sense !

tests/system_tests_cert_rotation.py
```diff
@@ -0,0 +1,194 @@
...
+        ]
+        config.extend(test_config)
+        return self.tester.qdrouterd(name, Qdrouterd.Config(config), **kwargs)
+
+    def wait_inter_router_conns(self, router, count):
+        ok = retry(lambda rtr=router, ct=count:
+                   len(rtr.get_inter_router_conns()) == ct)
+        self.assertTrue(ok, f"Failed to get {count} i.r. conns: {router.get_inter_router_conns()}")
+
+    def test_01_ordinal_updates(self):
```
Karen-Schoener, line 54: Out of curiosity, at some point, will there be a test that verifies graceful rotation of inter-router certs/connections?

Where the test would exercise creating a 2nd/new inter-router link, verifying the new inter-router link is established & healthy, and then deleting the 1st/old inter-router link?

ganeshmurthy (approved):

tests/system_tests_cert_rotation.py
kgiusti: Excellent question!  I'm literally working on testing right now - you can see the beginning WIP draft here:
https://github.com/skupperproject/skupper-router/pull/1755

Specifically I'm building a test that:

1. Creates two connected interior routers with TLS configured (sslProfile)
2. Creates a set of TCP streaming clients passing flows across both routers
3. The test will then simulate rotation by bumping up the tlsOrdinal configuration attribute on the Connector's sslProfile
4. The test verifies that a new set of inter-router connections are created based on the new sslProfile config
5. The test verifies that the older inter-router connections are still present and the existing TCP flows are still passing data
6. The test then creates another set of TCP flows and verifies the new flows use the updated (new) inter-router connections
7. The test then "expires" the certificate by bumping the Connector's sslProfile oldestValidOrdinal beyond the original value
8. The test then verifies that the old connections are closed and the first set of TCP flows are dropped
9. The test verifies that the second set of TCP flows are still active and passing traffic.
10. Lastly the test interrogates the router's topological state change timestamp to verify that routes were not re-computed during the test

I think that this test would exercise the whole logic chain involved in certificate rotation on inter-router connections. I'll be doing other negative tests as well but those are TBD.

Let me know what you think.

Karen-Schoener: Sounds great, thanks for explaining!

Karen-Schoener (approved):

PR #1749, opened 2025-03-03
Fixes skupperproject#1750: Added asan and tsan to the new arm64 runner

### LINKED ISSUE #1745: Remove the hosted runner (Raspberry Pi) that was doing the arm64 builds, use the arm runner now provided by github CI

ganeshmurthy:
…erry Pi (#1746)

### DIFF
--- a/tests/system_tests_edge_router1.py
+++ b/tests/system_tests_edge_router1.py
@@ -1059,7 +1059,7 @@ class StreamingMessageTest(TestCase):
                "-d"]
         self._container_index += 1
         env = dict(os.environ, PN_TRACE_FRM="1")
-        return self.popen(cmd, expect=expect, env=env)
+        return self.popen(cmd, expect=expect, env=env, abort=True)
 
     def spawn_sender(self, router, count, address, expect=None, size=None):
         if expect is None:
@@ -1075,7 +1075,7 @@ class StreamingMessageTest(TestCase):
                "-d"]
         self._container_index += 1
         env = dict(os.environ, PN_TRACE_FRM="1")
-        return self.popen(cmd, expect=expect, env=env)
+        return self.popen(cmd, expect=expect, env=env, abort=True)
 
     def spawn_clogger(self, router, count, address,
                       size, pause_ms, expect=None):
@@ -1089,7 +1089,7 @@ class StreamingMessageTest(TestCase):
                "-D",
                "-P", str(pause_ms)]
         env = dict(os.environ, PN_TRACE_FRM="1")
-        return self.popen(cmd, expect=expect, env=env)
+        return self.popen(cmd, expect=expect, env=env, abort=True)
 
     def _streaming_test(self, address):
 
@@ -1326,7 +1326,6 @@ class StreamingMessageTest(TestCase):
         self.EB1.wait_address("balanced/test-address", subscribers=2)
 
         # this will block one of the above receivers with a streaming message
-
         clogger = self.spawn_clogger(self.EA1,
                                      count=0,
                                      address="balanced/test-address",
@@ -1349,11 +1348,6 @@ class StreamingMessageTest(TestCase):
         for rx in balanced_rx:
             rx.teardown()
         clogger.teardown()
-
-        self._wait_address_gone(self.EA1, "balanced/test-address")
-        self._wait_address_gone(self.EB1,  "balanced/test-address")
-        self._wait_address_gone(self.INT_A,  "balanced/test-address")
-        self._wait_address_gone(self.INT_B,  "balanced/test-address")
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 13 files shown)


tests/system_tests_edge_router1.py
```diff
@@ -1350,11 +1349,6 @@ def test_13_streaming_balanced_parallel(self):
             rx.teardown()
         clogger.teardown()
 
-        self._wait_address_gone(self.EA1, "balanced/test-address")
```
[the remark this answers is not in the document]
ganeshmurthy: The test was waiting on the `_wait_address_gone()` and looping inside that function and ultimately timed out. I removed it because `test_13_streaming_balanced_parallel()` was the last test in the test suite (`StreamingMessageTest`). I understand that removing this code would be a problem if we ever add a `test_14` but this test annoyingly keeps failing intermittently in aarch64. I could put it back if you feel strongly about it.

tests/system_tests_interior_sync_up.py
```diff
@@ -136,13 +144,8 @@ def delay_timeout(self):
         self.add_receivers()
 
     def add_receivers(self):
-        if len(self.receivers) < self.count:
+        while len(self.receivers) < self.count:
```
kgiusti, line 147: This changes the fundamental behavior of the test.  But from looking at the router source code there is no MAX_KEPT_DELTAS appearing anywhere so maybe the functionality this test was intended to exercise is no longer present?  @ted-ross does this test ring a bell?

tests/system_tests_edge_router1.py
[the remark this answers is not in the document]
kgiusti: This makes sense, but it leaves a time bomb for some poor unsuspecting dev like myself when I forget that test_13_xxx must be last.  Would you be ok with changing the test number to some obvious indication that it must be last and drop a comment?  I know I've used the prefix _test_9999_xxx_ in those other test cases were a test needed to be run last in a test case (see tcp adaptor tests for an example)

tests/system_tests_priority.py
```diff
@@ -290,14 +261,17 @@ def __init__(self, parent, test_name, client_addrs, destination, magic_msg_prior
         self.A_addr          = self.client_addrs[0]
         self.B_addr          = self.client_addrs[1]
         self.C_addr          = self.client_addrs[2]
+        self.test_name       = test_name
```
kgiusti, line 264: This test makes me weep for all mankind.

I can only guess as to what it's trying to verify. What really concerns me is that with all the counter/timeouts and now multiple retries this test will eventually pass regardless of the router's behavior.

Is this trying to ensure all messages are passed via links that match the message or addresses priorities? If that's the case this test can be simplified to two routers, right?

I know this test isn't introduced by the patch but it seems like the patch makes it unlikely to fail even if the behavior is not correct due to the additional retries added by the patch and the fact that counters are not reset between retries.

tests/system_tests_topology_addition.py
```diff
@@ -208,8 +208,12 @@ def addRouter(self, name, more_config) :
             + more_config
 
         config = Qdrouterd.Config(config)
+        new_router = TopologyAdditionTests.tester.qdrouterd(name, config, wait=True)
+        TopologyAdditionTests.routers.append(new_router)
 
-        TopologyAdditionTests.routers.append(TopologyAdditionTests.tester.qdrouterd(name, config, wait=True))
+        if neighbors is not None:
```
kgiusti, line 214: Very nice!

tests/system_tests_priority.py
ganeshmurthy: > This test makes me weep for all mankind.
> 
> I can only guess as to what it's trying to verify. What really concerns me is that with all the counter/timeouts and now multiple retries this test will eventually pass regardless of the router's behavior.
> 
> Is this trying to ensure all messages are passed via links that match the message or addresses priorities? If that's the case this test can be simplified to two routers, right?
> 
> I know this test isn't introduced by the patch but it seems like the patch makes it unlikely to fail even if the behavior is not correct due to the additional retries added by the patch and the fact that counters are not reset between retries.

I agree that this test is complicated ! It took a long time for me to understand what it was doing and even longer to "fix" it. I will create an issue to simplify the test

ganeshmurthy: https://github.com/skupperproject/skupper-router/issues/1751

kgiusti: Thanks - resolving

kgiusti (approved):
Your patch LGTM.

tests/system_tests_interior_sync_up.py
ted-ross: MAX_KEPT_DELTAS is no longer in the router.  When the "mobile" functionality moved from Python to C, that feature was removed.  MAX_KEPT_DELTAS is now effectively 1.

tests/system_tests_interior_sync_up.py
```diff
@@ -97,7 +95,7 @@ def __init__(self, host_a, host_b, inter_router_port):
         self.timer             = None
         self.poll_timer        = None
         self.delay_timer       = None
-        self.count             = 2000
+        self.count             = 1000
         self.delay_count       = 12   # This should be larger than MAX_KEPT_DELTAS in mobile.py
```
ted-ross, line 99: A value of 2 would achieve the desired result.

PR #1752, opened 2025-03-05
New locks in message.c to eliminate contention on content lock

mgoulish:
Performance changes for flimflam workloads after this change:

                   Before        After                      Change

   Builtin      3.98         4.04     G Bits/s       + 1.5 %

   Iperf3       7.87         7.64     G Bits/s       - 2.9 %

   H2load       8.36         9.00     K Ops/s      + 7.1 %

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1112,6 +1112,8 @@ qd_message_t *qd_message(void)
 
     ZERO(msg->content);
     sys_mutex_init(&msg->content->lock);
+    sys_mutex_init(&msg->content->producer_lock);
+    sys_mutex_init(&msg->content->consumer_lock);
     sys_atomic_init(&msg->content->aborted, 0);
     sys_atomic_init(&msg->content->discard, 0);
     sys_atomic_init(&msg->content->no_body, 0);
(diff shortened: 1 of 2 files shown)


src/message_private.h
```diff
@@ -74,7 +74,12 @@ typedef struct {
 //
 
 typedef struct {
-    sys_mutex_t          lock;
+    sys_mutex_t          lock,
+                         producer_lock,                   // These locks prevent either side from activating
```
kgiusti, line 78: For code consistency we avoid the multiple vars per type syntax in favor of single var per type declaration.  In other words:

```
// Instead of
sys_mutex_t a, b, c;
// we do
sys_mutex_t a;
sys_mutex_t b;
sys_mutex_t c;
```
The extra bytes do slow the executable down a bit but it's really not noticeable.

src/message.c
```diff
@@ -1112,6 +1112,8 @@ qd_message_t *qd_message(void)
 
     ZERO(msg->content);
     sys_mutex_init(&msg->content->lock);
+    sys_mutex_init(&msg->content->producer_lock);
```
kgiusti, line 1115: There should be corresponding sys_mutex_free() calls for the new locks in qd_message_free() where the content is freed.

kgiusti (changes requested):
This is an excellent patch!  See inline for requested changes.
The system_test_two_routers failures are unrelated to this change (seen on main as well).

src/message_private.h
ted-ross, line 78: These should have more descriptive, less misleading names.  {producer|consumer}_activation_lock

kgiusti (approved):
LGTM - verified that the test failures are due to an unrelated proton issue.
Approval conditional to resolving @ted-ross requests

ganeshmurthy (approved):

ted-ross (approved):

PR #1763, opened 2025-03-31
Fixes #1753: Expire Listener-side connection on certificate rotation

### LINKED ISSUE #1753: Teardown old Listener-side inter-router connections on sslProfile oldestValidOrdinal update

kgiusti:
When an AMQP listener's sslProfile oldestValidOrdinal attribute is updated scan all connections that are children of that listener and check if the connection's tlsOrdinal value is less than the new oldestValidOrdinal value. If it is then close the connection.

Patch includes additional efforts to make the code fluffy:

1) Move qd_listener_t-specific code out of connection_manager.c and move it to the qd_listener.c file.
2) Modify the qd_connection_t deferred callback API to allow an extra context parameter to be passed to the callback. Hide the implementation details by moving the data structure from a header file into qd_connection.c.
3) Remove dead code.

### DIFF
--- a/src/adaptors/amqp/qd_listener.c
+++ b/src/adaptors/amqp/qd_listener.c
@@ -18,8 +18,10 @@
  */
 
 #include "qd_listener.h"
+#include "private.h"
 #include "qd_connection.h"
 #include "http.h"
+#include "entity.h"
 
 #include "qpid/dispatch/server.h"
 #include "qpid/dispatch/log.h"
@@ -36,6 +38,9 @@
 ALLOC_DEFINE(qd_listener_t);
 
 static const int BACKLOG = 50;  /* Listening backlog */
+
+static void handle_listener_ssl_profile_mgmt_update(const qd_tls_config_t *config, void *context);
+
 
 static void on_accept(pn_event_t *e, qd_listener_t *listener)
 {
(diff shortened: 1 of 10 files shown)


tests/system_tests_cert_rotation.py
```diff
@@ -527,6 +530,199 @@ def test_04_rotate_storm(self):
...
+                                             server_addr=('0.0.0.0', tcp_connector_port_2),
+                                             client_count=4, poll_timeout=0.2)
+        ok = retry(lambda: new_tcp_streamer.active_clients == 4)
+        self.assertTrue(ok, f"Streaming clients failed {new_tcp_streamer.active_clients}")
+        begin_recv = new_tcp_streamer.bytes_received
+        ok = retry(lambda: new_tcp_streamer.bytes_received > begin_recv)
+        self.assertTrue(ok, f"Failed to stream data {new_tcp_streamer.bytes_received}")
+
+        # Expect an additional 2 streaming links per TCP flow (links are uni-directional)
+        self.assertEqual(28, len(router_L.get_active_inter_router_data_links()),
```
Karen-Schoener, line 687: Is there a way that a user can verify that the new link (that is using the new SSL profile) is operational?

Can the user check that the new link is established before they increment the oldestValidOrdinal?

kgiusti: Yes that's a good idea. The connection's management object has the value of the TLS ordinal in use.  I'll update the test to verify the new conns have the expected latest ordinal.

kgiusti: @Karen-Schoener this patch verifies that the tlsOrdinal is present in the connection's management object. It verifies that the tlsOrdinals values are correct after the certificates have been rotated. 

https://github.com/skupperproject/skupper-router/pull/1763/commits/e6b203f627b4d849929699b26416f3f1c2016e55

Let me know if there's anything else the tests should cover - thanks!

Karen-Schoener: Just to make sure I am following correctly: is the tlsOrdinal availble for both the listener and for the connector?

(Can both sides of the inter-router link query the tlsOrdinal that is currently in use?)

Thanks!  Karen

kgiusti: Ah sorry - I misunderstood.  

To be pedantic the tlsOrdinal is not present in the listener or connector configuration records. 

The tlsOrdinal applies to the connection itself. So it **is present** in the connection management object, which represents the state of an active connection.

See the following link to the schema representation of the connection management object: https://github.com/skupperproject/skupper-router/blob/71e90b5e857217c511489cbe2d81b58e25151bf6/python/skupper_router/management/skrouter.json#L1751

You can use skmanage to get a dump of all connection management objects from a router:
```
skmanage -b 127.0.0.1:5672 QUERY --type connection
```

The value of the tlsOrdinal in the connection management object is copied from the sslProfile.tlsOrdinal attribute when the inter-router connection is established. Its value remains constant for the lifetime of the connection (In other words it doesn't change if the corresponding sslProfile.tlsOrdinal is changed).

However: Keep in mind that each router has a connection management object that reflects the connection's state from the point of view of that router.  In other words, for a single inter-router tcp connection there are two connection management objects: one from each router.

However part II: since the values for the sslProfile.tlsOrdinal are configured independently on each router it's very likely that their values will be different at each end of the connection. So the tlsOrdinals in the connection management objects for a given connection will not have the same values. This means you cannot use the value of the tlsOrdinal to match up peer connection management objects.

There are ways to match up peer connection management objects but that's a different kettle of fish (which we can discuss if you want).

Karen-Schoener (approved):

src/adaptors/amqp/qd_listener.c
```diff
@@ -113,20 +118,116 @@ static void handle_listener(pn_event_t *e, qd_server_t *qd_server, void *context
...
+            // qd_tls_config() sets qd_error_message():
+            qd_log(LOG_CONN_MGR, QD_LOG_ERROR, "Failed to configure TLS for Listener %s: %s",
+                   li->config.name, qd_error_message());
+            qd_listener_decref(li);
+            return 0;
+        }
+
+        li->tls_ordinal              = qd_tls_config_get_ordinal(li->tls_config);
+        li->tls_oldest_valid_ordinal = qd_tls_config_get_oldest_valid_ordinal(li->tls_config);
+        qd_tls_config_register_update_callback(li->tls_config, li,
```
ganeshmurthy, line 178:
```suggestion
     if(li->tls_config)
        qd_tls_config_register_update_callback(li->tls_config, li,
```

ganeshmurthy: I was wondering if `qd_listener_create()` and `qd_listener_delete()` can be moved into its own commit so that this commit only contains the changes related to expiring listener side connection.

PR #1764, opened 2025-04-01
Fixes #1762: log error when connection group is split among routers

### LINKED ISSUE #1762: Detect when inter-router connections are incorrectly split across peer routers
If two routers are connected via a load balancer we've seen cases where inter-router-data connections from a given Connector group are split across remote routers.

All connections within a connection group are treated as a single path to a peer router as far as the routing protocol is concerned.  Having connections within a group connect to different routers (due to load balancing) will lead to unpredictable message delays and failures.

Using a load balancer to spread inter-router connections across multiple downstream routers is NOT SUPPORTED.

Currently it is very hard to diagnose this issue in the field. It would be acceptable to address this by having the router detect when connections within the same connection group are split across routers and issue error-level log alerts.

kgiusti:
Also remove duplicate parameter for remote container identifier in the connection open API.


tests/system_tests_split_path.py
```diff
@@ -0,0 +1,235 @@
...
+
+import os
+import selectors
+import socket
+from threading import Event, Thread
+from system_test import TestCase, main_module, Qdrouterd, unittest
+from system_test import TIMEOUT, Logger
+
+
+class MyLoadBalancer:
```
ganeshmurthy, line 33: Can we try using haproxy to do tcp load balancing (round robin) instead of writing our own load balancer ?

https://www.haproxy.com/documentation/haproxy-configuration-tutorials/load-balancing/tcp/

tests/system_tests_split_path.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def test_01(self):
+        """
+        Simulate splitting connections to the wrong routers.  Creates a single
+        connector-based router and two listener-based routers. A mock load
+        balancer sits between the connector router and the listeners. The mock
+        load balancer splits the connection group connections between the two
+        listener routers
+        """
+        data_conn_count = 8
```
ganeshmurthy, line 197: In production, the data_conn_count usually defaults to 2, can we leave it at 2 to mimic production use case ?

tests/system_tests_split_path.py
ganeshmurthy: Nginx also allows TCP loadbalancing but unfortunately that feature is only available to `Nginx Plus` users.
https://docs.nginx.com/nginx/admin-guide/load-balancer/tcp-udp-load-balancer/

tests/system_tests_split_path.py
kgiusti: Done.

tests/system_tests_split_path.py
kgiusti: I don't have any other test cases that require haproxy and the mock load balancer gives the patch full coverage. Can haproxy be added to the CI infrastructure as a follow on to this patch?

PR #1775, opened 2025-04-10
Restore buildinfo with its new location

skitt:
The UBI base images now ship buildinfo in /usr/share rather than /root; this updates the container build to match.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -71,6 +71,7 @@ FROM scratch
 
 COPY --from=packager /output /
 COPY --from=packager /etc/yum.repos.d /etc/yum.repos.d
+COPY --from=packager /usr/share/buildinfo /usr/share/buildinfo
 
 USER 10000
 


ganeshmurthy:
> /cc @ganeshmurthy

Ah, so the `/root/buildinfo` is no more and the new path for the buildinfo is `/usr/share/buildinfo`
Any thoughts on why this was failing only on CI but was passing locally for me ? Thanks

skitt:
> > /cc @ganeshmurthy
> 
> Ah, so the `/root/buildinfo` is no more and the new path for the buildinfo is `/usr/share/buildinfo` Any thoughts on why this was failing only on CI but was passing locally for me ? Thanks

I could get it to build locally, but then I realised my local UBI base image was outdated — `podman pull registry.access.redhat.com/ubi9/ubi:latest` updated the image and broke the build, same as in CI.

ganeshmurthy:
> > > /cc @ganeshmurthy
> > 
> > 
> > Ah, so the `/root/buildinfo` is no more and the new path for the buildinfo is `/usr/share/buildinfo` Any thoughts on why this was failing only on CI but was passing locally for me ? Thanks
> 
> I could get it to build locally, but then I realised my local UBI base image was outdated — `podman pull registry.access.redhat.com/ubi9/ubi:latest` updated the image and broke the build, same as in CI.

I usually build with the `no-cache` option and it did not fail which surprised me.
`podman build --no-cache -t local/skupper-router:local -f ./Containerfile .`

skitt:
Ah, but `--no-cache` only disables the local layer cache, not the image cache.

ganeshmurthy:
> Ah, but `--no-cache` only disables the local layer cache, not the image cache.



> Ah, but `--no-cache` only disables the local layer cache, not the image cache.

Makes sense to me. The podman doc could point that out. I learned something new from you today. I am merging this to main - thank you very much

```

--no-cache

Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
```

PR #1777, opened 2025-04-10
Copy buildinfo either from /root or /usr/share

skitt:
UBI base images might have one or the other, so handle both instead of changing the build to match changing base images.


skitt:
/cc @ganeshmurthy 

We’ve seen base images flip back in forth in other projects :frowning_face:

PR #1788, opened 2025-04-30
Fixes #1787: Multiarch skupper-router container image build using doc…

### LINKED ISSUE #1787: Use the docker buildx tool to create multiarch container image for arches amd64, arm64, s390x, ppc64le
Currently, we are creating multiarch images consisting of the amd64 and arm64 images.  We need to add s390x and ppc64le as the supported architectures but we do not have machines that could produce builds in those arches.

docker buildx is a convenient way of producing mutiarch images. It uses QEMU underneath and takes about 3.5 hours to do a multiarch build for all 4 arches.

ganeshmurthy:
…ker buildx.


fgiorgetti (approved):
Looks great to me! Thank you Ganesh.

PR #1792, opened 2025-05-30
Temporarily stop creation of vflow observability records if too many at

mgoulish:
once. To prevent large memory growth in some test scenarios.

This one uses atomic variables to prevent multithread fratricide.
The use of these variables did not measurably slow execution relative to non-multithread-aware code.

### DIFF
--- a/include/qpid/dispatch/vanflow.h
+++ b/include/qpid/dispatch/vanflow.h
@@ -57,6 +57,34 @@ typedef enum vflow_record_type {
     VFLOW_RECORD_BIFLOW_APP    = 0x11,  // Bidirectional Application (L7) flow
 } vflow_record_type_t;
 
+#define VFLOW_RECORD_MAX VFLOW_RECORD_BIFLOW_APP // Update this if you add new record types to vflow_record_type
+
+// Most record-types support critical functionality and must be emitted every time the corresponding
+// event occurs. But a few record-types support observability functionality that can be interrupted,
+// if necessary, to prevent excessive memory growth.
+// This array identifies which is which.
+static const bool observability_records[VFLOW_RECORD_MAX + 1] = {
+    false,     // VFLOW_RECORD_SITE          (0x00)
+    false,     // VFLOW_RECORD_ROUTER        (0x01)
+    false,     // VFLOW_RECORD_LINK          (0x02)
+    false,     // VFLOW_RECORD_CONTROLLER    (0x03)
+    false,     // VFLOW_RECORD_LISTENER      (0x04)
+    false,     // VFLOW_RECORD_CONNECTOR     (0x05)
+    true,      // VFLOW_RECORD_FLOW          (0x06)
+    false,     // VFLOW_RECORD_PROCESS       (0x07)
+    false,     // VFLOW_RECORD_IMAGE         (0x08)
+    false,     // VFLOW_RECORD_INGRESS       (0x09)
+    false,     // VFLOW_RECORD_EGRESS        (0x0a)
+    false,     // VFLOW_RECORD_COLLECTOR     (0x0b)
+    false,     // VFLOW_RECORD_PROCESS_GROUP (0x0c)
+    false,     // VFLOW_RECORD_HOST          (0x0d)
+    false,     // VFLOW_RECORD_LOG           (0x0e)
+    false,     // VFLOW_RECORD_ROUTER_ACCESS (0x0f)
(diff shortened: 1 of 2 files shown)


include/qpid/dispatch/vanflow.h
```diff
@@ -57,6 +57,34 @@ typedef enum vflow_record_type {
     VFLOW_RECORD_BIFLOW_APP    = 0x11,  // Bidirectional Application (L7) flow
 } vflow_record_type_t;
 
+#define VFLOW_RECORD_MAX VFLOW_RECORD_BIFLOW_APP // Update this if you add new record types to vflow_record_type
+
+// Most record-types support critical functionality and must be emitted every time the corresponding
+// event occurs. But a few record-types support observability functionality that can be interrupted,
+// if necessary, to prevent excessive memory growth.
+// This array identifies which is which.
+static const bool observability_records[VFLOW_RECORD_MAX + 1] = {
```
ted-ross, line 66: "Observability" is not the correct characterization for this.  All of VanFlow is about observability.  A better term would be something like "discretionary", or "nonessential"

src/vanflow.c
```diff
@@ -484,6 +488,15 @@ static void _vflow_start_record_TH(vflow_work_t *work, bool discard)
 
     vflow_record_t *record = work->record;
 
+    // If this is one of the record types that supports observability,
+    // count it. If the number currently existing is above the threshold,
+    // indicate that we should temporarily stop producing them.
+    if (observability_records[record->record_type]) {
+        if (sys_atomic_inc(&state->observability_record_count) >= 100) {
```
ted-ross, line 495: This threshold should not be a numeric literal.  At the very least, it should be symbolically defined at the top of the file.  Also, 100 is way too small of a threshold for this feature.  The same applies for the lower limit (20).

src/vanflow.c
```diff
@@ -1829,6 +1851,9 @@ vflow_record_t *vflow_start_record_custom_id(vflow_record_type_t record_type, vf
 //=====================================================================================
 vflow_record_t *vflow_start_record(vflow_record_type_t record_type, vflow_record_t *parent)
 {
+    if (observability_records[record_type] && state->emit_observability_records == false) {
```
ted-ross, line 1854: Is `==` a valid operator for atomics?  Does this work?  I think you need to use `sys_atomic_get` to retrieve the value of the counter.

src/vanflow.c
```diff
@@ -1861,6 +1886,12 @@ vflow_record_t *vflow_start_co_record_iter(vflow_record_type_t record_type, qd_i
     // search/replacement algorithm in _vflow_process_co_record_TH will need to be re-written in a more general way.
     //
     assert(record_type == VFLOW_RECORD_BIFLOW_TPORT);
+    // The above record type is 'observability', which
+    // means we may have been told to temporarily halt
+    // production of them.
+    if (state->emit_observability_records == false ) {
```
ted-ross, line 1892: Ditto on `==`

src/vanflow.c
```diff
@@ -484,6 +488,15 @@ static void _vflow_start_record_TH(vflow_work_t *work, bool discard)
 
     vflow_record_t *record = work->record;
 
+    // If this is one of the record types that supports observability,
```
ted-ross, line 491: This count is being done in the wrong place because this is not where the records are created.  It should be in the non-in-thread version of `vflow_start_record`.

This location is on the back-end of the work queue and thus will not protect against large bursts of records being placed in the work queue.  (Also, in its current location, there is no need for atomics).

src/vanflow.c
```diff
@@ -1909,6 +1940,8 @@ void vflow_serialize_identity(const vflow_record_t *record, qd_composed_field_t
             snprintf(buffer, IDENTITY_MAX, "%s:%"PRIu64, record->identity.s.source_id, record->identity.record_id);
             qd_compose_insert_string(field, buffer);
         }
+    } else {
```
ted-ross, line 1943: The assert at the top of this function does not belong there.  This will cause crashes in the router when emission is suppressed.

ted-ross (changes requested):
Comments inline...  I suspect that the router with this PR will not pass a benchmark test for memory growth.

mgoulish:
I tested this idea of having the upper and lower thresholds very close together by running 50 tests. Each test was 1 million HTTP requests for a megabyte image.    Nothing bad happened, memory usage stayed low, and timings were only 1-2% slower than without the lock (and with previous values for the thresholds).

PR #1805, opened 2025-08-19
Fixes #1804 - Fix the default hosts for tcpListener and tcpConnector.

### LINKED ISSUE #1804: tcpListener host attribute default is inconsistent with description and listener entity
The tcpListener host attribute has a default value of "0.0.0.0". This conflicts with the description of the attribute, and is inconsistent with the behavior of the listener host attribute (no default.)

> A host name, IPV4 or IPV6 literal, or the empty string. The empty string listens on all local addresses...

Configuring a tcpListener with `host: ""` causes the router to bind to `0.0.0.0`, which is problematic for IPv6 only environments.

ted-ross:
This PR makes the host defaults in tcpListener and tcpConnector the same as those in listener and connector.

There should be no backward compatibility problems with this change because the behavior of the tcpListener will be the same in IPv4-only environments and mixed environments.  In IPv6-only environments, tcpListener did not work for all-hosts (only if a specific IPv6 address was configured) but will now work.


c-kruse (approved):
Thank you! Confirmed control plane end to end tests pass on an IPv6 cluster.

PR #1827, opened 2025-10-23
Fixes #1826 - Repurpose the unused area-routing address structure to …

### LINKED ISSUE #1826: Enable cross-network usefulness of dynamic addresses
In the case where two interior networks are joined by a normal-role connection (route-container), and auto links are used to provide cross-connectivity for service addresses, there is a difficulty regarding the reply-to addresses, that are normally derived in dynamic termini.

This enhancement allows an auto link to be created that summarizes the collection of possible destination topological addresses.  This provides seamless integration of existing client/server exchanges where the two endpoints are in different networks.

ted-ross:
…summarize topological addresses within a VAN.


ted-ross:
This PR will not be merged.  There are too many negative consequences to this approach.

PR #1835, opened 2025-11-04
Fixes #1826 - Add a cross-network dynamic address capability

### LINKED ISSUE #1826: Enable cross-network usefulness of dynamic addresses

ted-ross:
Fixes #1826 - Added a test for the cross-network feature
Fixes #1826 - Fixed asan issue
Fixes #1826 - Added support for edge-style cross-net dynamic addresses.
Fixes #1826 - Make the test less resource-intensive for the CI
Fixes #1826 - Introduced the inter-network role and inter-network-specific statistics
Fixes #1826 - Renamed entity managedRouter => network;  Added a doc file describing the feature.
Fixes #1826 - Removed the backward-compatibility features to simplify the patch.
              This will need to be carried separately until the next major release.
For inter-network deliveries, strip only the origin and trace annotations.


ted-ross:
~~This PR will not be merged until v4.0.0.  It is here for review and testing.~~
UPDATE:  The PR has been modified to allow backward compatibility and no longer needs to wait for a major release.

tests/system_tests_cross_network.py
```diff
@@ -0,0 +1,245 @@
...
+from proton import Message
+from proton.handlers import MessagingHandler
+from proton.reactor import Container
+
+from system_test import TestCase, Qdrouterd, main_module, TIMEOUT, TestTimeout
+from system_test import unittest
+
+
+class CrossNetworkPreconfiguredTest(TestCase):
+    """System tests involving a single router"""
```
gabordozsa: The comment does not seem to be correct.

python/skupper_router/management/skrouter.json
```diff
@@ -827,7 +845,8 @@
                         "inter-router",
                         "route-container",
                         "edge",
-                        "inter-edge"
+                        "inter-edge",
+                        "inter-network"
                     ],
                     "default": "normal",
                     "description": "The role of an established connection. In the normal role, the connection is assumed to be used for AMQP clients that are doing normal message delivery over the connection.  In the inter-router role, the connection is assumed to be to another router in the network.  Inter-router discovery and routing protocols can only be used over inter-router connections. route-container role can be used for router-container connections, for example, a router-broker connection.  In the edge role, the connection is assumed to be between an edge router and an interior router.",
```
gabordozsa: Maybe a sentence could be added to the description about the "inter-network" role.

python/skupper_router/management/skrouter.json
```diff
@@ -1001,7 +1020,8 @@
                         "inter-router",
                         "route-container",
                         "edge",
-                        "inter-edge"
+                        "inter-edge",
+                        "inter-network"
                     ],
                     "default": "normal",
                     "description": "The role of an established connection. In the normal role, the connection is assumed to be used for AMQP clients that are doing normal message delivery over the connection.  In the inter-router role, the connection is assumed to be to another router in the network.  Inter-router discovery and routing protocols can only be used over inter-router connections. route-container role can be used for router-container connections, for example, a router-broker connection.  In the edge role, the connection is assumed to be between and edge router and an interior router.",
```
gabordozsa: Same here, description could be extended to cover "inter-network" role.

gabordozsa (approved):
Only a few minor comments.

docs/notes/cross-network.md
```diff
@@ -0,0 +1,177 @@
...
+}
+```
+
+## Using Cross-Network Connectivity as an Endpoint
+
+It is the intent of the implementation that cross-network operations are transparent to the endpoints involved.  No changes to endpoint (producer, consumer, client, server) code need to be made to use this feature.
+
+## Backward Compatibility and Breaking Changes
+
+This feature represents an incompatible change from the version 3 router because the format of dynamic addresses has been changed to include a field for the network ID.  This feature must be introduced in a major release (version 4).  Until that time, the project team shall maintain a development branch that tracks main and carries a patch for this feature.
```
fgiorgetti: So should this PR change the target branch from main to the new development branch?

ted-ross: No.  This is intended to merge into main when approved.  That document text should be removed.

fgiorgetti:
@ted-ross one observation.
When a router does not have a network entity defined, a query against the network entity returns:

```json
[
  {
    "name": "network/0",
    "identity": "network/0",
    "type": "io.skupper.router.network"
  }
]
```

If I add a Network CR, which causes the `network.networkId` to be set, the VAN communication stops working, which should be expected. Next, if I remove the CR and query the management API for network entity I get:

```json
[
  {
    "networkId": "",
    "name": "network/0",
    "identity": "network/0",
    "type": "io.skupper.router.network"
  }
]
```

I believe because it can only be updated, but not removed.
The communication with the existing VAN is not restored unless the router is restarted.

Is that the expected behavior?

ted-ross:
> Is that the expected behavior?

No, it should go back to the original behavior.  I've pushed a fix up for this.

fgiorgetti (approved):
Worked great with VANs using router 3.4.2 and 3.5.0.
Setting the networkId worked as expected and removing it, returned the VAN to its working state.

PR #1849, opened 2026-01-28
Specify address and port attributes as requried

### LINKED ISSUE #1848: Spec: Correct tcpListener/tcpConnector required fields
## What
In skrouter.json the tcpListener and tcpConnector entity specs have their `address` and `port` attributes specified without `"required": true`, but the C implementation uses `qd_entity_get_string` for these indicating they are required by the implementation.

Inconsistencies like this make it difficult to generate go type definitions based off of the entity spec, and also causes operations with missing attributes to fall back on validations in C instead of in the initial python layer.

```
# required=false
[christian@toolbx skupper-router] (main)$ skmanage create --type tcpListener --name tl01 observer=auto
InternalServerErrorStatus: CError: Python: KeyError: 'port'
# required=true
[christian@toolbx skupper-router] (main *)$ skmanage create --type tcpListener --name tl01 observer=auto
BadRequestStatus: Missing required attribute 'port'
```

## Suggestion
Explicitly mark these attributes as required in the spec.

A step further would be to purge all of the usages of `"required": false` from the spec in order to avoid confusion over the default required-ness of attributes. It is my best guess that confusing this was a contributing factor to this misconfiguration.

c-kruse:
Sets required=true for the address and port attributes on the tcp adaptors: tcpListener and tcpConnector. These attributes have always been required by the implementation but have ben underspecified in the spec document until now.

Fixes https://github.com/skupperproject/skupper-router/issues/1848

### DIFF
--- a/python/skupper_router/management/skrouter.json
+++ b/python/skupper_router/management/skrouter.json
@@ -1089,6 +1089,7 @@
                 "address": {
                     "description":"Address of this bridge",
                     "type": "string",
+                    "required": true,
                     "create": true
                 },
                 "host": {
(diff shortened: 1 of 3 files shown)


ted-ross:
This PR has been overcome by events.  The changes related to multi-address listeners causes this PR to be mostly obsolete.  Closing.

PR #1854, opened 2026-02-27
Fixes #1852 - Unmap addresses from routers that become unreachable in topology changes.

ted-ross:
NOTE:  This is a draft PR.  I'm putting this here as a suggested way to solve the problem in the raised issue.

This needs more testing and examination before it can be accepted.

### DIFF
--- a/python/skupper_router_internal/router/node.py
+++ b/python/skupper_router_internal/router/node.py
@@ -161,9 +161,17 @@ class NodeTracker:
             self.container.router_adapter.set_radius(radius)
 
             ##
+            # Mark all of the nodes as un-visited so we can identify newly unreachable nodes
+            ##
+            visited = {}
+            for node_id in self.nodes.keys():
+                visited[node_id] = False
+
+            ##
             # Update the next hops and valid origins for each node
             ##
             for node_id, next_hop_id in next_hops.items():
+                visited[node_id] = True
                 node     = self.nodes[node_id]
                 next_hop = self.nodes[next_hop_id]
                 vo       = valid_origins[node_id]
@@ -171,6 +179,14 @@ class NodeTracker:
                 node.set_next_hop(next_hop)
                 node.set_valid_origins(vo)
                 node.set_cost(cost)
+
+            ##
+            # Un-map the addresses on each node that is no longer reachable
+            ##
+            for node_id, reachable in visited.items():
+                if not reachable:
+                    self.container.log_ls(LOG_INFO, "Node unreachable: %s" % node_id)
+                    self.nodes[node_id].unmap_all_addresses()
 
         ##
         # Send link-state requests and mobile-address requests to the nodes
--- a/python/skupper_router_internal/router/node.py
+++ b/python/skupper_router_internal/router/node.py
@@ -43,13 +43,17 @@ class NodeTracker:
         self.nodes                 = {}  # id => RouterNode
         self.nodes_by_link_id      = {}  # link-id => node-id
         self.maskbits              = []
+        self.unreachable_ids       = []  # Router-ids of unreachable routers
+        self.unreach_holdoff       = 0
+        self.invalidate_unreach    = False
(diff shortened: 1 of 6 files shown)


python/skupper_router_internal/router/node.py
```diff
@@ -542,6 +535,9 @@ def mobile_address_requested(self):
             return True
         return False
 
+    def has_mobile_addresses(self):
```
ted-ross: The fact that the mobile_address_sequence is greater that zero does not indicate that the node has mobile addresses.  It's possible that mobile addresses were added and then removed.  The sequence will be incremented at each change.  This function's name is misleading.

python/skupper_router_internal/router/node.py
```diff
@@ -187,26 +182,24 @@ def tick(self, now):
...
+                self.invalidate_unreach = True
 
         ##
         # Invalidate the addresses on any remote routers after the holdoff interval
         ##
         if self.invalidate_unreach:
             if self.unreach_holdoff == 0:
                 for node_id in self.unreachable_ids:
-                    if node_id in self.nodes:
+                    if node_id in self.nodes and self.nodes[node_id].has_mobile_addresses():
```
ted-ross: Is this a necessary optimization?  Are there cases where the second clause prevents the unnecessary unmapping of addresses?

python/skupper_router_internal/router/node.py
```diff
@@ -92,7 +92,9 @@ def _do_expirations(self, now):
             if now - node.link_state.last_seen > self.ls_max_age:
                 if node.link_state.has_peers():
                     node.link_state.del_all_peers()
-                    self.recompute_topology = True
+                    if node.is_neighbor() or node.next_hop_router is not None:
+                        # Recompute topology if the node is reachable
+                        self.recompute_topology = True
```
ted-ross, line 97: Is this a necessary optimization?  Does this always take effect when the unmap interval is less than the LS_MAX_AGE?

ted-ross (commented):
The changes to the PR are very good.  I inserted a couple of questions for discussion before we approve this.

python/skupper_router_internal/router/node.py
gabordozsa: Yes, I agree. Maybe  `has_mobile_address_information() ` would be a better name?

python/skupper_router_internal/router/node.py
gabordozsa: I have seen that happens while testing the PR on my laptop.  After deleting  a link, the unreachable router node is kept around for ~60s (max link state age). If another topology recompute happens in the meantime then unmapping is called again. When I deleted one link after another I saw unmapping called multiple times for the same node.

python/skupper_router_internal/router/node.py
gabordozsa: I observed the following sequence of events:

1. link to `routerB` gets deleted => `roterB` is deleted from our peers
2. topology gets recomputed =>` rourerB` is marked unreachable
3. addresses get unmapped for `routerB`
4. link state expires for `routerB` => peers of `routerB` are deleted = > topology gets recomputed but it does not actually change since `routerB` was marked unreachable already at step 2.

python/skupper_router_internal/router/node.py
ted-ross: Ok.  Since the unmapping is a relatively expensive process, I agree that this is a good optimization.

python/skupper_router_internal/router/node.py
ted-ross: Thanks.  Again, if this happens every time a node goes off-line, this is a good optimization.  I believe it is safe because the node being removed is already not a part of the currently computed topology.

python/skupper_router_internal/router/node.py
ted-ross: Or `received_mobile_updates`, which is a little more accurate.

gabordozsa (commented):
LGTM

PR #1862, opened 2026-04-17
Re-factor and stabilize the grpc system test

### LINKED ISSUE #1861: system_tests_grpc is failing on CentOS 9 proton main
The grpc test started failing on various PRs. 

`Compile and Test (centos:stream9, RelWithDebInfo, OFF, proton main, shard 2 of 2) ` CI log tail: 
```
/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:1514: in tearDownClass
    cls.tester.teardown()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <system_test.Tester object at 0x7f491492b550>

    def teardown(self):
        """Clean up (tear-down, stop or close) objects recorded via cleanup()"""
        self.cleanup_list.reverse()
        errors = []
        for obj in self.cleanup_list:
            try:
                for method in ["teardown", "tearDown", "stop", "close"]:
                    cleanup = getattr(obj, method, None)
                    if cleanup:
                        cleanup()
                        break
            except Exception as exc:
                errors.append(exc)
        if errors:
>           raise RuntimeError("Errors during teardown: \n\n%s" % "\n\n".join([str(e) for e in errors])) from errors[0]
E           RuntimeError: Errors during teardown: 
E           
E           Expecting value: line 1 column 1 (char 0)
E           
E           Expecting value: line 1 column 1 (char 0)

/home/runner/work/skupper-router/skupper-router/skupper-router/tests/system_test.py:1372: RuntimeError
---------------------- generated xml file: /__w/skupper-router/skupper-router/skupper-router/build/tests/junitxmls/system_tests_grpc.xml -----------------------
=================================================================== short test summary info ====================================================================
ERROR ::GrpcServiceMethodsTest::test_grpc_01_unary - ConnectionRefusedError: [Errno 111] Connection refused
ERROR ::GrpcServiceMethodsTest::test_grpc_02_bidirectional_stream - ConnectionRefusedError: [Errno 111] Connection refused
ERROR ::GrpcServiceMethodsTest::test_grpc_03_server_stream - ConnectionRefusedError: [Errno 111] Connection refused
ERROR ::GrpcServiceMethodsTest::test_grpc_04_client_stream - ConnectionRefusedError: [Errno 111] Connection refused
ERROR ::GrpcServiceMethodsTestOverTcpTls::test_grpc_04_client_stream - RuntimeError: Errors during teardown: 

Expecting value: line 1 column 1 (char 0)

Expecting value: line 1 column 1 (char 0)
=========================================================== 4 passed, 5 errors in 302.83s (0:05:02) ============================================================


97% tests passed, 1 tests failed out of 38

Total Test time (real) = 323.66 sec

The following tests FAILED:
	 56 - system_tests_grpc (Failed)
```

gabordozsa:
The CI failure is related to the grpcio module which can have problem with fork() support in older python versions (python 3.9 in this case). The suggested fix is to disable fork support with a grpcio specific env var.

I also realized that the test launches duplicate router and grpc server instances for the TLS runs because of the inheritance structure between the two test classes. I factored-out. a common base class to fix that.

Fixes #1861

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -853,12 +853,14 @@ class Qdrouterd(Process):
                     "BadRequestStatus" in exception_string:
                 pass
             else:
-                raise e
-        try:
-            super(Qdrouterd, self).teardown()
-        except Exception as exc:
-            # re-raise _after_ dumping all the state we can
-            teardown_exc = exc
+                # re-raise _after_ dumping all the state we can
+                teardown_exc = e
+        if teardown_exc is None:
+            try:
+                super(Qdrouterd, self).teardown()
+            except Exception as exc:
+                # re-raise _after_ dumping all the state we can
+                teardown_exc = exc
 
         def check_output_file(filename, description):
             """check router's debug dump file for anything interesting (should be
--- a/tests/system_tests_grpc.py
+++ b/tests/system_tests_grpc.py
@@ -17,11 +17,15 @@
 # under the License.
 #
 import unittest
+import os
 
 from http1_tests import wait_tcp_listeners_up
 from system_test import TestCase, Qdrouterd, TIMEOUT, CA_CERT, SERVER_CERTIFICATE, SERVER_PRIVATE_KEY, \
     CLIENT_CERTIFICATE, CLIENT_PRIVATE_KEY, CLIENT_PRIVATE_KEY_PASSWORD, SERVER_PRIVATE_KEY_PASSWORD
 try:
+    # grpcio specific setting to make the test reliable on CentOS 9 (python 3.9)
+    os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "false"
+
     import grpc
     import friendship_server as fs
     from friendship_pb2_grpc import FriendshipStub
@@ -38,8 +42,7 @@ def skip_test():
     return _GRPC_UNAVAILABLE
 
 
-class GrpcServiceMethodsTest(TestCase):
-
+class GrpcServiceMethodsBase():
     """
     Data for the grpc service
     """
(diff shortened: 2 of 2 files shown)


tests/system_test.py
```diff
@@ -853,12 +853,14 @@ def teardown(self):
...
                 pass
             else:
-                raise e
-        try:
-            super(Qdrouterd, self).teardown()
-        except Exception as exc:
-            # re-raise _after_ dumping all the state we can
-            teardown_exc = exc
+                # re-raise _after_ dumping all the state we can
+                teardown_exc = e
```
ganeshmurthy, line 857: In the earlier code, we did `raise e`  which stopped execution on that line which means the router never got torn down. In the new code we do `teardown_exc = e` and call `super(Qdrouterd, self).teardown()` on the router only if there was no `teardown_exc`. I was wondering if `super(Qdrouterd, self).teardown()` should always be called whether `teardown_exc` is None of not. 

Also, we are not doing anything with `teardown_exc`, should it be raised at the end of the function so it can be caught by the caller ?

tests/system_tests_grpc.py
```diff
@@ -210,6 +168,54 @@ def test_grpc_04_client_stream(self):
...
+                                              wait=True)
+        wait_tcp_listeners_up(cls.router_qdr.addresses[0])
+
+        # If you wanna try it without the router, set the grpc_channel
+        # directly to the grpc_server_port
+        cls.grpc_channel = grpc.insecure_channel('127.0.0.1:%s' %
+                                                 cls.router_listener_port)
+        cls.grpc_stub = FriendshipStub(cls.grpc_channel)
+
+    @classmethod
```
ganeshmurthy, line 211: Can this `tearDownClass` function be moved to base class or does that not work ?

tests/system_test.py
gabordozsa: I was trying to keep the original behaviour,  so `tearDown()` is not called if exception is raised at line 856. Do you think that this should be changed?  I only added an extra step to dump logs before the exception is raised. 

The original exception, which is saved in `teardown_exc`,   is still raised below at line 922.

tests/system_tests_grpc.py
gabordozsa: I think that would not work. The base class does not inherit from `TestCase`, so it cannot call  `TestCase.tearDownClass()`.

ganeshmurthy: Yeah ok, that's what I thought. Sounds good.

tests/system_test.py
ganeshmurthy: > so `tearDown()` is not called if exception is raised at line 856

I was wondering what would happen if teardown_exc is not None and we never call router teardown(), will the router not shut down and hang around after the tests are run ? 



> The original exception, which is saved in `teardown_exc`, is still raised below at line 922.

Ah yes, I missed that. Apologies.

ganeshmurthy (approved):

tests/system_test.py
gabordozsa: > I was wondering what would happen if teardown_exc is not None and we never call router teardown(), will the router not shut down and hang around after the tests are run ?

Yes, it can happen that  routers hang around. I think this is okay because we had an exception within the `tearDown()` method, which we want to raise , so we cannot expect a clean final state. (E.g. `grpcio `servers will also hang around if this exception happen in the grpc test because they are only removed after `Qdrouterd.tearDown()` completes.)

ted-ross (approved):

tests/system_test.py
ganeshmurthy: Sounds good. Also, before this PR, we did a `raise e` which also led to the router's teardown() not getting called. Please go ahead and merge this PR, thanks.

PR #1867, opened 2026-05-25
Enable dynamic change of connector cost

### LINKED ISSUE #1857: Enable connector cost to be dynamically changed without interruption to existing connection

gabordozsa:
Enable dynamic change of the cost of inter-router connector and its existing connections.

Fixes #1857 

The major steps of connector cost change are as follows.

Connector side:
1.  Mgmt thread calls update on ConnectorEntity
2. Update function (in C) changes the connector cost, then it posts  a core action for each existing connections.
3. Core thread updates  `qdr_connection_t->inter_router_cost`, then it posts a general work item to update the peer cost in the local link state.
4.  Peer cost in local link state (in Python) is updated, `ls_seq` is bumped

Listener side: 

5. Llistener side receives LSU from connector side with updated cost. It posts a core action to update the cost of the connection.
6. Listener side  core thread updates the `qdr_connection_t->inter_router_cost`, then it posts a general work item to update the peer cost in the local link state.

Notes:

- The same core function (`qdr_core_update_connection_cost_CT`) handles both steps 3 and 6.  
- The remote cost is saved in `qdr_connection_t` at connection open time. It is used at the connector side to find the maximum cost of connector and listener when the dynamic update happens. This approach works as long as listener cost is immutable.

### DIFF
--- a/include/qpid/dispatch/protocol_adaptor.h
+++ b/include/qpid/dispatch/protocol_adaptor.h
@@ -363,6 +363,7 @@ typedef void (*qdr_connection_bind_context_t) (qdr_connection_t *context, void *
  * @param incoming True iff this connection is associated with a listener, False if a connector
  * @param role The configured role of this connection
  * @param cost If the role is inter_router, this is the configured cost for the connection.
+ * @param remote_cost If the role is inter_router, this is the cost sent by the peer router in the open frame
  * @param management_id - A unique identifier that is used in management and logging operations.
  * @param label Optional label provided in the connection's configuration.
  * @param strip_annotations_in True if configured to remove annotations on inbound messages.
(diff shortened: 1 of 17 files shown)


c-kruse:
Interested to hear what the upgrade experience is in mixed router version networks. I.e. if a connector is updated and the listener is on a current router release what happens/what should happen.

That said, hooked it up with a control plane [experiment](https://github.com/c-kruse/skupper/tree/exp/dynamic-link-cost) and it went very smoothly!

gabordozsa:
> Interested to hear what the upgrade experience is in mixed router version networks. I.e. if a connector is updated and the listener is on a current router release what happens/what should happen.
> 

The cost of existing inter-router connection will still be updated correctly at the connector side. The connection cost at listener side will not change,  resulting asymmetric peer costs at connector and listener sides. That should not be a problem from functional point of view, as far as I understand it.

For new connections, everything would work as expected after connector cost update. 

> That said, hooked it up with a control plane [experiment](https://github.com/c-kruse/skupper/tree/exp/dynamic-link-cost) and it went very smoothly!

Thank you for giving it a try!

PR #1868, opened 2026-05-27
Fix clang-analyzer workflow

gabordozsa:


ganeshmurthy (approved):

ganeshmurthy:
On an unrelated note, I was wondering if we should remove RPM builds ? If I remember correctly, we do not produce upstream skupper-router rpms. Is there a reason we want to keep these rpm builds ? I ask because the builds are failing

ted-ross:
> On an unrelated note, I was wondering if we should remove RPM builds ? If I remember correctly, we do not produce upstream skupper-router rpms. Is there a reason we want to keep these rpm builds ? I ask because the builds are failing

I'm in favor of removing the RPM builds.

gabordozsa:
> > On an unrelated note, I was wondering if we should remove RPM builds ? If I remember correctly, we do not produce upstream skupper-router rpms. Is there a reason we want to keep these rpm builds ? I ask because the builds are failing
> 
> I'm in favor of removing the RPM builds.

+1 

@ganeshmurthy Could you open PR to remove the RPM builds? Thanks!

ganeshmurthy:
> > > On an unrelated note, I was wondering if we should remove RPM builds ? If I remember correctly, we do not produce upstream skupper-router rpms. Is there a reason we want to keep these rpm builds ? I ask because the builds are failing
> > 
> > 
> > I'm in favor of removing the RPM builds.
> 
> +1
> 
> @ganeshmurthy Could you open PR to remove the RPM builds? Thanks!

https://github.com/skupperproject/skupper-router/pull/1869

PR #1872, opened 2026-06-10
Get memory limit by walking up the cgroup tree.

mgoulish:
This change gets the router memory limit by starting at the cgroup node that represents the router process, and then walking up the tree, parent to parent, to find the most restrictive memory limit. 

When I ran the router this way:    

         systemd-run --scope --user -p MemoryMax=1G  skrouterd -c ./A.conf

It correctly produced this log output:

        2026-06-10 01:02:27.453308 -0400 TCP_ADAPTOR (info) Router buffer memory ceiling: 1.00 GiB (263689 buffers)

And then it correctly ran my test of 20,000 TCP connections across a 2-router network, and both routers stayed below (just barely below)  1.0 GB.

### DIFF
--- a/src/adaptors/tcp/tcp_adaptor.c
+++ b/src/adaptors/tcp/tcp_adaptor.c
@@ -17,7 +17,6 @@
  * under the License.
  */
 
-
 #include <qpid/dispatch/ctools.h>
 #include <qpid/dispatch/amqp.h>
 #include <qpid/dispatch/enum.h>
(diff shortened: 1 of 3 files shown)


ted-ross (changes requested):
Please raise an issue for this bug.  Also, please remove the changes to tcp_adaptor.c and server.c, which are just added empty lines.

ted-ross (approved):

PR #1874, opened 2026-06-22
Evan/fake half close

evanwang9x:
Test-Script to simulate half closed connections

### DIFF
--- a/half_close_sim.py
+++ b/half_close_sim.py
@@ -1,0 +1,328 @@
+#!/usr/bin/env python3
+"""
+half_close_sim.py - Simulate half-closed TCP connections to observe FD/connection
+leaks in the Skupper router.
+
+
+When a client half-closes (sends FIN via shutdown(SHUT_WR)), a router that does
+not propagate the close will remain in CLOSE_WAIT indefinitely, leaking file
+descriptors and memory.
+
+
+Modes:
+ server   Run a local TCP echo server that simulates the CLOSE_WAIT bug
+ client   Connect to a target, send data, half-close, and hold the connection open
+ both     Run server and clients together in one process
+
+
+Examples:
+
+
+ # Local loopback test — server simulates the CLOSE_WAIT bug, client half-closes:
+ python half_close_sim.py both --port 9090 --count 5
+
+
+ # Against a real Skupper listener (router must already be running):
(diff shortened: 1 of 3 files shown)


evanwang9x:
I forgot to include a.conf but it should be in skupper-router/build

router {
   mode: interior
   id: A
}

listener {
   port: amqp
   authenticatePeer: no
   saslMechanisms: ANONYMOUS
}

listener {
   port: 10000
   authenticatePeer: no
   saslMechanisms: ANONYMOUS
   role: inter-router
}

tcpListener {
   port: 9090
   address: tcp-echo
   name: tcp-echo-listener
}

tcpConnector {
   port: 9091
   host: 127.0.0.1
   address: tcp-echo
   name: tcp-echo-connector
}

evanwang9x:
I tried using skmanage DELETE --type io.skupper.router.connection --identity <ID> to force-close connections, but this returns rc=1 with no error message for all connection objects. Running Skstat -c afterwards shows the half closed tcp connections still existing. So I switched to using update adminstatus instead. 

Also RSS memory usage is usually around 9.3MiB per 100 half closed connections. Doesn't reset for some reason after the connection killer is run, but if you rerun the test script to make TCP connections again it will reclaim the used RSS memory instead of causing a memory leak which is good.

evanwang9x:
Note my half_close_sim.py doesn't perfectly replicate the half-closed connection bug. The symptoms of the half closed connections is the router failing to propagate the client's FIN to the backend, leaving the backend unaware the client left and keeping the connection alive indefinitely on the server side. What this script does is simulate the effects of that by holding connections open by having the backend deliberately sleep instead of closing after receiving the FIN, which produces the same result: stuck TCP connections accumulating in the router with memory allocated to track their state.

This difference matters as it limits the options I have for connection reaper right now. Currently, I am just checking lastDlvSeconds to see the last time data flowed through the connections and if it exceeds the set time to have it be deleted. A way I wanted to test was writing a script to forcefully send a small packet of data from client to server and if that fails then to kill it. But my simulation right now would let that succeed so I can't test that alternative method. I'm not sure if it's a much more efficient method though as this might slow down the router if there's a ton of tcp connections that have to send data. But it could be better though as it would close the half closed connections early instead of letting them stay alive for X amount of time before closing them.

